PR #211, opened 2024-02-28
Checking that docker is running fails on MAC if docker.sock is not enabled

### LINKED ISSUE #210: Checking that docker is running fails on MAC if docker.sock is not enabled
### Description of the bug

For Ksail to detect propperly that the docker daemon is running, MAC needs to be configured to have the docker.sock enabled.

To allow Ksail to talk with docker the user must enable the docker.sock See attached image. Somehow this needs to be documented or somehow ksail should help the user to know what to do.

Screenshot 2024-02-28 at 11 21 19

devantler:
Fixes #210


devantler:
Hey @AlmiS

Does this solution solve your issue with the default Docker Socket not being enabled?

The output will now be:

image

When the the user is running the `ksail up <name-of-cluster>` command on a Mac. The library I use to connect to Docker is not able to infer what caused the connection issue, so the message will be shown in all cases where the connection fails on a Mac, hence why I formulated it as a question, rather than an instruction :-)

PR #218, opened 2024-03-01
Refactor template engine and generator classes

devantler:
This pull request refactors the template engine and generator classes to improve code organization and readability. It introduces the `ITemplateEngine` and `IGenerator` interfaces, and updates the `TemplateEngine` and `Generator` classes to implement these interfaces. The refactoring also includes moving the `RenderAsync` method from the `TemplateEngine` class to the `ITemplateEngine` interface, and updating the method implementation in the `TemplateEngine` class. Additionally, the `Generator` class now accepts an `ITemplateEngine` instance in its constructor, allowing for better dependency injection.


src/KSail/Commands/Init/Generators/K3dGenerator.cs
```diff
@@ -0,0 +1,25 @@
+using KSail.Models.K3d;
+using KSail.TemplateEngine;
+
+namespace KSail.Commands.Init.Generators;
+
+class K3dGenerator
+{
+  readonly Generator _generator = new(new TemplateEngine.TemplateEngine());
+  internal async Task GenerateK3dConfigAsync(string filePath, string clusterName)
```
devantler:
```suggestion
  readonly Generator _generator = new(new TemplateEngine.TemplateEngine());

  internal async Task GenerateK3dConfigAsync(string filePath, string clusterName)
```

PR #354, opened 2024-11-19
Update dependencies

devantler:
Correct command arguments in KSailUpdateCommandTests and update KubernetesProvisioner dependencies to version 0.0.24.

### DIFF
--- a/KSail.Tests/Commands/Update/KSailUpdateCommandTests.cs
+++ b/KSail.Tests/Commands/Update/KSailUpdateCommandTests.cs
@@ -56,7 +56,7 @@ public class KSailUpdateCommandTests : IAsyncLifetime
 
     //Act
     int initExitCode = await ksailInitCommand.InvokeAsync("");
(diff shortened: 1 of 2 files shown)


KSail.Tests/Commands/Update/KSailUpdateCommandTests.cs
```diff
@@ -56,7 +56,7 @@ public async Task KSailUpdate_WithDefaultOptions_SucceedsAndCreatesAndUpdatesDef
 
     //Act
     int initExitCode = await ksailInitCommand.InvokeAsync("");
-    int upExitCode = await ksailUpCommand.InvokeAsync("--reconcile false --destroy");
+    int upExitCode = await ksailUpCommand.InvokeAsync("ww --destroy");
```
devantler:
```suggestion
    int upExitCode = await ksailUpCommand.InvokeAsync("--reconcile false --destroy");
```

PR #702, opened 2025-03-05
feat: Add support for Cilium CNI in cluster configuration

### LINKED ISSUE #335: Add an option to use Cilium on all supported distributions
### Description

I imagine it being through installing a Helm Chart, or if the distribution has native support for it, it should use that approach.

### LINKED ISSUE #628: [epic]: Cilium support in KSail
### Description

_No response_

devantler:
Introduce Cilium as an option for the Container Network Interface (CNI) in the cluster configuration, enabling users to select it instead of the distributions default option. Update relevant configurations and tests to accommodate this new CNI support.

- Closes #628 
- Closes #335

### DIFF
--- a/src/KSail/Commands/Up/Handlers/KSailUpCommandHandler.cs
+++ b/src/KSail/Commands/Up/Handlers/KSailUpCommandHandler.cs
@@ -290,7 +290,7 @@ class KSailUpCommandHandler
     switch (config.Spec.Project.CNI)
     {
       case KSailCNIType.Cilium:
-        Console.WriteLine("🔼 Bootstrapping Cilium CNI");
+        Console.WriteLine("🐝 Installing Cilium CNI");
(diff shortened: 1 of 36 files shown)


src/KSail/Commands/Up/Handlers/KSailUpCommandHandler.cs
```diff
@@ -283,6 +285,21 @@ await _deploymentTool.BootstrapAsync(
     Console.WriteLine();
   }
 
+  async Task BootstrapCNI(KSailCluster config, CancellationToken cancellationToken)
+  {
+    switch (config.Spec.Project.CNI)
+    {
+      case KSailCNIType.Cilium:
+        Console.WriteLine("🔼 Bootstrapping Cilium CNI");
```
devantler:
```suggestion
        Console.WriteLine("🐝 Installing Cilium CNI");
```

PR #710, opened 2025-03-06
feat: Add overwrite option to generator configuration

### LINKED ISSUE #605: `ksail init --overwrite` to support applying new KSail configuration into existing setups.
### Description

I believe this is a good enough solution, as I expect most projects to be in git, so any lost configurations can be reapplied to the files that might be lost.

It might make sense to support `ksail init --merge` in the future, to support a less intrusive approach.

devantler:
Introduce an overwrite option in the generator configuration to control file generation behavior. This allows users to specify whether existing files should be overwritten during the generation process.

- Closes #605


devantler:
- [ ] Add console log handling to all Kubernetes Generators

PR #1478, opened 2025-12-18
BREAKING CHANGE: Enable push trigger for release workflow on main

devantler:


devantler:
@copilot fix golangci-lint issues and tests

Issue #206, opened 2024-02-28
Create a KSail YAML config with a hosted schema


devantler:
### Description

For example which agekey it uses.

PR #4141, opened 2026-04-18
refactor(update-skills): use reusable workflow for copilot skills updates

devantler:
Switches `.github/workflows/update-skills.yaml` over to the [`devantler-tech/reusable-workflows/.github/workflows/update-copilot-skills.yaml`](https://github.com/devantler-tech/reusable-workflows/blob/main/.github/workflows/update-copilot-skills.yaml) reusable workflow so the daily run:

1. Resolves the latest ref + commit SHA per source and writes them back to `skills-lock.json` (via the new [`update-copilot-skills`](https://github.com/devantler-tech/actions/tree/main/update-copilot-skills) action).
2. Installs the pinned skills via `setup-copilot-skills`.
3. Opens a PR on change.

Before this, the workflow only reinstalled whatever was already pinned, so `skills-lock.json` never actually got bumped — exactly the bug [devantler-tech/actions#87](https://github.com/devantler-tech/actions/issues/87) tracked.

## Changes

- Replace the hand-rolled `gh >= 2.90.0` bootstrap + `setup-copilot-skills` + `peter-evans/create-pull-request` steps with a single `uses:` call to the shared workflow (-72 / +5 lines).
- Pin to [`devantler-tech/reusable-workflows@6eea016`](https://github.com/devantler-tech/reusable-workflows/commit/6eea0169) (post-#206) until a tagged release is cut.

## Context

- Final piece of the stacked series: [devantler-tech/actions#88](https://github.com/devantler-tech/actions/pull/88) + [#92](https://github.com/devantler-tech/actions/pull/92) → [devantler-tech/reusable-workflows#206](https://github.com/devantler-tech/reusable-workflows/pull/206) → this PR.

### DIFF
--- a/.github/workflows/update-skills.yaml
+++ b/.github/workflows/update-skills.yaml
@@ -12,78 +12,11 @@ permissions: {}
 
 jobs:
   update-copilot-skills:
-    runs-on: ubuntu-latest
     permissions:
       contents: write
       pull-requests: write
-    steps:
-      - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@f808768d1510423e83855289c910610ca9b43176 # v2.17.0
-        with:
-          egress-policy: audit
-
-      - name: 📄 Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          persist-credentials: true
-
-      # Ensure gh >= 2.90.0 (required for `gh skill`, introduced in v2.90.0).
-      # The ubuntu-latest runner image currently ships gh 2.89.0, and the
-      # upstream setup-copilot-skills action's in-place /usr/local/bin install
-      # is masked by bash's hash cache, so we prepend a fresh gh to PATH here.
-      - name: 🔧 Install gh >= 2.90.0
-        shell: bash
-        env:
-          GH_VERSION: "2.90.0"
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-skills.yaml
```diff
@@ -12,75 +12,8 @@ permissions: {}
...
-            - `skills-lock.json`
-            - `.agents/skills/`
-          branch: deps/copilot-skills-update
-          labels: dependencies,automation
-          delete-branch: true
+    uses: devantler-tech/reusable-workflows/.github/workflows/update-copilot-skills.yaml@6eea016969bceed84f4fb38c8e79fb04a2cadc31 # main @ 2026-04-18
+    with:
+      skills-lock: skills-lock.json
+      agent: github-copilot
+      scope: user
```
devantler: The idea is to install the skills into the repo, so they are available to both local and cloud runners. So installing it in the user scope does not make much sense.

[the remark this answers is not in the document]
devantler: Good catch — added `permissions: { contents: write, pull-requests: write }` at the `update-copilot-skills` job level in 5fb7aa1. Caller permissions do cap the reusable workflow, so the job-level grant is needed here.

devantler: Switched to `scope: project` in 0e8e679 so the skills land in the repo and are available to both local and cloud runners.

PR #4892, opened 2026-05-26
fix(ci): exclude self-contained sub-apps from MegaLinter linters

devantler:

## What this PR does

Excludes the **self-contained sub-apps** from MegaLinter, so the `🧹 Lint - mega-linter` job's duplicate detection stops flagging code that lives in app-private toolchains:

- **jscpd** — all 6 reported TypeScript clones were inside `vsce/` (the self-contained VS Code extension). Added `vsce/`, `web/ui/` and `desktop/` to `.jscpd.json`'s ignore list (jscpd is a "project" linter, so it reads `.jscpd.json` ignore — not MegaLinter's `FILTER_REGEX_EXCLUDE`). This drops jscpd from **7 clones to 1**. ✅ verified in CI (run on `43306b57`/`04c61f02`).
- **desktop cgo filter** (original commit) — keeps `desktop/.*` in `FILTER_REGEX_EXCLUDE` so the per-file repo-wide linters skip the separate cgo module.

This is intentionally scoped to the **clean config layer**: the one change that monotonically reduces the lint failures with correct, low-risk config.

## The lint gate masks real backlogs — three remaining blockers (this PR alone does NOT turn it green)

Investigation showed the `lint` job has been failing on **three independent linters**, two of which were masking real findings behind *environmental* failures:

1. **jscpd — 1 real in-tree Go clone.** `pkg/svc/tenant/argocd.go [400:413]` ↔ `pkg/fsutil/configmanager/talos/manager.go [228:241]` (the YAML-files-in-a-directory loop; `talos` already factored its copy into `forEachYAMLFile`). With `threshold: 0` this still fails after the sub-app exclusion. Clean fix = a shared `fsutil` helper used by both — but the two copies differ in file-reading (`fsutil.ReadFileSafe` vs `os.ReadFile`+`//nolint:gosec`), so unifying them is a **security-sensitive refactor** that deserves its own reviewed PR.
2. **zizmor — 401 masks a ~119-finding backlog.** MegaLinter strips `GITHUB_TOKEN`, so zizmor can't reach the GitHub API and reports a single HTTP 401. Passing the token (`ACTION_ZIZMOR_UNSECURED_ENV_VARIABLES: [GITHUB_TOKEN]`) lets it authenticate — but it then completes its audits and surfaces **~119 findings** (incl. High-confidence `github-app` "dangerous use of GitHub App tokens" in `cd.yaml`). So the token fix makes zizmor *functional*, not *passing*; the revealed findings are a separate triage effort and were deliberately **not** bundled here (it would raise the error count, not reduce it).
3. **osv-scanner — cgo crash masks a dependency-vuln backlog (security).** In the MegaLinter `go` image, `osv-scanner scan source --recursive .` crashes type-checking `desktop/`'s webview cgo (`could not import C`) and aborts *before* scanning. Running the same command in the standalone `osv-scanner:2.3.8` image (no Go toolchain) completes and surfaces ~68 advisories on indirect deps (`x/crypto` 0.51→0.52, `x/net` 0.54→0.55, `x/image` 0.38→0.39, plus `k8s.io/kubernetes` no-fix meta-module false-positives). Fix path: make the scanner functional (`--no-call-analysis=all`, **validate in the `go` image**) **and** bump the vulnerable indirect deps **and** osv-ignore the k8s findings. Security + dependency work, not config.

**Net:** the gate needs a coordinated effort (Go-clone refactor + zizmor findings triage + osv dep/security work) to go green. This PR ships the one clean config win and maps the rest.


devantler:

**Corrected root cause — this PR's `FILTER_REGEX_EXCLUDE: desktop/.*` does not unblock the merge.** CodeQL was cleared by the `7a89c520` re-trigger, so the only remaining gate is the `🧹 Lint - mega-linter` job in `✅ Validate Go Project`. That job fails on **three independent linters**, and a file-list filter addresses none of them:

| Descriptor | Linter | Errs | Cause | Fix |
|---|---|---|---|---|
| REPOSITORY | `osv-scanner` v2.3.8 | 1 | Megalinter runs `osv-scanner scan source --recursive .`, whose Go **call-graph analysis** type-checks the whole root module. `desktop/` has **no own `go.mod`**, so it's part of `…/ksail/v7` and gets loaded → `could not import C` on the cgo `desktop/main.go`. A whole-module source scan ignores `FILTER_REGEX_EXCLUDE` (that only filters file-based linters). | `REPOSITORY_OSV_SCANNER_ARGUMENTS: --no-call-analysis=all` — keeps lockfile vuln scanning, drops the cgo build (validate locally). |
| ACTION | `zizmor` v1.25.0 | 1 | `couldn't list tags for actions/checkout` → `HTTP 401`: megalinter strips `GITHUB_TOKEN`, so zizmor's online audits (`artipacked`) can't reach the GitHub API. | `ACTION_ZIZMOR_UNSECURED_ENV_VARIABLES: [GITHUB_TOKEN]` — identical to the fix already merged in `go-template#63`. |
| COPYPASTE | `jscpd` | 7 | 7 clones vs a **0% threshold**; 6 in `vsce/` (self-contained TS extension), 1 in Go (`pkg/svc/tenant/argocd.go` ↔ `pkg/fsutil/configmanager/talos/manager.go`). Pre-existing — not introduced here. | Policy call: exclude the self-contained TS sub-app(s) from repo-wide jscpd (mirrors the existing `web/ui/` exclusion) + refactor or accept the one 13-line Go clone. |

The same gate blocks the dependabot backlog (#4881–#4890). Parts 1 + 2 are mechanical; part 3 is a maintainer-policy decision (refactor vs. exclude/threshold). I'll prepare the **validated** config fix via a local megalinter run on the next substantive pass. Flagging now so this PR isn't merged expecting the `desktop/.*` exclusion alone to clear the gate.

Issue #4936, opened 2026-05-29
bug: cluster update upgrades Kubernetes to a version incompatible with the pinned Talos version


devantler:
## Summary

With `spec.cluster.talos.version` pinned (to match a pinned Hetzner ISO) but **no Kubernetes version specified**, `ksail cluster update` computes a desired Talos machine config that bumps the Kubernetes component images from the running **v1.32.0** to **v1.36.0**. Talos v1.12.4 rejects this with *"version of Kubernetes 1.36.0 is too new to be used with Talos 1.12.4,"* so the update fails on every control-plane. There is no obvious config field to pin the Kubernetes version.

## Environment

- **ksail v7.18.0** (built 2026-05-22, Git SHA `69d747803cf5e7d076de3235aee92e121f15b019`)
- `ksail.yaml`: `spec.cluster.distribution: Talos`, `spec.cluster.talos.version: v1.12.4`, `iso: <pinned>`; **no Kubernetes version field set**.
- Provider: Hetzner. Running Kubernetes: **v1.32.0**.

## Steps to reproduce

1. ksail config with Talos pinned to `v1.12.4` and no Kubernetes version.
2. `ksail --config <cfg> cluster update` (the intended change was unrelated — adding a worker).

## Expected

Either (or both) of:

- When `talos.version` is pinned, the desired Kubernetes version defaults to one **compatible with that Talos release** (or to the currently-running version) — yielding a no-op machine-config diff rather than an unrequested upgrade.
- A **documented config field** to pin the Kubernetes version (e.g. `spec.cluster.kubernetesVersion` or `spec.cluster.talos.kubernetesVersion`), analogous to the Talos version pin. The DR-runbook guidance *"bump the Kubernetes version in the ksail config"* implies such a field exists, but it is not discoverable via the CLI flags (`cluster create/update --help`), the generated config, `ksail init`, or the docs site.

## Actual

machine.config diff (from the change summary):

```
kubelet:                 v1.32.0 -> v1.36.0
kube-apiserver:          v1.32.0 -> v1.36.0
kube-controller-manager: v1.32.0 -> v1.36.0
kube-scheduler:          v1.32.0 -> v1.36.0
```

Apply result on each control-plane:

```
* v1alpha1.Config: 4 errors occurred:
* kubelet image is not valid: version of Kubernetes 1.36.0 is too new to be used with Talos 1.12.4
* kube-apiserver image is not valid: ... too new ...
[...]

Issue #4935, opened 2026-05-29
bug: cluster update exits 0 when changes fail to apply


devantler:
## Summary

`ksail cluster update` prints **`✗ N changes failed to apply`** but exits with status code **`0`**. Automation that gates on the exit code treats a fully-failed update as success.

## Environment

- **ksail v7.18.0** (built 2026-05-22, Git SHA `69d747803cf5e7d076de3235aee92e121f15b019`)
- Talos v1.12.4 on Hetzner, Kubernetes v1.32.0, macOS.

## Steps to reproduce

1. Trigger a `cluster update` whose apply fails (e.g. an invalid Talos machine config — one easy trigger is a Kubernetes version too new for the pinned Talos; see the companion report on the Kubernetes-version-vs-pinned-Talos bug).
2. Inspect `$?`.

## Expected

Non-zero exit status when one or more changes fail to apply.

## Actual

```
✗ 7 changes failed to apply:
    - talos.config: apply control-plane config: ... rpc error: code = InvalidArgument ...
    - talos.config: build desired config: align secrets for config comparison: failed to create config bundle: failed to parse PEM block
    ...
```

…yet `echo $?` → **`0`**.

## Impact

Retry loops, CI, and GitOps automation that gate on the exit code will believe a failed update succeeded. Concretely: a `until ksail ... cluster update; do ...; done` retry loop reported **success on a fully-failed apply** because ksail exited 0 while reporting 7 failed changes.

Related prior work: **#4868** (`cluster update` warns instead of failing when the cluster is unreachable) — same theme of `cluster update` not surfacing failure through its exit behavior.

## Notes

Reported together with two companion ksail bugs found in the same session: **"`cluster update --dry-run` applies changes instead of previewing"** and **"`cluster update` upgrades Kubernetes to a version incompatible with the pinned Talos version."**

---
> 🤖 Filed via Claude Code at the maintainer's request, from a firsthand `ksail cluster update` session against a prod Talos + Hetzner cluster (ksail v7.18.0).

PR #4941, opened 2026-05-29
fix(update): honor --dry-run when combined with --config

### LINKED ISSUE #4934: bug: cluster update --dry-run applies changes instead of only previewing
## Summary

`ksail cluster update --dry-run` does **not** behave as a preview. After printing the change summary it proceeds into **`🔄 Applying changes...`** and performs real mutations against the cluster — per-node Talos `apply config` RPCs and creation of the cluster-autoscaler config `Secret`. The flag is documented as *"Preview changes without applying them."*

## Environment

- **ksail v7.18.0** (built 2026-05-22T13:36:53Z, Git SHA `69d747803cf5e7d076de3235aee92e121f15b019`)
- Distribution / provider: **Talos v1.12.4 on Hetzner**
- Running Kubernetes: **v1.32.0**
- Host: macOS (darwin, arm64)
- Invocation: `ksail --config ksail.prod.yaml cluster update --dry-run --output json`

## Steps to reproduce

1. Have a running Talos + Hetzner cluster managed by ksail.
2. Run `ksail --config <cfg> cluster update --dry-run --output json`.

## Expected

Only the change summary / diff is printed. **No** mutations: no Talos `apply config` calls to nodes, no `Secret` creation.

## Actual

ksail prints the JSON change summary…

```json
{ "totalChanges": 1, "inPlaceChanges": [ { "field": "machine.config", "category": "in-place", ... } ], "requiresConfirmation": false }
```

…and then continues into a **real apply**:

```
🔄 Applying changes...
  ✓ Synced cluster secrets and endpoint (<cp-ip>) from <cp-ip>
  ⚠ Config update failed on <cp-ip> (control-plane): apply control-plane config: ... rpc error: code = InvalidArgument ...
  ...
Applying cluster-autoscaler config secret...
  ✓ Cluster autoscaler config secret created      <-- real mutation
✗ 7 changes failed to apply:
  ...
```

The `✓ Cluster autoscaler config secret created` line, plus the per-node Talos `apply config` RPCs returning real `InvalidArgument` / gRPC errors *from the nodes*, confirm ksail contacted and attempted to mutate live infrastructure despite `--dry-run`.

## Impact

`--dry-run` cannot be used to safely preview changes — operators relying on it will unintentionally mutate production. In this instance the node applies failed validation (so no node changed), but the autoscaler `Secret` **was** (re)created, and a valid change set would have applied for real.

## Notes / hypothesis

Possibly an interaction with `--output json`, or `--dry-run` only gates the change-detection/summary phase and not the apply phase. Reported together with two companion ksail bugs found in the same `cluster update` session: **"`cluster update` exits 0 when changes fail to apply"** and **"`cluster update` upgrades Kubernetes to a version incompatible with the pinned Talos version."**

---
> 🤖 Filed via Claude Code at the maintainer's request, from a firsthand `ksail cluster update` session against a prod Talos + Hetzner cluster (ksail v7.18.0).

devantler:
## Summary

Fixes #4934 — `ksail cluster update --dry-run` performed real mutations (per-node Talos `apply config` RPCs, autoscaler `Secret` creation) instead of only previewing, when invoked together with the global `--config` flag (the issue's exact invocation: `ksail --config ksail.prod.yaml cluster update --dry-run --output json`).

## Root cause

The `cluster update` dry-run gate was already correct:

```go
// pkg/cli/cmd/cluster/cluster.go — applyOrReportChanges
dryRun := cfgManager.Viper.GetBool("dry-run")
if dryRun {
    return reportDryRun(cmd, diff) // early return, no apply
}
```

…but `Viper.GetBool("dry-run")` returned `false` despite `--dry-run` being passed, **only when `--config` was also set**.

When `--config <file>` is used, the config manager resolves the path lazily during `Load()` and then **replaced** its Viper instance with a freshly-constructed one:

```go
// pkg/fsutil/configmanager/ksail/manager.go — resolveConfigFile (before)
v, _ := InitializeViper(cfgPath)
m.Viper = v // ← discards every BindPFlag binding registered at command construction
```

The fresh instance has no pflag bindings, so behavioral flags read directly via `cfgManager.Viper.GetBool(...)` — `--dry-run`, `--force`, `--update-kubernetes`, `--update-distribution` — silently reverted to their defaults whenever `--config` was passed. For `--dry-run` that meant the gate was skipped and the update applied for real.

Config-field flags (`--distribution`, `--provider`, `--name`, …) were unaffected: they reach `m.Config` through a separate cobra-direct override path (`captureChangedFlagValues` → `applyFlagOverrides`), not the Viper key.

## Fix

Reconfigure the **existing** Viper instance in place (point it at the explicit config file) instead of replacing it, so the bindings survive:

```go
// resolveConfigFile (after)
err = configureViperConfigSource(m.Viper, cfgPath)
```

`InitializeViper` is refactored to share the new `configureViperConfigSource` helper, leaving its behavior unchanged. `viper.SetConfigFile` still takes precedence over the directory-discovery settings, so the explicit file is used and traversal is skipped exactly as before.

This also restores `--force`, `--update-kubernetes`, and `--update-distribution` when combined with `--config`.

## Testing

- Added `TestExplicitConfigFlagPreservesBoundFlags` (`pkg/fsutil/configmanager/ksail/manager_test.go`) — fails before the fix, passes after; asserts a `BindPFlag`-bound flag is still readable after `Load` resolves an explicit `--config` file.
- `go build ./...` — clean.
- `go test ./...` — green (the only failures are the pre-existing env-flaky `pkg/cli/cmd/chat` exec/diagnostic tests that `signal: killed` under full-parallel load; they pass in isolation and are untouched by this change).
- `golangci-lint run --new-from-rev=HEAD` on the changed package — 0 issues.

## Scope

Targeted fix for #4934 only. The companion reports #4935 (`cluster update` exits 0 on a failed apply) and #4936 (Kubernetes upgraded to a version incompatible with the pinned Talos version) are separate issues and out of scope here.

### DIFF
--- a/pkg/fsutil/configmanager/ksail/manager.go
+++ b/pkg/fsutil/configmanager/ksail/manager.go
@@ -266,14 +266,17 @@ func (m *ConfigManager) resolveConfigFile() error {
 
 	m.ConfigFile = cfgPath
 
-	// Re-initialize Viper with the explicit config file path so it
-	// skips directory traversal and uses the exact file.
-	v, viperErr := InitializeViper(cfgPath)
-	if viperErr != nil {
-		return fmt.Errorf("invalid --config path: %w", viperErr)
+	// Point the existing Viper at the explicit config file so it skips directory
+	// traversal and uses the exact file. Reconfigure in place rather than
+	// replacing the instance, so that flag bindings registered via BindPFlag at
+	// command construction (e.g. --dry-run, --force, --update-kubernetes) are
+	// preserved. Replacing the instance silently dropped those bindings, making
+	// such flags revert to their defaults whenever --config was passed — e.g.
+	// `cluster update --config <file> --dry-run` applied changes for real (#4934).
+	err = configureViperConfigSource(m.Viper, cfgPath)
+	if err != nil {
+		return fmt.Errorf("invalid --config path: %w", err)
 	}
-
-	m.Viper = v
 
 	return nil
 }
--- a/pkg/fsutil/configmanager/ksail/manager_test.go
+++ b/pkg/fsutil/configmanager/ksail/manager_test.go
@@ -1449,3 +1449,71 @@ func TestLazyConfigFlagResolution_DefaultDiscovery(t *testing.T) {
 
 	require.NoError(t, root.Execute())
 }
+
+// TestExplicitConfigFlagPreservesBoundFlags verifies that flag bindings registered
+// via BindPFlag at command-construction time survive the lazy --config resolution
+// performed during Load. Behavioral flags such as --dry-run are read directly via
+// cfgManager.Viper.GetBool(...); if resolving --config replaces the Viper instance
+// (dropping its bindings), those flags silently revert to their defaults.
+//
+// Regression test for https://github.com/devantler-tech/ksail/issues/4934:
+// `cluster update --dry-run` applied changes instead of previewing them when
+// combined with the global --config flag, because --dry-run read false from a
+// freshly-constructed (binding-less) Viper.
+//
+//nolint:paralleltest // Uses t.Chdir to isolate file system state for config loading.
+func TestExplicitConfigFlagPreservesBoundFlags(t *testing.T) {
+	tempDir := t.TempDir()
+	t.Chdir(tempDir)
+
+	workloadDir := filepath.Join(tempDir, "k8s")
+	require.NoError(t, os.MkdirAll(workloadDir, 0o750))
+
+	kindCfg := "apiVersion: kind.x-k8s.io/v1alpha4\nkind: Cluster\nname: dryrun-test\n"
+	kindPath := filepath.Join(tempDir, "kind.yaml")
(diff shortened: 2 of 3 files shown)


devantler:
The 📊 Benchmark check itself **passed** — the performance-alert comment above is a non-gating noisy-runner false positive, not a regression introduced by this PR:

- **Allocations are byte-identical** across every flagged benchmark (e.g. `Cluster_MarshalJSON/WithBasicConfig`: `18581` vs `18579` B/op, `448` vs `448` allocs/op; `YAMLEncode/FullProductionCluster`: `36432` vs `36432` B/op). A real regression moves allocations — here only wall-clock `ns/op` shifted (~1.5–2.6× uniformly), the classic signature of a loaded/slow runner.
- **The flagged benchmarks live in `pkg/apis/cluster/v1alpha1` and `pkg/svc/image`**, neither of which imports the config manager. This PR only touches `pkg/fsutil/configmanager/ksail/` (Viper flag-binding on `--config`), so there is no code path by which it could affect JSON/YAML marshaling or image-manifest parsing.

No action needed.

PR #4982, opened 2026-06-01
test(uiserver): cover Listen loopback boundary and NewServer wiring

devantler:

## What

The `pkg/cli/uiserver` package had **no tests**. This adds black-box tests for its exported surface, taking package coverage from **0% → 45.1%**.

`uiserver.Listen` is a security-relevant critical path: it enforces the **loopback-only** bind that is the boundary for the unauthenticated local UI/REST API (documented on `uiserver.Host`). `NewServer` wires the read-write local server.

## Tests added

- **`Listen(port 0)`** picks a kernel-chosen free port and advertises the *actually bound* port in the returned URL (the contract callers rely on to open a browser).
- **Loopback boundary** — the listener binds to loopback only (`127.0.0.1`), never a routable address.
- **Error paths** — an out-of-range port (`99999`) and an already-bound port both surface an error (wrapped with the bind address), rather than silently succeeding.
- **`NewServer` wiring** — pins the security-relevant `ReadOnly=false` default, a non-empty creatable-distributions list, the cluster service, and the embedded SPA assets.

The remaining uncovered code is `settings.go`'s OS-secure-store credential paths, which need a `credentials.Manager`/store fake — out of scope for this focused PR (a sensible follow-up).

## Validation

- `go test ./pkg/cli/uiserver/...` — pass; `-count=5` — no flakiness
- `go vet ./pkg/cli/uiserver/...` — clean
- `golangci-lint run ./pkg/cli/uiserver/...` — 0 issues

Behavior-preserving: test-only change, no production code touched.

### DIFF
--- a/pkg/cli/uiserver/uiserver_test.go
+++ b/pkg/cli/uiserver/uiserver_test.go
@@ -1,0 +1,116 @@
+package uiserver_test
+
+import (
+	"context"
+	"net"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/devantler-tech/ksail/v7/pkg/cli/uiserver"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestListenPicksFreePort verifies that port 0 binds to a kernel-chosen free port and that the
+// returned URL reflects the actually bound port (the contract callers rely on to open a browser).
+func TestListenPicksFreePort(t *testing.T) {
+	t.Parallel()
+
+	listener, url, err := uiserver.Listen(context.Background(), 0)
+	require.NoError(t, err)
+
+	t.Cleanup(func() { _ = listener.Close() })
+
+	_, boundPort, splitErr := net.SplitHostPort(listener.Addr().String())
+	require.NoError(t, splitErr)
+
+	port, convErr := strconv.Atoi(boundPort)
+	require.NoError(t, convErr)
+	assert.Positive(t, port, "port 0 must resolve to a concrete kernel-assigned port")
+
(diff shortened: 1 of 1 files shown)


devantler:

Root-caused the persistent `🛡️ Vulnerability Scan` failure on this PR — it is **not** a real finding and **not** transient: `govulncheck ./...` (`-scan symbol`) builds a whole-program call graph whose peak memory exhausts the hosted runner's RAM on this large module (k8s/Flux/Talos/Omni clients), so the host kills the runner mid-scan (`exit 143` / *"runner has received a shutdown signal"* / *"operation was canceled"*, ~2 min in, zero govulncheck output). **Re-running just reproduces it** — please don't keep re-triggering.

Fix is upstream in the reusable workflow: **devantler-tech/reusable-workflows#270** (caps `GOMEMLIMIT` + adds a `timeout-minutes`) / details in **devantler-tech/reusable-workflows#269**. The failing job here resolves the workflow `@main`, so this PR's scan should pass once #270 lands on `reusable-workflows` `main`. Holding this PR until then.

devantler:

**Unblock path is now concrete.** The `🛡️ Vulnerability Scan` blocking this PR fails on 4 reachable-but-unpatched advisories (GO-2026-4887, GO-2026-4883 in docker/docker; GO-2025-3547, GO-2025-3521 in k8s.io/kubernetes — all `Fixed in: N/A`). reusable-workflows **v5.4.0** (released today) added the `.govulncheck-allow.txt` allowlist mechanism, and **[#4985](https://github.com/devantler-tech/ksail/pull/4985)** risk-accepts exactly those 4. Once #4985 merges, **rebase this PR onto `main`** to pick up the allowlist file and the scan goes green — no code change needed here. (Don't re-trigger the scan before then; it can't pass without the file.)

PR #5068, opened 2026-06-05
refactor(notify): drop unused gosec nolint directive

devantler:
## What

Removes the `//nolint:gosec` directive on the `uintptr`→`int` file-descriptor conversion in the progress writer (`pkg/notify/progress.go`). `nolintlint` reports the directive as **unused** — `gosec` no longer flags the `term.IsTerminal` usage, so the suppression is dead.

## Why

Part of a small series driving the `golangci-lint` baseline toward zero. Dead lint-suppression directives are noise: they imply a finding that no longer exists and can mask a real one if the surrounding code changes.

## Validation

- `go build ./pkg/notify/...` ✓
- `go test ./pkg/notify/...` — 208 passing ✓
- `golangci-lint run --new-from-rev=origin/main ./pkg/notify/...` — 0 issues ✓

Behavior is unchanged (single-line deletion of a comment directive).

### DIFF
--- a/pkg/notify/progress.go
+++ b/pkg/notify/progress.go
@@ -283,7 +283,6 @@ func NewProgressGroup(
 	termWidth := 0
 
 	if file, ok := writer.(*os.File); ok {
-		//nolint:gosec // uintptr to int conversion is safe for term.IsTerminal usage
 		fd := int(file.Fd())
 		isTTY = term.IsTerminal(fd)
 


devantler:
Closing — this change is invalid. CI's `gosec` fires `G115: integer overflow conversion uintptr -> int` on `int(file.Fd())` (progress.go:286), so the `//nolint:gosec` directive is **required** in CI and is not actually dead.

The `nolintlint` "directive is unused" finding that motivated this PR is a **local golangci-lint version skew**: golangci-lint v2.12.2 bundles a `gosec` that no longer fires G115 on uintptr→int, so its `nolintlint` flags the directive as unused — but CI's golangci-lint still fires G115, where the directive is doing its job. There's no single edit that satisfies both (local nolintlint wants it removed; CI gosec needs it), and CI is authoritative. The original code is correct.

(Minor follow-up for maintainer: aligning the local/pinned golangci-lint version with CI would remove this false local nolintlint signal.)

PR #5078, opened 2026-06-05
fix(hetzner): restart cluster-autoscaler when its config secret changes on update

devantler:
## Problem

On a Talos + Hetzner cluster with the KSail-managed Cluster Autoscaler, bumping the Kubernetes version (`spec.cluster.kubernetesVersion`) and running `ksail cluster update` upgrades the static nodes but leaves **autoscaler-provisioned nodes on the old Kubernetes version** (and old Talos snapshot).

## Root cause

`cluster update` does regenerate the `cluster-autoscaler-config` Secret with the new config:

- `hcloud_image` — the Hetzner snapshot image ID
- `hcloud_cloud_init` — the gzipped Talos worker machine config, whose `machine.kubelet.image` sets the new node's Kubernetes version

…but it never restarts the `cluster-autoscaler` Deployment. The autoscaler consumes both keys as **environment variables** via `valueFrom.secretKeyRef` (the chart's `extraEnvSecrets`), and Kubernetes does **not** live-reload env vars from a Secret. The running autoscaler therefore keeps provisioning nodes from the config it loaded at pod start — the stale Kubernetes version / snapshot — until something else restarts it.

This is autoscaler-specific: static nodes are upgraded in place, but the autoscaler is a long-running controller holding a cached copy of the config.

## Fix

- `ApplyAutoscalerConfigSecret` now reports whether the Secret's data was **created or changed**.
- On `cluster update`, when the Secret changed, KSail rolls the `cluster-autoscaler` Deployment so it reloads the env-from-secret — a `kubectl rollout restart`-style pod-template annotation, targeting the Deployment by its `app.kubernetes.io/instance=cluster-autoscaler` label (robust to the chart's fullname template).
- Gated on *changed*, so an unrelated `cluster update` (e.g. a firewall sync) never needlessly bounces the autoscaler.
- **Cluster create is unaffected**: the autoscaler installer runs after the Secret is written, so its pod starts with the correct config (the create path passes `restartIfChanged=false`).
- A missing Deployment is **not** an error — the autoscaler may not be installed yet.
- Adds a reusable `k8s.RolloutRestartDeploymentsByLabel` helper.

## Test plan

- `go build ./...` ✓
- `go vet ./pkg/k8s/... ./pkg/svc/provisioner/cluster/talos/...` ✓
- `golangci-lint run` — no new issues in changed files ✓
- `go test ./pkg/k8s/... ./pkg/svc/provisioner/cluster/talos/...` ✓
- New unit tests:
  - `k8s.RolloutRestartDeploymentsByLabel`: stamps the restart annotation (preserving existing pod-template annotations), no-match-is-not-an-error, list-error surfaced.
  - `ApplyAutoscalerConfigSecret` returns `changed=true` on create and `changed=false` when re-applying identical config.

## Manual verification suggestion

On a live Hetzner cluster: bump `spec.cluster.kubernetesVersion`, `ksail cluster update`, then confirm the `cluster-autoscaler` pod restarted and a freshly scaled-up node joins at the new version (no longer requires a manual `kubectl rollout restart deployment/cluster-autoscaler -n kube-system`).

### DIFF
--- a/pkg/k8s/deployment.go
+++ b/pkg/k8s/deployment.go
@@ -1,0 +1,89 @@
+package k8s
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+// RolloutRestartAnnotation is the pod-template annotation that triggers a rolling
+// restart, matching the key `kubectl rollout restart` stamps. Changing its value
+// mutates the pod template, so the Deployment controller recreates the pods — the
+// only way pods that source environment variables from a Secret or ConfigMap pick
+// up data changes, since Kubernetes does not live-reload env vars.
+const RolloutRestartAnnotation = "kubectl.kubernetes.io/restartedAt"
+
+// RolloutRestartDeploymentsByLabel triggers a rolling restart of every Deployment
+// in namespace whose labels match labelSelector, the same way
+// `kubectl rollout restart deployment -l <selector>` does: it stamps each pod
+// template's RolloutRestartAnnotation with the current time.
+//
+// Callers that rewrite a Secret or ConfigMap consumed via env-var valueFrom must
+// call this for the change to reach the running pods. A selector that matches no
+// Deployment is not an error (returns 0) — the workload may not be installed yet.
+// Returns the number of Deployments restarted.
+func RolloutRestartDeploymentsByLabel(
+	ctx context.Context,
+	client kubernetes.Interface,
+	namespace string,
+	labelSelector string,
+) (int, error) {
+	deploymentsClient := client.AppsV1().Deployments(namespace)
+
+	list, err := deploymentsClient.List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	if err != nil {
+		return 0, fmt.Errorf("listing deployments %q in %s: %w", labelSelector, namespace, err)
+	}
+
+	timestamp := time.Now().Format(time.RFC3339)
+	restarted := 0
+
+	for index := range list.Items {
+		err = restartDeployment(ctx, deploymentsClient, list.Items[index].Name, timestamp)
+		if err != nil {
+			return restarted, err
+		}
+
+		restarted++
+	}
+
+	return restarted, nil
+}
+
+// restartDeployment stamps a single Deployment's pod-template restart annotation,
+// retrying on optimistic-concurrency conflicts.
+func restartDeployment(
+	ctx context.Context,
+	deploymentsClient appsv1client.DeploymentInterface,
+	name, timestamp string,
+) error {
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		deployment, getErr := deploymentsClient.Get(ctx, name, metav1.GetOptions{})
+		if getErr != nil {
+			return fmt.Errorf("getting deployment %s: %w", name, getErr)
+		}
+
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = map[string]string{}
+		}
+
+		deployment.Spec.Template.Annotations[RolloutRestartAnnotation] = timestamp
+
+		_, updateErr := deploymentsClient.Update(ctx, deployment, metav1.UpdateOptions{})
+		if updateErr != nil {
+			return fmt.Errorf("restarting deployment %s: %w", name, updateErr)
+		}
+
+		return nil
+	})
+	if retryErr != nil {
+		return fmt.Errorf("rolling restart of deployment %s: %w", name, retryErr)
+	}
+
+	return nil
+}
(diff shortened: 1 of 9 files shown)


devantler:
Investigate if there is a more graceful way to do this, that does not require a full restart.

devantler:
I looked into whether the running `cluster-autoscaler` can pick up the regenerated config **without** a pod restart. **Short answer: it can't — a rolling restart is the only mechanism the Hetzner provider allows, and that's exactly what this PR already does (gated on an actual change).**

## Why there is no live-reload

The Hetzner cloud provider reads all of its node-provisioning config — `HCLOUD_IMAGE`, `HCLOUD_CLOUD_INIT`, and the newer `HCLOUD_CLUSTER_CONFIG` / `HCLOUD_CLUSTER_CONFIG_FILE` — **exactly once, at process startup in `newManager()`**, and caches the parsed result in the `hetznerManager` struct for every subsequent scale-up. There is no reload path:

- `newManager()` calls `os.Getenv("HCLOUD_CLUSTER_CONFIG")` / `HCLOUD_CLOUD_INIT` / `HCLOUD_IMAGE` once and stores the result ([hetzner_manager.go @ cluster-autoscaler-1.32.0, ~L95–L133](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.32.0/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go)).
- `Refresh()` — the per-loop hook where a provider *could* re-read config — is a no-op: `func (m *hetznerManager) Refresh() error { return nil }` (same file, ~L156–158).
- This is **identical on `master`**, so bumping the chart wouldn't change it.

Switching the *delivery* mechanism doesn't help either:

- **Mounting the Secret as a volume** (so the kubelet updates the file in place) changes nothing — the provider only reads the file once in `newManager()` and never re-reads it.
- **`HCLOUD_CLUSTER_CONFIG_FILE`** is documented as merely "an alternative to `HCLOUD_CLUSTER_CONFIG` … the path to a file containing the JSON structure described above" — no file-watching, no runtime reload, functionally equivalent to the env var ([hetzner README @ master](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/hetzner/README.md)).

So there's no SIGHUP, no file-watch, and no `Refresh()` re-read: the cached config lives for the lifetime of the process. The upstream-recommended way to apply a config change is precisely `kubectl rollout restart deployment/cluster-autoscaler` — which is what this PR automates.

## The current restart is already the graceful form

- It stamps the `kubectl.kubernetes.io/restartedAt` pod-template annotation → the Deployment performs a **rolling** restart, not a hard pod kill. With HA (`replicas=2`) + the autoscaler's default leader election, the standby keeps leading and there's effectively **zero autoscaling gap**; a single-replica install has only a brief few-second gap while the replacement pod starts and acquires the lease.
- It's **gated on an actual Secret data change**, so an unrelated `cluster update` (e.g. a firewall sync) never bounces the autoscaler.
- A restart is **safe for in-flight scale-ups**: the autoscaler is stateless and rebuilds its view from the cloud + cluster on start; Hetzner servers already being created continue and get reconciled.

Autoscaling isn't latency-critical to the second, so a ~seconds rolling restart *only when the config actually changed* is about as gentle as it gets.

## Options if we want to go further (none avoid the restart)

1. **Upstream feature** — the only true "no-restart" fix is teaching the Hetzner provider to re-read `HCLOUD_CLUSTER_CONFIG_FILE` on `Refresh()` (or watch the file). That's a change in `kubernetes/autoscaler`, not ksail; until then a restart is unavoidable. I can open an upstream issue if you'd like.
2. **Declarative checksum annotation** (Stakater-Reloader style) — hash the Secret content into the pod template so a `helm upgrade` rolls the Deployment automatically. More declarative, but it's the **same rolling restart** (same disruption) and a bigger restructure, since the Secret is owned by the provisioner, outside the chart.
3. **Minor refinement to this PR** — `restartDeployment` could stamp the annotation with a single strategic-merge **patch** instead of get-modify-`Update` + `RetryOnConflict`, matching how `kubectl rollout restart` itself does it (atomic, no conflict loop). Pure code-cleanliness, not a gracefulness change.

## Recommendation

Keep the gated rolling restart — it is the least-disruptive option the provider supports, and it mirrors the upstream-recommended remediation. I'm happy to apply option 3 to tidy the helper, and/or file the upstream issue for option 1 if you want to track a real no-restart path.

— 🤖 Investigated with [Claude Code](https://claude.com/claude-code)

devantler:
Applied **option 3** in ff20e420: `RolloutRestartDeploymentsByLabel` now stamps the `restartedAt` annotation with a single atomic **strategic-merge patch** (the exact mutation `kubectl rollout restart` performs) instead of the get-modify-`Update` + `RetryOnConflict` loop. The patch merges into existing pod-template annotations (preserving them) and is conflict-free, so the retry is gone.

Behaviour is unchanged — the existing tests (annotation stamped, other annotations preserved, no-match-not-an-error, list-error surfaced) all pass, plus `go build` / `go vet` / `golangci-lint --new-from-rev` are clean. Net −34/+40 but functionally simpler.

The restart itself is unchanged, so the [earlier conclusion](https://github.com/devantler-tech/ksail/pull/5078#issuecomment-4637902023) stands: a rolling restart remains the only way the Hetzner provider picks up changed config.

PR #5097, opened 2026-06-06
feat: warn and hint `--update-distribution` on ignored Talos version pin

### LINKED ISSUE #5077: bug: `cluster update --update-distribution` strips Talos system extensions on Hetzner (upgrades to bare `ghcr.io/siderolabs/installer`, ignoring the schematic)
## Summary

On a **Talos × Hetzner** cluster whose nodes carry Image Factory **system extensions**, `ksail cluster update --update-distribution` performs the rolling Talos OS upgrade with the **bare upstream installer** `ghcr.io/siderolabs/installer:<tag>` — it ignores the configured schematic / `spec.cluster.talos.extensions`. Because a Talos OS upgrade replaces the boot assets, upgrading to an installer that lacks the extensions **strips them from every node**. For a cluster that depends on e.g. `siderolabs/iscsi-tools` (Longhorn data plane) and `siderolabs/qemu-guest-agent` (Hetzner Cloud integration), this is a destructive, data-plane-breaking upgrade.

This is inconsistent with `cluster create`, the Hetzner snapshot path, and the autoscaler, all of which **do** build a Factory image from the schematic. Only the upgrade path drops it.

A second, related problem (details below): a plain `cluster update` silently ignores a changed `spec.cluster.talos.version` pin — it reports the version diff but takes no action and gives no hint that `--update-distribution` is required.

## Environment

- **ksail v7.26.0** (`fede6e18340ef7c94991df23a67d9711206bf8df`)
- Provider: **Hetzner**, Distribution: **Talos**
- `spec.cluster.talos.version: v1.13.3`, `spec.cluster.talos.extensions: [iscsi-tools, util-linux-tools, qemu-guest-agent]`
- Nodes installed from the Factory installer `factory.talos.dev/installer/<schematic>:v1.13.3`
- Encountered in [`devantler-tech/platform`](https://github.com/devantler-tech/platform) (prod cluster).

---

### Code path

`UpgradeDistribution` builds the installer image purely from the version tag:

```go
// pkg/svc/provisioner/cluster/talos/upgrader.go:41
installerImage := installerImageFromTag(toVersion)
err := p.rollingUpgradeNodes(ctx, clusterName, installerImage, toVersion)
```

```go
// pkg/svc/provisioner/cluster/talos/utils.go:84-91
const installerImageRepository = "ghcr.io/siderolabs/installer"

func installerImageFromTag(tag string) string {
	return installerImageRepository + ":" + tag   // -> ghcr.io/siderolabs/installer:v1.13.3 (NO schematic)
}
```

`rollingUpgradeNodes` → `upgradeNodeTalosVersion` then `ImageService.Pull` + `LifecycleService.Upgrade` that bare image and reboots each node. The configured schematic is never consulted.

### Contrast — create/snapshot/autoscaler ARE schematic-aware

`ensureSnapshotImage` (provisioner_hetzner.go) and `detachOrWaitForReboot` resolve `p.talosOpts.SchematicID` (or auto-compute it from `extensions` via `p.talosConfigs.SchematicID()`) and boot new nodes from a Factory snapshot built from `version` + `schematicID`. The autoscaler even hard-requires a schematic (`ErrAutoscalerRequiresSchematic`). The upgrade path is the one place that drops this.

### Expected

`UpgradeDistribution` should upgrade to the **same Factory image** the rest of the lifecycle uses — i.e. `factory.talos.dev/installer/<schematicID>:<tag>` when a schematic is configured (explicit `schematicId` or auto-computed from `extensions`), falling back to `ghcr.io/siderolabs/installer:<tag>` only when no schematic is set. That preserves extensions across upgrades and matches create/snapshot/autoscaler behavior.

### Actual

Nodes are rolled onto `ghcr.io/siderolabs/installer:<tag>`; on reboot the Factory extensions (iscsi-tools, util-linux-tools, qemu-guest-agent) are gone → Longhorn iSCSI volumes fail to attach and the Hetzner guest agent disappears.

### Notes

- There is **no escape hatch**: `cluster update` exposes no `--installer-image` / schematic override flag (only `force`, `yes`, `dry-run`, `output`, `update-kubernetes`, `update-distribution`), so a user cannot work around it from the CLI.
- Verified present on `origin/main` as well (no commits to `upgrade.go`/`upgrader.go`/`utils.go` since v7.26.0).

---

## Problem 2 (UX / consistency): plain `cluster update` silently ignores a changed `talos.version` pin

When `spec.cluster.talos.version` is changed (e.g. v1.12.4 → v1.13.3) and the user runs a plain `ksail cluster update` (no `--update-distribution`):

- The diff engine reports `cluster.talos.version` as a change but classifies it `ChangeCategoryInPlace` with reason *"version pin change only affects future operations (image selection, upgrade cap)"* (`pkg/svc/diff/engine.go`), so **nothing is applied** to existing nodes.
- `handleVersionUpgrades` only runs when `--update-kubernetes || --update-distribution` is set (`pkg/cli/cmd/cluster/update.go:163`), so the OS upgrade is skipped.
- The user is given **no hint** that `--update-distribution` is required to act on the pin they just changed.

This is the intended opt-in behavior from #4260 (don't auto-upgrade when the version is *unset*) — but here the version is **explicitly pinned and differs from the running version**, and the result is a confusing no-op. It's also **inconsistent with `spec.cluster.kubernetesVersion`**, which *is* applied by a plain `cluster update` (re-rendered into the machine config — see #4936).

### Suggested

When a plain `cluster update` detects that the pinned `spec.cluster.talos.version` differs from the running version, emit an actionable hint, e.g.:

```
Talos version pin (v1.13.3) differs from the running version (v1.12.4).
Run `ksail cluster update --update-distribution` to roll the OS upgrade.
```

(Could be split into its own enhancement issue if preferred.)

---

## Steps to reproduce (Problem 1)

1. Create a Talos × Hetzner cluster with `spec.cluster.talos.extensions: [siderolabs/iscsi-tools, ...]` and a pinned `spec.cluster.talos.version`.
2. Bump `spec.cluster.talos.version` to a newer tag.
3. `ksail --config <cfg> cluster update --update-distribution`.
4. Observe each node is upgraded with `ghcr.io/siderolabs/installer:<newtag>` (see the "Pulling installer image …" log line) — not the Factory image.
5. After reboot, `talosctl get extensions` on a node shows the extensions are gone.

## Impact

Any Talos × Hetzner user with extensions (which is **required** for Longhorn and recommended for Hetzner) cannot use `ksail cluster update --update-distribution` to upgrade Talos without breaking their storage/data plane. The documented "bump version + cluster update" upgrade flow is therefore unusable for these clusters; the only safe path today is a manual `talosctl upgrade --image factory.talos.dev/installer/<schematic>:<tag>` per node.

---
> 🤖 Investigated and filed with Claude Code (Opus 4.8) at the maintainer's request.

### LINKED ISSUE #5094: feat: warn (and hint `--update-distribution`) when `cluster update` detects a changed `spec.cluster.talos.version` pin it won't act on

## Problem

Split out from the now-closed #5077 (problem 1 — extension-stripping on `--update-distribution` — was fixed in #5079; this is the deferred **problem 2**).

A plain `ksail cluster update` (without `--update-distribution`) **silently ignores a changed `spec.cluster.talos.version` pin**. It computes and reports the version diff in its plan/output, but takes **no action** on the Talos OS version and gives **no hint** that the distribution upgrade is gated behind `--update-distribution`.

For an operator who bumps `spec.cluster.talos.version` in `ksail.yaml` and runs `cluster update` expecting the cluster to move to the pinned version, the command appears to succeed while the nodes stay on the old version — a silent no-op footgun. The information needed to act (the version delta) is already computed; only the user-facing signal is missing.

## Proposed direction

When `cluster update` (Talos provider) detects that the **effective `spec.cluster.talos.version` differs from the running cluster version** but `--update-distribution` was **not** supplied, surface an explicit, actionable message instead of silently skipping it, e.g.:

```
⚠️  spec.cluster.talos.version changed (vX.Y.Z → vA.B.C) but the Talos OS upgrade was skipped.
    Re-run with `--update-distribution` to perform the rolling OS upgrade.
```

Keep it advisory (a warning/hint, not a hard error) so the existing non-distribution update flow is unchanged for everyone who intentionally runs without the flag. No behaviour change to what `update` actually *does* — only the diagnostic when a version-pin delta is detected and not acted on.

## Acceptance criteria

- [ ] `cluster update` (Talos) with a changed `spec.cluster.talos.version` and **no** `--update-distribution` emits a clear warning naming the old→new version and pointing at `--update-distribution`.
- [ ] No warning when the version pin is unchanged, or when `--update-distribution` **is** supplied (that path already performs the upgrade).
- [ ] Exit code / overall update behaviour is unchanged (advisory only).
- [ ] Unit test covering: changed-pin-without-flag → warning; unchanged-pin → no warning.

## Notes

- Discovered while wrapping up the #5077 line of work (problem 1 shipped in #5079). Filing so the deferred UX gap is tracked rather than lost.
- Provider scope: Talos (the `spec.cluster.talos.version` pin is Talos-specific).

devantler:

Fixes #5094 (the deferred *problem 2* split out from the now-closed #5077; problem 1 — extension-stripping on `--update-distribution` — was fixed in #5079).

## Problem

A plain `ksail cluster update` (without `--update-distribution`) **computes and reports** a changed `spec.cluster.talos.version` pin in its diff, but the diff engine classifies the pin change as **in-place** (`pkg/svc/diff/engine.go` — reason: *"version pin change only affects future operations (image selection, upgrade cap)"*). So the running Talos OS is left at its current version: the distribution upgrade only happens via the `--update-distribution` path (`handleVersionUpgrades`).

The trap: an operator bumps `spec.cluster.talos.version`, runs `ksail cluster update`, sees the version change reported as an in-place (🟢) change — implying it was applied — yet **the OS is not upgraded and nothing hints** that the upgrade is gated behind `--update-distribution`.

## Change

In `handleUpdateRunE`, after the diff is displayed, when `--update-distribution` was **not** passed, scan the diff for a changed `cluster.talos.version` field and emit:

- a **warning** that a plain update does not upgrade the running OS (the pin only affects future image selection), and
- an **info hint** to re-run with `--update-distribution`.

Written to `cmd.OutOrStderr()`, so machine-readable `--output json` on stdout is unaffected. No behaviour change otherwise — no config field, flag, schema, or generated file is touched.

```
⚠ Talos version pin changed (v1.8.0 → v1.9.0) but a plain update does not upgrade the running OS; the pin only affects future image selection.
ℹ Run 'ksail cluster update --update-distribution' to upgrade the Talos OS to the pinned version.
```

## Validation

- `go build` ✅
- `go test ./pkg/cli/cmd/cluster/` ✅ (524 passed, incl. the new `TestWarnTalosVersionPinIgnored`: warns on a talos.version change with the version transition + gating flag; silent for unrelated/empty diffs)
- `golangci-lint run --new-from-merge-base=main ./pkg/cli/cmd/cluster/...` ✅ no issues

## Notes / trade-offs

- Detection is on the diff **field path** (`cluster.talos.version`), which only appears for Talos clusters — no distribution-type branching needed.
- Fires for both apply and `--dry-run` (the warning is part of the diff presentation), and is suppressed when `--update-distribution` is set (that path already handles the upgrade).
- Helper is unit-tested via a thin `export_test.go` shim, matching the existing `ExportDisplayChangesSummary` pattern.

### DIFF
--- a/pkg/cli/cmd/cluster/cluster_test.go
+++ b/pkg/cli/cmd/cluster/cluster_test.go
@@ -4379,6 +4379,73 @@ func TestReconcileComponents_UnknownField_Skipped(t *testing.T) {
 	assert.Empty(t, result.FailedChanges, "unknown field should not be recorded as failed")
 }
 
+// TestWarnTalosVersionPinIgnored verifies that a plain `cluster update` warns and
+// hints --update-distribution when it detects a changed Talos version pin it will
+// not act on, and stays silent otherwise (issue #5094).
+func TestWarnTalosVersionPinIgnored(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name      string
+		diff      *clusterupdate.UpdateResult
+		wantWarn  bool
+		wantParts []string
+	}{
+		{
+			name: "talos version pin change warns and hints the gating flag",
+			diff: &clusterupdate.UpdateResult{
+				InPlaceChanges: []clusterupdate.Change{
+					{Field: "cluster.talos.version", OldValue: "v1.8.0", NewValue: "v1.9.0"},
+				},
+			},
+			wantWarn:  true,
+			wantParts: []string{"v1.8.0", "v1.9.0", "--update-distribution"},
+		},
+		{
+			name: "unrelated change stays silent",
+			diff: &clusterupdate.UpdateResult{
+				InPlaceChanges: []clusterupdate.Change{
+					{Field: "cluster.nodes", OldValue: "1", NewValue: "3"},
+				},
+			},
+			wantWarn: false,
+		},
+		{
+			name:     "empty diff stays silent",
+			diff:     &clusterupdate.UpdateResult{},
+			wantWarn: false,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			cmd := newReconcileTestCmd()
+
+			// The helper writes via cmd.OutOrStderr(), which cobra resolves to the
+			// out writer (then os.Stderr). Capture both to one buffer so the test
+			// sees the output regardless of cobra's fallback.
+			var out bytes.Buffer
+			cmd.SetOut(&out)
+			cmd.SetErr(&out)
+
+			cluster.ExportWarnTalosVersionPinIgnored(cmd, testCase.diff)
+
+			got := out.String()
+
+			if testCase.wantWarn {
+				for _, part := range testCase.wantParts {
+					assert.Contains(t, got, part,
+						"warning should mention %q", part)
+				}
+			} else {
+				assert.Empty(t, got, "no warning expected for this diff")
+			}
+		})
+	}
+}
+
 // TestReconcileComponents_RecordsFailedChange verifies that a component error is captured
 // in result.FailedChanges while processing of remaining changes continues.
 //
--- a/pkg/cli/cmd/cluster/export_test.go
+++ b/pkg/cli/cmd/cluster/export_test.go
@@ -151,6 +151,11 @@ func ExportDisplayChangesSummary(cmd *cobra.Command, diff *clusterupdate.UpdateR
 	displayChangesSummary(cmd, diff)
 }
 
+// ExportWarnTalosVersionPinIgnored exports warnTalosVersionPinIgnored for testing.
+func ExportWarnTalosVersionPinIgnored(cmd *cobra.Command, diff *clusterupdate.UpdateResult) {
+	warnTalosVersionPinIgnored(cmd, diff)
+}
+
 // ExportDiffToJSON exports diffToJSON for testing.
 func ExportDiffToJSON(diff *clusterupdate.UpdateResult) DiffJSONOutput {
 	return diffToJSON(diff)
--- a/pkg/cli/cmd/cluster/update.go
+++ b/pkg/cli/cmd/cluster/update.go
@@ -217,8 +217,45 @@ func handleUpdateRunE(
 	// Display changes summary
 	displayChangesSummary(cmd, diff)
 
+	// A plain `cluster update` reports a changed Talos version pin as an in-place
+	// change, but the pin only governs future image selection / the upgrade cap —
+	// it does NOT upgrade the running OS. Without --update-distribution the bump is
+	// effectively a silent no-op, so warn and point at the gating flag.
+	if !updateDist {
+		warnTalosVersionPinIgnored(cmd, diff)
+	}
+
 	return applyOrReportChanges(cmd, cfgManager, ctx, deps, updater,
 		clusterName, currentSpec, diff, outputTimer)
+}
+
+// talosVersionField is the diff field path for the Talos OS version pin
+// (spec.cluster.talos.version), as classified by the diff engine.
+const talosVersionField = "cluster.talos.version"
+
+// warnTalosVersionPinIgnored emits a warning when a plain `cluster update`
+// (without --update-distribution) detects a changed spec.cluster.talos.version
+// pin it will not act on. The diff engine classifies the pin change as in-place
+// because it only affects future image selection and the Kubernetes upgrade cap,
+// so the running Talos OS is left at its current version. The warning makes that
+// silent no-op explicit and hints the gating flag so the operator isn't left
+// believing the OS was upgraded.
+func warnTalosVersionPinIgnored(cmd *cobra.Command, diff *clusterupdate.UpdateResult) {
+	for _, change := range diff.AllChanges() {
+		if change.Field != talosVersionField {
+			continue
+		}
+
+		notify.Warningf(cmd.OutOrStderr(),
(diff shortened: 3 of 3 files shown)


devantler:
We just removed thes flags in favor of auto updating kubernetes and distribution flag with fallback to explecitly setting versions. Please reconsider this PR in that new context.

devantler:

Closing per the maintainer's redirect — reconsidered in the new context.

#5095 (`feat(update): reconcile cluster versions declaratively, removing --update-* flags`, merged 19:09Z) **removed** `--update-distribution`/`--update-kubernetes` and made `ksail cluster update` reconcile the distribution and Kubernetes versions declaratively on **every** run: a *set* `spec.cluster.talos.version` now pins (and upgrades to) that version automatically, with no flag gate.

This obsoletes #5097 on both counts:
- The warning it adds points operators at `--update-distribution`, which **no longer exists**.
- The silent-no-op footgun it was surfacing (a changed `spec.cluster.talos.version` ignored unless the flag was passed) is now **fixed at the root** by #5095 — the pin is reconciled, not skipped.

Closing this PR and its tracking issue #5094 as superseded by #5095. Thanks for the steer.

Issue #4963, opened 2026-05-30
bug: cluster update can't build existing worker config on Hetzner — 'align secrets for config comparison: failed to parse PEM block'


devantler:
## Summary

During `ksail cluster update` on a **Talos × Hetzner** cluster, building the desired machine config for **existing worker** nodes fails at:

```
build desired config: align secrets for config comparison: failed to create config bundle: failed to parse PEM block
```

Control-planes and **newly-created** workers build/apply fine — only the **existing-worker** secret-alignment path fails. As a result, an intended `machine.config` change (here: committed `sysctls` hardening) is applied to the control-planes but **silently skipped on every existing worker**.

## Environment

- ksail **v7.23.1** (Git SHA `8d63e65fd7b3fff1edeabc7681ffc0cc830c7d5f`)
- Talos **v1.12.4**, Kubernetes **v1.32.0**, provider **Hetzner**
- Local Talos creds are healthy: `talosctl config info` → Roles **`os:admin`**, cert valid ~11 months — so this is **not** a missing/limited-credential problem.

## Steps to reproduce

1. Talos × Hetzner cluster with existing workers.
2. Make any `machine.config` change (e.g. add sysctls via a Talos patch) and run `ksail cluster update`.

## Actual

```
✓ Config applied to <cp-1> (control-plane, no reboot)
✓ Config applied to <cp-2> (control-plane, no reboot)
✓ Config applied to <cp-3> (control-plane, no reboot)
⚠ Config update failed on <worker-1> (worker): build desired config: align secrets for config comparison: failed to create config bundle: failed to parse PEM block
⚠ Config update failed on <worker-2> (worker): ... failed to parse PEM block
⚠ Config update failed on <worker-3> (worker): ... failed to parse PEM block
✗ N changes failed to apply
```

(The same run also created a new worker whose config built/applied fine — so the failure is specific to the **existing-worker** "align secrets for config comparison" step, not a globally-broken secret bundle.)

## Expected

Existing-worker desired-config build / secret-alignment succeeds — or, if a secret/PEM really is malformed, the error names *which* secret/field so it's actionable. Either way, in-place `machine.config` changes should reach existing workers, not just control-planes.

## Impact

In-place machine-config changes (security sysctls, kubelet args, etc.) cannot be applied to existing workers via `cluster update`; they silently diverge from the control-planes.

## Notes

[...]

PR #5146, opened 2026-06-08
fix(talos): push worker-scoped machine-config patch drift on cluster update

devantler:
## What

`ksail cluster update` now detects **and pushes worker-scoped Talos machine-config patch changes** (`talos/workers/*.yaml`) to worker nodes. Previously it pushed control-plane and cluster-wide patch changes but silently dropped worker-only ones.

## Why

Drift detection (`detectInPlaceMachineConfigDrift`) only fetched a **control-plane** node and diffed the CP desired-vs-running config. Worker-scoped patches never reach the CP config (`bundle.WithPatchWorker`), so a worker-only change produced **no detected drift** → `TotalChanges == 0` → `shouldApplyInPlaceChanges` was false → the apply path (which is already role-correct and pushes worker configs to workers) never ran.

Reported symptom: *"cluster update pushes count/version/ISO only; a diff in patch files isn't pushed to the correct Talos machines."*

## How

1. **Detect drift on both roles.** Extracted `detectRoleMachineConfigDrift`, added `detectWorkerMachineConfigDrift`, and generalized `fetchRunningControlPlaneConfig` → `fetchRunningConfigForRole`. The CP node covers cluster/control-plane patches; a worker node covers cluster/worker patches. Worker detection is **best-effort** — an unreachable worker is logged and skipped rather than suppressing already-detected control-plane drift. Both roles seed the rebuild from the control-plane PKI (worker configs lack the CA private key, #4963).

2. **Fixed a latent worker version misread.** A *parsed* worker config's `KubernetesVersionFromProvider` resolves its empty `kube-apiserver` image to the vendored Talos **machinery's** bundled default version (e.g. `v1.36.1`) instead of falling back to the real kubelet image (`v1.36.0`). Aligning a worker from its own config therefore caused a phantom kubelet upgrade on every update. `buildDesiredNodeConfig` now aligns the Kubernetes version from `secretsSource` (the control-plane config, which carries the authoritative apiserver image). For the control-plane path `secretsSource == running`, so it's a no-op — no regression.

## Reviewer notes

- Detection and apply stay consistent: both go through `buildDesiredNodeConfig` with the CP config as the secrets/version source, so what's detected is exactly what's pushed (including patch removals).
- Adds one extra Talos API call per update diff (fetch a representative worker config) only when workers exist.
- No public API / schema / docs changes.

## Testing

- 3 new regression tests in `detect_inplace_test.go`:
  - worker-scoped patch is detected at the worker role but **not** the control-plane role (proves the bug + the fix);
  - a clean worker config reports no drift (no false positive);
  - an unpinned update tracks the running worker Kubernetes version (no phantom kubelet upgrade).
- `go build ./...` ✓ · talos package **449 tests pass** ✓ · provisioner + configmanager **1767 tests pass** ✓ · new-code `golangci-lint` clean ✓.


devantler:

The MegaLinter SPELL/lychee failure is a **transient flake, not actionable in this PR**:

- All 4 errors are `500 Internal Server Error` (server-side, not broken links) on valid URLs in **files this PR never touches** — `.github/actions/ksail-cluster/action.yml`, `.github/workflows/cd.yaml`, `.goreleaser.yaml`, `vsce/README.md`. This PR changes only 3 Go files under `pkg/svc/provisioner/cluster/talos/`.
- My latest push (`b5c0e571`) already re-triggered the full pipeline, so `🧹 Lint - mega-linter` is re-running; a transient 500 typically clears on retry.

No code change made — masking a transient upstream 500 by editing unrelated files would be wrong. If lychee keeps 500-flaking on GitHub URLs across PRs, the durable CI-health fix is to make lychee tolerate/retry `5xx` (server errors), which belongs in a separate workflow PR, not here.

devantler:

The failing **VCluster** system-test legs are **not caused by this PR** — this is a VCluster-specific cluster-bringup problem, not a code regression.

Decisive evidence from this PR's own CI run (`27146244827`), system-test legs by distribution:

| Distro | Result |
|---|---|
| K3s | 5 ✅ |
| KWOK | 5 ✅ |
| Vanilla | 5 ✅ |
| **Talos** (the distro this PR changes) | running, **0 failures** |
| **VCluster** (untouched by this PR) | **4 ❌** |

- This PR changes **only** `pkg/svc/provisioner/cluster/talos/detect_inplace.go`. The Talos provisioner is never instantiated for a VCluster cluster, so there is no path from this diff to VCluster bring-up — and indeed the **Talos legs are green/clean while only VCluster fails.**
- Failure signature is node/CNI bring-up, not application logic: `CIDRAssignmentFailed`, `cni plugin not initialized`, flannel `install-cni-plugin` stuck, `no nodes available to schedule pods`. One leg (Flux) even shows Flux reconciling successfully (`Ready: True`) before the node went unhealthy.
- It's not generic runner-load flakiness either — that would also hit K3s/Vanilla/KWOK, which all passed.

No code change — fixing a VCluster CNI bring-up failure from a Talos-only PR would be wrong. These required legs just need a **re-run** (which I can't trigger; `gh run rerun` is guardrail-blocked for my auth). If VCluster bring-up is failing on other PRs too, that's a separate CI-health item to investigate, independent of this PR.

PR #5242, opened 2026-06-12
feat: phase 1 codegen rails — make generate, enum/distribution single-sourcing, schema enrichment, meta resourceKinds, AI confirm-flag annotation

devantler:

Phase 1 (of 6) of the simplification & refactoring plan — the rails later phases ride on. **Stacked on #5240** (retarget to main when it merges). Behavior-preserving except the enumerated AI tool-surface deltas below.

## What's here (6 workstreams)

**1.1 `make generate` + full drift gating.** One command runs all six generators (schemas, CRD/deepcopy, docs, chat docs, mockery, web-ui types); chat docs join the `go:generate` graph; the CI generate job gains the chat-docs step and a drift check on the committed `web/ui/src/generated/ksail-config.ts`. New **MCP tool-surface golden snapshot test** (`pkg/toolgen/toolsurface_snapshot_test.go`) — the mechanical gate behind every future "tool surface unchanged" claim.

**1.5 Enum & distribution single-sourcing (the load-bearing item).** `setEnum[T]`/`validValueStrings[T]` generics replace 17 hand-copied 5-method pflag.Value implementations (`Set()` error text verified byte-identical); each enum's value list is declared exactly once; a registry-driven conformance suite (`TestEnumConformance`) replaces ~1,300 lines of copy-paste enum tests with strictly stronger assertions (exact ValidValues equality, full-value case-insensitivity round-trips). `distributionMeta` table replaces ~10 per-distribution switches — **adding a distribution is now one table row + one const**. `coverage_test.go` dissolved with every deleted test mapped to a named subsumer.

**1.2 Schema enrichment — editor schema only.** Field descriptions 51→152 (Go doc comments via `AddGoComments`), `deprecated: true` on alias fields, `metadata.name` pattern/maxLength, distribution×provider `allOf/if-then` guidance. **The CRD gains descriptions ONLY** — verified zero new validation keywords in the CRD diff, so existing CRs cannot be rejected server-side. `gen:types` now strips conditionals json2ts can't represent (it collapses such nodes to index signatures, which broke the SPA build — caught and fixed in this PR).

**1.3 Docs generator owns the inventories.** MCP tool catalog partial (true 6-tool surface incl. `tenant_write` and the previously missing subcommands), `cli-flags/index.mdx`, ProviderSpec/Options/AutoscalerConfig reference tables — all generator-owned with byte-equality in-sync tests. Found & fixed a real generator bug: the MCP partial was rendered from the same cobra root the cli-flags walk had already mutated (help-command injection), so generated output ≠ fresh render; the partial now renders from a fresh root.

**1.4 `/api/v1/meta` resourceKinds.** Additive `[{kind, namespaced, scalable, restartable, reconcilable, deletable, browsable}]` payload derived from the Go allowlist (exact-JSON wire test); the SPA builds its kind lists from meta with the TS constants kept permanently as the older-backend fallback. Adding a browsable kind is now a Go-only change.

**1.6 AI tool-surface safety.** Force injection is now driven by an explicit `ai.toolgen.confirm-flag` annotation instead of flag-name matching:
- **chat/MCP `workload_write` delete no longer receives kubectl's destructive `--force`** (PDB-bypass semantics) — regression-tested
- `init` no longer receives `force=overwrite` injection
- cluster update/delete, tenant delete, cipher rotate keep today's prompt-skip auto-confirm (annotated)
- executor resolves its own binary via `os.Executable()`; home-dir flag defaults normalized to `~/...` (no more machine-specific paths in tool schemas); non-applicable flags warn instead of silently dropping

## Tool-surface deltas (deliberate, snapshot-reviewed)
Schema changes from 1.6 only: home-dir DefValue normalization, neutralized descriptions for colliding merged flags, and the confirm-flag injection behavior changes above. Everything else verified byte-identical via the new snapshot.

## Validation
`go build ./...` ✅ · `go test ./...` ✅ (only the documented chat external-CLI flake under parallel load; passes isolated) · `golangci-lint --new-from-rev=origin/main` ✅ 0 issues · `make generate` run end-to-end ✅ (validates the target itself) · docs `npm run build` ✅ · web/ui `npm run build` ✅ (tsc) · CRD diff eyeballed: descriptions only ✅ · `go -C desktop mod tidy` clean ✅

## SemVer note
Typed `feat:` (minor). Exported-symbol changes are Go-SDK-internal (enum helper internals, `MetricsServer.EffectiveValue` signature gaining a Provider arg — single in-repo caller). Veto here if you want different typing.


devantler:
Addressed the performance-regression alert in b742302c.

Root cause: \`distributionMetas()\` rebuilt the full per-distribution table (6 rows of structs + slices) on **every** call, and \`distributionMetaFor\` is hit per-field by \`ComputeDiff\`'s \`EffectiveValue\` calls and by (un)marshalling's context/name lookups — turning a handful of allocs into 100+ per diff. Fixed by building the immutable table **once** at package init (rows still returned by value; no caller mutates them).

Verified locally — allocs back to the pre-PR baseline:

| Benchmark | Alert | After fix |
|---|---|---|
| \`ComputeDiff_NoChanges\` | 109 allocs / 9136 B | **1 alloc / 208 B** |
| \`ComputeDiff_*\` (all) | 22–151 allocs | 1–7 allocs |
| \`Cluster_MarshalJSON/Full\` | 643 allocs | 598 |
| \`PruneClusterDefaults/MostlyDefaults\` | 339 allocs | 285 |

The residual YAML/JSON-encode ns/op deltas are the known noisy-runner variance (unchanged allocs).

PR #5248, opened 2026-06-13
refactor: phase 2 shared infrastructure — context/name resolution, retry unification, pkg/di deletion, docker client narrowing

devantler:

Phase 2 (of 6) of the simplification & refactoring plan — the shared-infrastructure consolidation later phases build on. **Stacked on #5242** (retarget down the stack as parents merge). Behavior-preserving; **net −5.2k lines** across 171 files.

## What's here (7 workstreams)

**2.4 Delete `pkg/di`.** The samber/do container delivered exactly one production dependency — a two-timestamp timer. Commands now construct `timer.New()` directly or receive it via `lifecycle.Deps.Timer`; `*di.Runtime` unthreaded from ~52 files; `samber/do` dropped from go.mod; the no-op `Timer.Stop()` and its nil-guards removed. This unblocks the Phase 3 command decomposition (same `New*Cmd` signatures).

**2.5 Docker client narrowing.** New `pkg/client/docker.Client` (~25 actually-used methods, satisfied by `*client.Client`) replaces the 122-method `client.APIClient` across ~54 signatures; `GetConcreteDockerClient` retained for the two SDK injection points. Generated mock **8,262 → 1,759 lines** (the `MockAPIClient` type name is preserved via a `structname` override so all 239 test references keep compiling).

**2.2 Retry/poll/wait unification.** `netretry.Do(ctx, attempts, base, max, op, opts...)` — returns the last error unwrapped so callers keep sentinel identity via `errors.Is`; options for predicate override (chat), cancel-error mapping, and retry hooks. Replaces ~10 copy-pasted retry loops. Six provider poll loops moved to `wait.PollUntilContextTimeout` with sentinel remapping (and a `mapWaitError` helper that correctly propagates a condition error wrapping `context.Canceled` instead of misclassifying it as the poller's own timeout). New `internal/retry` for vcluster/kwok `createWithRetry`; `reconciler.PollUntilReady` for flux/argocd. **Side-effect bug fix:** Hetzner firewall/infrastructure deletes were uncancellable `time.Sleep` loops — now context-aware, so Ctrl-C during `cluster delete` is honored.

**2.1 Context/name resolution, one owner.** `pkg/svc/detector/cluster` owns inverse context mapping (`StripContextPrefix`/`MatchContexts`, the `k3k-` alias handled once, nested `vcluster-` convention modeled explicitly); `lifecycle.ResolveClusterName` is the single resolver; `k8s.ResolveKubeconfigPath` honors `KUBECONFIG`. 6+ divergent call sites converge (builds on Phase 1's `distributionMeta`/`ContextName`).

**3.1a (pulled forward) switch.go move-only split.** `switch.go` 951 → 561 lines + new `testing.go` (the 13 `Set*ForTests` seams the `delete.go` comment already pointed at), `output.go` (JSON/format helpers), `ttl.go` (`--ttl` helpers). Pure cut-paste, zero logic change — landed early per the plan's coordination rule so Phase 3's extraction doesn't rewrite the same file twice.

**2.6 setup test-seam normalization.** The four `post_cni.go` global override hooks moved into `InstallerFactories` fields with real defaults (~194 lines of mutex/getter/`Set*ForTests` quadruplets deleted).

**2.3 fsutil shared helpers.** `SplitYAMLDocuments`, one promoted `IsPathWithinDirectory` (fixes two divergent symlink-escape guards — a security drift), `UpdateKubeconfigFile` (options encode each caller's intentional prologue differences), `IsYAMLFile`.

## Behavior & release notes
- MCP/chat tool surface **byte-identical** — `pkg/toolgen` `TestToolSurfaceSnapshot` green (the pkg/di unthreading and docker narrowing touch no command wiring).
- Release-note tags: `cluster delete` now recognizes Talos-on-Docker / k3k contexts it previously skipped (deliberate); `k8s.ResolveKubeconfigPath` empty-path now resolves to `KUBECONFIG` when set.

## Validation
`go build ./...` ✅ · `go test ./...` ✅ (only the documented chat external-CLI parallel-load flake; passes isolated) · `golangci-lint --new-from-rev=origin/refactor/phase-1-codegen-rails` ✅ 0 issues · `mockery` regen ✅ · `go mod tidy` + `go -C desktop mod tidy` ✅ (desktop already tidy) · `TestToolSurfaceSnapshot` byte-identical ✅

## SemVer
Typed `refactor:`. Exported-symbol removals (`pkg/di`, docker interface change, constructor signatures) are Go-SDK-internal; no CLI/config/wire surface change. Veto here if you want different typing.


devantler:
## Phase 2 regression review — verdict: safe

Ran an extra-high-effort regression review of this PR rebased onto `main` (9 finder angles → 1-vote verify → sweep; 39 candidates → 27 deduped → 25 survived + 3 sweep). Focused on "does this break anything."

**No correctness regressions found.** One cleanup applied; the rest are either intended single-sourcing behavior changes (more-correct, worth a release note) or unreachable/platform-bound divergences.

### Applied
- **Removed 3 unused exported test seams** (`SetNodeSchedulabilityWaitForTests`, `SetCSRApproverWaitForTests`, `SetCloudProviderInitInstallForTests`) left by the `post_cni` globals→`InstallerFactories` migration — `post_cni_internal_test.go` sets the factory fields directly, so they had zero callers. `SetClusterStabilityCheckForTests` stays (still used).

### Intended behavior changes (release-note candidates, verified benign)
1. **`ksail cluster switch ""`** (explicit empty-string arg) now returns `ErrContextNotFound` instead of silently switching to the sole context / `ErrAmbiguousCluster`. Consequence of single-sourcing into `clusterdetector.MatchContexts`, which guards `clusterName == ""`. New behavior is more correct; no test/script asserts the old one. Only reachable via a literal empty arg (`pickCluster` never yields empty).
2. **`k8s.ResolveKubeconfigPath("")` now honors `KUBECONFIG`** for an empty path. Insulated on every CLI create/update/diff path by the `Connection.Kubeconfig` field-selector default (`~/.kube/config`, always non-empty), so no CLI regression. Only reachable via a hand-written empty-`kubeconfig` nested-K8s Cluster CR **and** a `KUBECONFIG` injected into the operator process (the chart injects none) — not a live path.
3. **`cluster delete` now probes `admin@`/`k3k-` context prefixes** (Talos / nested-K3s) via `ContextPrefixes()`. Wider detection, but the Talos network-name path resolves to `clusterName` — identical to the pre-existing unknown-distribution fallback — so no wrong cleanup output.
4. **`mapWaitError`** maps a parent-context deadline to the not-ready sentinel rather than a cause-wrapped message. Latent only: no production caller passes a deadlined ctx to the waiters, and internal-poll-timeout→sentinel is the intended contract (codified by `TestMapWaitErrorTimeoutReturnsSentinel`).
5. **GitOps/tenant YAML detection** switched to `fsutil.SplitYAMLDocuments` (YAML-aware). Strictly more correct — a `---` inside a block scalar no longer false-splits — and no realistic manifest framing was found that flips a detection result.

Build, full test suite, diff-lint, and full-run whole-block lint (dupl/funlen/cyclop/gocognit/wrapcheck/unparam) all clean on the rebased branch.

PR #5249, opened 2026-06-13
refactor: phase 3 god-file & service decomposition

devantler:
> 🤖 Generated by the Daily AI Assistant

Phase 3 (of 6) — the decomposition phase. Breaks up every god file the plan identified and pushes reusable engines out of cobra command files into `pkg/svc`. **Stacked on #5248.** Behavior-preserving; MCP/chat tool surface **byte-identical** (`TestToolSurfaceSnapshot`). 14 of 16 workstreams; 2 deferred (below).

## Move-only splits (landed first, byte-identical bodies)
- **cluster cmd reshuffle** — `reconciler.go`/`info.go`/`factories.go`/`dockerutil.go`/`oidc.go` split out of misnamed files (`componentReconciler` lived in list.go, info in diagnose.go, DI globals in delete.go).
- **cluster_test.go (7,486 lines)** → per-command test files + `testhelpers_test.go` + bench file (split via AST extraction, every body byte-identical).
- **talos/update.go (1,503)** → client/inventory/introspect/autoscaler_secret.
- **operator cluster_controller.go (913)** → namespace_gc/drift/status.

## Extractions & decomposition
- **update.go (1,970)** → `render.go` + an orchestrator type with struct-held run state (eliminates the 14-positional-arg signatures + their nolint escapes).
- **backup/restore engine** → new `pkg/svc/backup` (thin cobra wrappers remain).
- **workload** — Flux substitution engine + `pkg/svc/workloadwatch` extracted; 12 kubectl wrappers table-driven with cobra command groups; host-debug → `pkg/svc/hostdebug`; validate config double-load fixed.
- **chat.go (1,694)** → client/config/nontui/tui/tools; inverted layering fixed (chatui imports chatsvc); one SDK-event dispatcher feeding both frontends; generic `pickerState[T]`.
- **webui backend** — one shared `ResourceClient` resolver behind the two marker interfaces; `ClusterUpdater` capability; **one SSE snapshot broker per process** (was per-connection — N tabs no longer = N cloud-API loops); single `restConfigForCluster` seam; shared `BuildDistributionConfig`.
- **webui frontend** — `ResourcesView` 776 → ~300 + `ResourceDetailPanel`/`PodSession`; App.tsx 13 useState → one config; `lib/views.tsx` registry; `lib/health.ts`.
- **small services** — multi-doc-safe `rbaccm.go`; release-based GitOps detection with namespace fallback; image exporter split; ctx threaded through `DetectInfo`.
[...]

PR #5250, opened 2026-06-13
refactor: phase 4 per-surface UX — additive CLI targeting, cluster list --output json, VSCE rehab, config single-sourcing

devantler:

Phase 4 (of 6) — per-surface UX simplification. **All non-breaking** (additive flags / opt-in behaviors). **Stacked on #5249.** Tool surface changes only by the flags intentionally added below (snapshot regenerated, deltas reviewed: `cluster list` joins `--output`, new `--include-version-drift`, `-p`/`--name` additions).

## Config model (4.1)
- **Defaults single-sourcing** — FieldSelector defaults + OIDC/source-dir consts now reference `defaults.go`, guarded by a `default:` tag == const equality test.
- **`GitOpsEngine.IsNone()`/`Normalize()`** collapse the `"" || None` equivalence checks across 6 call sites.
- **Toggle bool-alias widening (non-breaking half)** — the loader accepts `certManager: true` as an alias for `Enabled` on the 9 toggle fields via a mapstructure decode hook; old string YAML still works; **field types unchanged** (the bool→enum type change is Phase 5).
- **⚠️ marshal-pipeline move DEFERRED.** Removing `Cluster.MarshalJSON` would change the operator/CR **wire payload** (controller-runtime + cr_service encode through it — that's why the `fullCluster` escape hatch exists). It needs operator chart e2e + a stored-CR upgrade test I can't run here. Left `MarshalJSON` in place and shipped `marshal_pipeline_move_test.go` as the regression gate for a dedicated follow-up PR.

## CLI consistency (4.2)
- One `RegisterClusterTargetFlags` + a shared resolution-priority help block; **additive `--name`** on connect/backup/restore; **additive `-p` shorthand** on create/update/init.
- **`cluster list --output json`** (`{name, provider, distribution, ttl}`) — additive (list had no `--output` before).
- **`cluster diff --include-version-drift`** — opt-in (default-off keeps `--exit-code` CI pipelines stable); renders the same drift set as `update --dry-run`.
- FieldSelector gains explicit `FlagName`/`Shorthand` + an init-time assertion (kills the silent `"unknown"` flag failure mode).
- Talos generation: declarative `patchSpec` table + `writePatchFile` + `[]GeneratedFile{Path, Action}` — fixes the false "created" notifications.

## VSCE rehabilitation (4.3)
- Consumes `cluster list --output json` — **Cloud Explorer was permanently empty** (parsed a long-gone format). Deleted the dead `schemaClient` (3-levels-dead; caused ~10s wizard-step hangs). One enum/description module for all 6 distributions + 5 providers. `resolveContext` helper + cache-invalidating `refreshAllViews` (fixes the 30s stale-status window). **First-ever test suite: 26 tests** (was 0).

## Cross-surface (4.4)
- `componentsInstall` capability so the web UI stops **silently dropping component selections** on local create.
- `clusterdiscovery` reports run-state; the local backend stops hardcoding `Ready`, so the web UI no longer shows green for a **Docker-stopped** cluster; additive capability-gated start/stop endpoints. (The CLI STATUS column + new `Stopped` phase *value* are Phase 5.)
- **Desktop module now passes `go test` + `golangci-lint`** under the shared config — it's finally CI-gateable. The `/desktop` build artifact is gitignored.
- **4.4b deferred** (local create-flow reuse) — depends on the deferred 3.5a installer-tree unification.

## Validation
`go build ./...` ✅ · `go test ./...` ✅ (only the documented chat external-CLI parallel-load flake; passes isolated) · `golangci-lint --new-from-rev` ✅ **0 issues** · `make generate` clean · tool-surface snapshot regenerated — deltas are exactly the additive flags ✅ · web/ui build ✅ · vsce compile + 26 tests ✅ · `go -C desktop test ./...` + desktop lint ✅

## SemVer
`refactor:` (additive only). New flags extend the MCP/chat tool surface (additive parameters) — called out so tool consumers see them in release notes.


devantler:
## Phase 4 regression review — verdict: 2 high-severity bugs fixed

Ran an extra-high-effort regression review of the rebased phase-4 (9 finder angles → 1-vote verify → sweep; 41 candidates → 38 survived). All findings are in **new additive surface** (the `--name` targeting flag + the web-UI Start/Stop lifecycle), so existing behavior was never at risk — but two were real high-severity defects. All confirmed bugs fixed in `ddc89872`.

### Fixed — bug-high
1. **web-UI Start/Stop cancelled mid-operation.** The new `Start`/`Stop` ran the provisioner action with the *request-scoped* ctx, so it was cancelled the instant the `202 Accepted` returned — a half-started/stopped Docker cluster. Threaded the background `context.WithoutCancel` ctx through the action func, matching Create/Delete.
2. **`backup`/`restore --name` could target the wrong cluster (destructive).** `--name` only changed the kubeconfig *path*, never the *context*, so on a shared `~/.kube/config` it silently operated on current-context — e.g. restoring a prod backup into the dev cluster. Now resolves `--name` to the actual kubeconfig context (the resolver `cluster switch` uses) and threads it to `kubectl --context` (new `kubectl.Client.WithKubeContext`); errors on no/ambiguous match instead of silently mis-targeting.

### Fixed — bug-medium / bug-low
- **`connect --name`** derived the k9s context from the *config's* distribution → wrong prefix for a cross-distribution target; now resolves the real context. (`connect --help` also advertised a `--provider` flag it doesn't register — gave it a no-provider help variant.)
- **Talos OIDC patch re-generation** read the CA file *before* the exists/!force skip, breaking idempotent re-runs when the CA moved; `writePatchFile` now checks existence before invoking the side-effecting content func.
- **`nodeAutoscaling: true`** was rejected while every other toggle accepts the bool alias → added `NodeAutoscaling` to `ToggleEnumTypes`.
- **`cluster list`** did a wasted per-cluster Docker run-state probe it never displays → gated behind `Discoverer.ProbeRunState` (web UI opts in; CLI skips).

### Deferred (documented follow-up, not a regression)
- The web UI renders a *degraded* Docker cluster (some nodes down, control-plane up) as the grey "Stopped" badge. This matches the documented 3-state `RunState` model (`Stopped` = "none **or not all** running"); a proper fix is a 4th `Degraded` state spanning Go + the web-UI TS rendering, out of scope for this safety pass.

### Verification
build ✅ · full test ✅ (only the known chat env-flake, passes isolated) · `deadcode -test ./...` empty ✅ · `jscpd` 0 clones ✅ · `golangci-lint --new-from-merge-base` 0 + full-lint on changed pkgs clean (gosec/funcorder/nonamedreturns resolved) ✅ · vsce compile + 26 tests ✅

PR #5254, opened 2026-06-14
feat: phase 5 changes — webui api move, --force/--yes split, flag/output normalization, imageVerification promotion, bool→enum toggles, interactive cmds out of AI surface, Stopped status

devantler:

Phase 5 (of 6) — the **fenced phase**. Each item ships with a backward-compatible deprecation path so **old `ksail.yaml` and old flags keep working + warn for one release**. **Stacked on #5250.** Batched into one major-release window per the plan's SemVer rule.

### CLI
- **`update --force` split.** `--force` no longer bypasses PodDisruptionBudgets — it's now a **hidden deprecated alias of `--yes`** (skip-prompts only) that still works for one release + warns. New **`--force-drain`** carries the PDB-bypassing drain + partition-wipe semantics. *CI scripts using `--force` to skip prompts keep that, but lose the unsafe drain — add `--force-drain` to restore it.* The `ai.toolgen.confirm-flag` annotation was re-pointed to `--yes`.
- **`cluster backup`/`restore` path is now positional** (`backup [<output>]`, `restore [<input>]`), matching `workload export/import`. `--output/-o` and `--input/-i` are **hidden deprecated aliases** for one release; omitting both now errors.
- **`backup -n`** deprecated on `--namespaces` (`-n` reserved for `--name` across the cluster group); still maps for one release with a warning.

### Config (`ksail.yaml`)
- **`spec.cluster.imageVerification` promoted** from `spec.cluster.talos.imageVerification` — now steers Kind/K3s/Talos. Old location is a **deprecated alias** via `migrateDeprecated*` (conflicting values error; equal values migrate silently).
- **`sops.enabled` bool → `SOPSEnabled` tri-state** `{Default, Enabled, Disabled}` (Default = old auto-detect/nil). **`autoscaler.node.enabled` bool → `NodeAutoscalerEnabled`**. Old YAML booleans are coerced by the Phase-4 load-time decode hook, so existing files keep working. `--node-autoscaler-enabled` becomes a value flag (legacy `true`/`false` accepted for one release; the bare no-value form no longer works).

### AI tool surface (MCP + chat)
- **Interactive commands leave the tool surface**: `cluster connect` (k9s), picker-mode `cluster switch`, `workload edit` ($EDITOR) — they can't be driven by a tool client. No-arg `switch` now returns a clean non-TTY error instead of hanging. Driven via a new `ai.toolgen.interactive` annotation the consolidation collector now honors.
- Tool-param changes (`--force`→`--yes`/`--force-drain`, backup/restore path flags gone, `node-autoscaler-enabled` bool→string) — snapshot regenerated; deltas reviewed to match exactly these intended breaks.

### Status vocabulary
- **New `ClusterPhase` value `Stopped`** + a STATUS column in `cluster list` and a `status` field in `cluster list --output json` (additive to the JSON contract — existing consumers just gain a field). The stopped state is **also** emitted as `Ready=False/reason=Stopped`, so pre-enum consumers keep working. This is the durable contract that lets VSCE's `spawnDocker` status-sniffing retire in a follow-up.

### Go SDK
- **`pkg/operator/api` → `pkg/webui/api`** (import-path move; the HTTP layer belongs next to the embedded SPA). Wire/routes/JSON/SSE identical. `NewCRClusterService*` constructors → `pkg/operator`. Migration: one-line import rewrite. *(No forwarding shim shipped — flag if you want a one-release `pkg/operator/api` re-export stub for external SDK consumers.)*

## Deferred (NOT in this PR)
- **5.8 `*EnvVar`/OptionsAWS rationalization** — needs a keep-secrets-out-of-config decision (Hetzner/Omni have no plain secret-value fields). Bringing as a design proposal, not a refactor.

## Validation
`go build ./...` + `go build -C desktop ./...` ✅ · `go test ./...` ✅ (only the documented chat external-CLI parallel-load flake; passes isolated) · `golangci-lint --new-from-rev` ✅ **0 issues** · `make generate` clean (CRD enum gains `Stopped`; schema sops/autoscaler now string enums) · tool-surface snapshot regenerated — deltas are exactly the intended breaks ✅ · deprecation round-trip tests (old YAML + old flags still load) ✅ · web/ui + docs builds ✅ · binary spot-check: `--force`/`--force-drain`/`--yes` + positional backup verified in `--help`


devantler:
> 🤖 Generated by an autonomous review pass

## Phase 5 — rebased onto `main`, reviewed twice, made backward-compatible (one accepted operator exception)

This branch was rebased onto current `main` and put through **two** extra-high-effort
multi-agent regression reviews (9 finder angles + a dedicated breaking-change detector
→ per-finding verification → sweep; the second run on the de-breaked diff to verify the
fixes hold). It now ships as a **minor (`feat:`)**, not a major — the CLI, `~/.ksail`
state files, ksail.yaml loading, REST/web-UI and AI surfaces are all backward-compatible.
There is **one intentional, documented exception** (operator CRD), called out below.

### Backward-compat fixes applied (each regression-tested)

1. **`spec.json` + REST `bool→enum` load failure** — `sops.enabled` / `autoscaler.node.enabled`
   only had the YAML decode-hook alias; the `json.Unmarshal` read paths (persisted state +
   operator/REST ingest) had none. Fixed with a shared `unmarshalToggleEnumJSON` helper +
   `UnmarshalJSON` on both enums (true→Enabled / false→Disabled / null no-op / string verbatim).
2. **`update --force` lost its destructive semantics** — the deprecated `--force` now retains
   its full pre-split behavior (skip prompts **and** the PDB-bypassing drain + partition wipe)
   for one release; `--yes` (consent-only, AI-injectable) and `--force-drain` (destructive)
   are the new split.
3. **`--node-autoscaler-enabled` bare flag** — restored via `NoOptDefVal` so the valueless
   form still means `Enabled`; `=true`/`=Enabled` still work.
4. **Toggle JSON validation (robustness/security)** — a third focused security pass found the new
   `UnmarshalJSON` accepted *any* string verbatim, so a typo'd `sops.enabled: "Enabld"` silently
   fell through to SOPS auto-detect (fail-open at a secret boundary). It now validates against the
   enum's value set like the CLI path (rejecting unknown values, normalising case) — which also
   restores the reject-invalid behavior the old `*bool`/`bool` fields had.

### Verified already-safe

webui pkg move (desktop updated in lockstep), backup/restore `--output`/`-i`/`-n` deprecated
aliases, imageVerification load-migration, `Stopped` status (additive + `Ready=False/reason=Stopped`),
interactive cmds (CLI intact, only off the AI surface).

### Known behavior changes (call-outs)

- **`update --yes` narrowed to consent-only.** On the prior release `--yes` was a full alias of
  `--force` (it also force-drained / authorized wipes). It is now skip-prompts-only so the chat
  assistant can safely auto-inject it. `--force` preserves the old combined behavior for one
  release; a script using `--yes` purely for a PDB-bypass should move to `--force-drain`.
- **Operator / GitOps CRD (the one accepted incompatibility).** The bool→enum change regenerates
  the operator CRD so `spec.cluster.sops.enabled` and `spec.cluster.autoscaler.node.enabled` are
  now `type: string` (were `type: boolean`). The kube-apiserver enforces this on writes, so a
  Cluster CR / GitOps manifest that sets these as a YAML **boolean** must switch to the string
  form (`true`→`"Enabled"`, `false`→`"Disabled"`; sops also supports `"Default"`). All other
  surfaces (CLI, `~/.ksail` spec.json, REST/web-UI, ksail.yaml loading) still accept the boolean.
  A structural CRD cannot accept both bool and string for one field, and the in-process bool
  coercion runs after apiserver admission — so this surface can't be shimmed. Released as a minor
  by maintainer decision (avoiding a v8 module-path major); operator users migrate the two fields.

### Gates

`go build ./...`, full `go test ./...` (only the known chat env-flake, passes isolated),
`golangci-lint` (full, changed pkgs), `jscpd` (0 clones), full-module `deadcode`, desktop
`go vet`, `make generate` (clean; mock version-skew churn reverted), plus legacy-bool and
bare-flag reproductions — all green.

PR #5256, opened 2026-06-15
feat(autoscaler): support a priority list of expanders

devantler:
## Summary

KSail's cluster-autoscaler `expander` was a single enum (`Price | LeastWaste | LeastNodes | Random`), so there was no way to configure upstream cluster-autoscaler's **priority-list** form (`--expander=least-nodes,least-waste`), where expanders are applied as a chain — the first filters node groups and each later one breaks the previous tie. Because KSail owns the autoscaler install, the only workaround was to bypass KSail entirely.

This PR lets `expander` accept **either a single value or an ordered priority list**, fully backward-compatible with existing scalar configs.

```yaml
spec:
  cluster:
    autoscaler:
      node:
        enabled: true
        expander: [LeastNodes, LeastWaste]   # or just: LeastWaste
```

## What changed

- **API** — new `AutoscalerExpanderList` type; `NodeAutoscalerConfig.Expander` changed from the scalar enum to the list.
- **Backward-compatible loading** — a mapstructure decode hook normalises a scalar into a one-element list, so `expander: LeastWaste` still works. A comma scalar (`expander: "LeastNodes,LeastWaste"`) and a YAML sequence both decode too.
- **Validation** — each entry is validated, duplicates are rejected (new `ErrDuplicateAutoscalerExpander`), and the Hetzner "Price unsupported" check scans the whole list.
- **Install/detect round-trip** — installer joins the list into the chart's `extraArgs.expander` (→ `--expander=least-nodes,least-waste`); detector splits it back. Empty falls back to `least-waste`.
- **Diff** — compares lists via `String()`, still classified in-place.
- **Schema / CRD / docs** — `expander` schema is `oneOf(string-enum, array-of-string-enum)` so both shapes validate in editors; CRD, DeepCopy, and reference docs regenerated.

## Testing

- `go build ./...` — clean
- Affected packages pass, with added coverage for list `String()`, multi/invalid/duplicate validation, Price-in-list rejection on Hetzner, the comma round-trip, and scalar/comma/sequence config loading.
- `golangci-lint run` — no new source issues.


devantler:
Both red signals here are non-code and have been addressed:

- **System Test (Talos, Docker, true)** failed at the `docker/login-action` step with `Get "https://registry-1.docker.io/v2/": context deadline exceeded` — a transient Docker Hub login timeout, before any KSail code ran. Every other Talos variant (incl. the config-heavy `--cni Cilium`, `--gitops-engine Flux/ArgoCD`, `--name ...` jobs) passed. I've re-run the failed job.
- **Performance Alert** is CI-runner noise, not a regression: the slowdown is a uniform ~2–2.7× across *all* benchmarks, including `BenchmarkYAMLEncode/Minimal` and `BenchmarkJSONEncode`, which encode a cluster with **no** autoscaler config and are therefore completely unaffected by this change (a nil `[]AutoscalerExpander` marshals identically to the previous empty scalar). A real regression from this diff would only touch autoscaler marshaling, not Minimal/Prune/JSONEncode.

Issue #5294, opened 2026-06-16
cluster update runs the Talos OS upgrade before reconciling machine config


devantler:
## Summary

`ksail cluster update` performs the **distribution (Talos OS) upgrade before** it reconciles/applies the machine config. Because a Talos OS upgrade validates the machine config **already committed on the node**, a config change that is a *prerequisite* for the target version can never be applied through `cluster update` — the upgrade runs first, validates the stale on-node config, and aborts before the config-reconcile step is ever reached.

## Evidence

- `pkg/cli/cmd/cluster/orchestrator.go` `run()` calls `reconcileClusterVersions(provisioner)` (distribution upgrade) **first**, and only *afterwards* computes the diff and calls `applyOrReportChanges` (the machine-config reconcile).
- The distribution upgrade path does **not** stage/apply the desired config first: `UpgradeDistribution` → `rollingUpgradeNodes` → `upgradeNodeTalosVersion` → `lifecycleUpgradeNode` does `pull → LifecycleService.Upgrade → reboot` with no `ApplyConfiguration`. (`stageNodeConfigForReboot` exists but is only used by the *config-driven* rolling path in `rolling.go`, not by the version upgrade.)

## Concrete repro (how this surfaced)

A Talos **v1.13.4** installer added validation rejecting `install.extraKernelArgs` together with `install.grubUseUKICmdline=true`. A cluster created on ≤v1.13.3 has both committed (extraKernelArgs from a user patch + grubUseUKICmdline:true from the ≥1.12 contract default). On `cluster update` to v1.13.4:

```
Error: machine configuration is invalid: 1 error occurred:
  * install.extraKernelArgs and install.grubUseUKICmdline can't be used together
✗ distribution upgrade to pinned version v1.13.4 failed: ... talos upgrade failed
```

Even after the user fixes the config (e.g. `grubUseUKICmdline:false`), a **single** `cluster update` that also bumps the version still fails, because the upgrade validates the not-yet-reconciled on-node config. The only workaround is a two-step deploy: reconcile config at the old version first, *then* bump.

## Proposed fix

Apply/stage the reconciled machine config to each node **before** triggering its OS upgrade (the upgrade reboots anyway), so the installer validates the *desired* config rather than the stale committed one — letting a config prerequisite for the upgrade self-heal in one `cluster update`. Per-node staging machinery already exists (`buildDesiredNodeConfig` / `stageNodeConfigForReboot`).

## Impact

[...]

PR #5296, opened 2026-06-16
feat(talos): bake extra kernel args into the Image Factory schematic

### LINKED ISSUE #5295: Support declaring schematic extra kernel args (spec.cluster.talos.extraKernelArgs)
## Problem

The only way to add kernel arguments to a KSail-managed Talos node today is the Talos field `machine.install.extraKernelArgs` (via a user patch). That field is **deprecated** upstream:

```go
// config/types/v1alpha1/v1alpha1_types.go
// Deprecated: Use Image Factory/imager instead to build a proper installer.
InstallExtraKernelArgs []string `yaml:"extraKernelArgs,omitempty"`
```

…and on Talos **≥1.13.4** it actively conflicts with the `≥1.12` version-contract default `install.grubUseUKICmdline: true` (`VersionContract.GrubUseUKICmdlineDefault()` returns true for `>1.11`), so a config carrying both is rejected mid-upgrade:

```
install.extraKernelArgs and install.grubUseUKICmdline can't be used together
```

KSail generates configs with the `TalosVersion1_12` contract, so **every** KSail Talos config defaults `grubUseUKICmdline:true` — meaning any user who adds `machine.install.extraKernelArgs` hits this on Talos ≥1.13.4.

## Best-practice direction

Talos says to bake kernel args into the **installer image via the Image Factory schematic** (`customization.extraKernelArgs`) instead of the deprecated machine-config field. When the arg is in the schematic/UKI, `grubUseUKICmdline:true` is fine and there is no conflict.

KSail already computes one Image Factory schematic from `spec.cluster.talos.extensions` (`pkg/fsutil/configmanager/talos/schematic.go` `NewSchematic`) and flows that single schematic to **all** node paths — `machine.install.image` (static, `applySchematic`), the Hetzner autoscaler snapshot, and the rolling-upgrade installer (`resolveSchematicID`/`resolveInstallerImage`). But `NewSchematic` only populates `SystemExtensions.OfficialExtensions`; the upstream `schematic.Customization` type also has `ExtraKernelArgs []string` (image-factory `pkg/schematic`), which KSail never sets.

## Proposal

Add `spec.cluster.talos.extraKernelArgs []string`, threaded into `NewSchematic`'s `Customization.ExtraKernelArgs` (alongside `extensions`). Because the schematic ID then incorporates the kernel args and is reused everywhere, the args are baked consistently into static install images, autoscaler snapshots, and upgrade installers — the non-deprecated, conflict-free path. An explicit `schematicId` still takes precedence (kernel args ignored), mirroring `extensions`.

### Acceptance criteria
- New `spec.cluster.talos.extraKernelArgs` field (schema + declarative-config docs regenerated, not hand-edited).
- `NewSchematic` populates `Customization.ExtraKernelArgs`; the computed schematic ID changes deterministically with the args.
- Args flow to `machine.install.image`, the autoscaler snapshot, and the rolling-upgrade installer via the existing single-schematic resolution.
- Skipped when an explicit `schematicId` is set (consistent with `extensions`).
- Unit tests covering the schematic ID change and the skip-on-`schematicId` behavior.

Real-world consumer: the devantler-tech/platform prod cluster currently sets `machine.install.extraKernelArgs: [security=apparmor]`; this lets it move to the schematic instead.

devantler:
> Fixes #5295

## Problem

The only way to add kernel arguments to a KSail-managed Talos node was the Talos field `machine.install.extraKernelArgs`, which is **deprecated upstream** (`// Deprecated: Use Image Factory/imager instead to build a proper installer.`) and, on Talos **≥1.13.4**, conflicts with the `≥1.12` version-contract default `install.grubUseUKICmdline: true` — a config carrying both is rejected mid-upgrade:

```
install.extraKernelArgs and install.grubUseUKICmdline can't be used together
```

Since KSail generates configs with the `TalosVersion1_12` contract, this trips **any** KSail user who adds `machine.install.extraKernelArgs`. (Surfaced by a real prod cluster — see #5294 for the related upgrade-ordering bug it exposed.)

## Change

Add **`spec.cluster.talos.extraKernelArgs`**, folded into the same Image Factory schematic KSail already computes from `spec.cluster.talos.extensions` (the schematic's `customization.extraKernelArgs`). Because that *one* schematic ID drives every node path, the args are baked consistently into:

- `machine.install.image` for static nodes (`applySchematic`),
- the Hetzner autoscaler snapshot, and
- the rolling-upgrade installer (`resolveInstallerImage`).

This is the **non-deprecated, conflict-free** path: with the arg in the schematic/UKI, `grubUseUKICmdline: true` is fine. An explicit `schematicId` still takes precedence (kernel args ignored), exactly mirroring `extensions`.

### Behavior notes
- Kernel args are **order-preserving** (trim + drop-empties only, no sort/dedup — unlike extensions) since arg order can be semantically meaningful; the declared order is what's hashed into the schematic ID.
- An empty/absent list omits `customization.extraKernelArgs`, leaving the schematic ID **identical** to the extensions-only result — no change for existing users.

## Implementation
`NewSchematic(extensions, extraKernelArgs)` → threaded through `applySchematic` / `newConfigsWithEndpointAndSecrets` / the `Configs` regenerate-snapshot path (so args survive secret realignment) → `ConfigManager.WithKernelArgs` → wired from `spec.cluster.talos.extraKernelArgs` in the ksail config manager. Generated artifacts regenerated via `make generate` (CRD, deepcopy, JSON schema, web-UI types).

## Tests
- Schematic-level: kernel args change the ID; deterministic; order-significant; whitespace/empty normalization; kernel-args-only produces a valid ID.
- Config-level: SchematicID set + install image patched with kernel-args-only; args differ from extensions-only; **preserved across `WithName`/`WithEndpoint`** (guards the regenerate threading).
- Env-var drift test updated (kernel args are consumed verbatim into the schematic, like `extensions`).

`go build ./...`, `golangci-lint run`, and `go test ./...` pass (the unrelated `pkg/cli/cmd/chat` Copilot-CLI subprocess tests fail only in this local sandbox with `signal: killed`).

## Consumer
devantler-tech/platform will adopt `spec.cluster.talos.extraKernelArgs: [security=apparmor]` and drop its deprecated `machine.install.extraKernelArgs` patch once this ships in a release.

### DIFF
--- a/charts/ksail-operator/crds/ksail.io_clusters.yaml
+++ b/charts/ksail-operator/crds/ksail.io_clusters.yaml
@@ -463,6 +463,20 @@ spec:
                         items:
                           type: string
                         type: array
+                      extraKernelArgs:
+                        description: |-
+                          ExtraKernelArgs lists extra kernel command-line arguments to bake into the node image
+                          via the Talos Image Factory schematic (the schematic's customization.extraKernelArgs).
+                          KSail folds these into the same schematic it computes from Extensions, so they are applied
+                          consistently to the static install image, the Hetzner autoscaler snapshot, and the
+                          rolling-upgrade installer. This is the non-deprecated replacement for the Talos
+                          machine.install.extraKernelArgs config field (deprecated upstream in favour of Image
+                          Factory/imager), and it avoids that field's conflict with install.grubUseUKICmdline on
+                          Talos >= 1.13.4. Example: ["security=apparmor"].
+                          When SchematicID is set, it takes precedence and these are ignored.
+                        items:
+                          type: string
+                        type: array
                       extraPortMappings:
                         description: |-
                           ExtraPortMappings defines additional port mappings from Docker containers to the host.
--- a/pkg/apis/cluster/v1alpha1/envvar_drift_test.go
+++ b/pkg/apis/cluster/v1alpha1/envvar_drift_test.go
@@ -138,6 +138,7 @@ func skippedClusterWorkloadConfigFields() []string {
 		"Cluster.SOPS.Extract.File",
 		"Cluster.SOPS.Extract.PublicKeys[]",
 		"Cluster.Talos.Extensions[]",
+		"Cluster.Talos.ExtraKernelArgs[]",
 		"Cluster.Talos.ExtraPortMappings[].Protocol",
 		"Workload.Flux.Verify.MatchOIDCIdentity[].Issuer",
 		"Workload.Flux.Verify.MatchOIDCIdentity[].Subject",
--- a/pkg/apis/cluster/v1alpha1/options.go
+++ b/pkg/apis/cluster/v1alpha1/options.go
@@ -70,6 +70,16 @@ type OptionsTalos struct {
 	// The Image Factory resolves extension versions automatically per Talos release.
 	// When SchematicID is also set, it takes precedence over Extensions.
 	Extensions []string `json:"extensions,omitzero"`
+	// ExtraKernelArgs lists extra kernel command-line arguments to bake into the node image
+	// via the Talos Image Factory schematic (the schematic's customization.extraKernelArgs).
+	// KSail folds these into the same schematic it computes from Extensions, so they are applied
+	// consistently to the static install image, the Hetzner autoscaler snapshot, and the
+	// rolling-upgrade installer. This is the non-deprecated replacement for the Talos
+	// machine.install.extraKernelArgs config field (deprecated upstream in favour of Image
+	// Factory/imager), and it avoids that field's conflict with install.grubUseUKICmdline on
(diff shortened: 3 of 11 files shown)


devantler:
Superseded by #5297.

Reworked per maintainer feedback: rather than adding a `spec.cluster.talos.extraKernelArgs` **config field** (this PR), kernel args stay declared in **native Talos patch files** (`machine.install.extraKernelArgs`). KSail now reads them from the patches, folds them into the Image Factory schematic, and clears the deprecated field — keeping the ksail config lean while still moving off the deprecated field and avoiding the Talos ≥1.13.4 `grubUseUKICmdline` conflict. Still `Fixes #5295`.

PR #5313, opened 2026-06-17
feat(talos): reuse lowest-available node index when provisioning (#5312)

### LINKED ISSUE #5312: feat: reuse lowest-available node index when provisioning, not highest+1
## Problem

When KSail provisions a replacement node during `ksail cluster update`, it names the new node by **appending to the highest existing index + 1**, rather than filling the **lowest available** index in the series.

So when a node is lost and removed (etcd member removed + Hetzner VM deleted), recreating it does **not** reclaim the freed name — it allocates a brand-new, higher one.

## Observed (real prod recovery, 2026-06-17)

A 3-node Talos/Hetzner control plane lost `prod-control-plane-2`. After cleanly removing it (etcd `remove-member` + `hcloud server delete`), the cluster had `prod-control-plane-1` and `prod-control-plane-3`. Running `ksail cluster update` to restore the 3rd control plane created:

```
✓ control-plane node prod-control-plane-4 created
```

i.e. **`prod-control-plane-4`** (highest `3` + 1), instead of reusing the freed **`prod-control-plane-2`**.

## Expected

Pick the **lowest available index** in the node series (fill gaps first). With `prod-control-plane-1` and `prod-control-plane-3` present, the next control plane should be `prod-control-plane-2`.

## Why it matters

- A recreated node **reclaims the lost node's name**, so recovery restores the original topology/names instead of drifting.
- Names stay a clean contiguous set; over repeated node failures/replacements the indices don't climb arbitrarily (`-4`, `-7`, `-12`, …).
- Runbooks, dashboards, alerts, and operator mental-models that reference node names by index stay valid after a recovery.
- Avoids a manual post-recovery rename (which on a live Talos etcd node is disruptive — node certs are bound to the hostname, so a rename ≈ re-provision).

## Suggested behaviour

When allocating a new node name in a `<prefix>-<role>-<n>` series, choose `min(n)` not in the set of existing indices for that role, rather than `max(existing)+1`. Applies to control-plane and worker series alike.

## Repro

1. Start a 3 control-plane Talos cluster (`prod-control-plane-1..3`).
2. Remove the middle node: `talosctl etcd remove-member <id>` + delete its VM.
3. `ksail cluster update`.
4. Observe a `prod-control-plane-4` is created instead of `prod-control-plane-2`.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) on behalf of @devantler

devantler:
## What

Closes #5312.

When provisioning a replacement node during `ksail cluster update`, KSail named the new node **max(existing index) + 1** instead of filling the **lowest freed** index. After a lost node was cleanly removed (etcd `remove-member` + `hcloud server delete`), recreating it allocated a brand-new higher name (`prod-control-plane-4`) rather than reclaiming the freed `prod-control-plane-2` — so names drift ever higher (`-4`, `-7`, `-12`, …) over repeated failures.

## Change

Replace the shared `max+1` allocator (`nextNodeIndexFromNames`) with **`availableNodeIndices`**, which returns the next *N* suffixes preferring the **lowest free** indices: gaps within `1..max` are reclaimed (lowest first) before the series extends past its highest used index.

Applied uniformly via the two role-specific wrappers (`availableHetznerNodeIndices`, `availableDockerNodeIndices`) across every node-allocation path:

| Path | Function | Trigger |
|------|----------|---------|
| Hetzner scale-up | `addHetznerNodes` | `cluster update` raises node count (the reported scenario) |
| Hetzner recreate | `createReplacementServer` | rolling server-type replacement |
| Docker scale-up | `addDockerNodes` | `cluster update` raises node count |

This covers **control-plane and worker** series alike. For Docker, the reclaimed index also reclaims the freed node's **static IP** (IPs are derived from the index).

### Why a per-node index *list*

Callers now receive a list of indices (one per node) rather than a single base index plus a running `+offset`. This keeps multi-node scale-ups that span a gap collision-free — e.g. existing `{1,3}`, add 2 → `{2,4}` (not `{2,3}`, which would collide with the existing `3`).

## Behaviour example

3-node control plane loses `prod-control-plane-2`; after cleanup the cluster has `-1` and `-3`. `ksail cluster update`:

- **Before:** creates `prod-control-plane-4`
- **After:** creates `prod-control-plane-2` ✅

## Tests

- Rewrote `scale_shared_test.go` around `availableNodeIndices` with 13 cases: lowest-gap fill, the exact prod-recovery scenario, gap-then-extend multi-node, lowest-of-multiple-gaps, leading-gap reclaim, non-numeric/zero-count guards.
- `go test ./pkg/svc/provisioner/cluster/talos/...` → **519 passed**.
- Full `go test ./...` green except the known env-flaky `chat` Copilot-CLI tests (pass in isolation; unrelated to this package).
- `golangci-lint run --new-from-rev=HEAD` → clean.

## Note (out of scope)

The k3d provisioner (`pkg/svc/provisioner/cluster/k3d/update.go` `nextAgentIndex`) has the same `max+1` pattern for K3s agent nodes. Left untouched here since #5312 targets the Talos/Hetzner scenario (where name/cert/IP reclamation matters for recovery) and k3d's batch-offset logic warrants its own change + tests.

### DIFF
--- a/pkg/svc/provisioner/cluster/talos/export_test.go
+++ b/pkg/svc/provisioner/cluster/talos/export_test.go
@@ -34,9 +34,9 @@ func NewNodeWithRoleForTest(ip, role string) NodeWithRoleForTest {
 	return nodeWithRole{IP: ip, Role: role}
 }
 
-// NextNodeIndexFromNamesForTest exposes nextNodeIndexFromNames for unit testing.
-func NextNodeIndexFromNamesForTest(names []string, prefix string) int {
-	return nextNodeIndexFromNames(names, prefix)
+// AvailableNodeIndicesForTest exposes availableNodeIndices for unit testing.
+func AvailableNodeIndicesForTest(names []string, prefix string, count int) []int {
+	return availableNodeIndices(names, prefix, count)
 }
 
 // AddDockerNodesForTest exposes addDockerNodes for unit testing.
@@ -206,13 +206,19 @@ func CalculateNodeIPForTest(
 }
 
 // PreCalculateNodeSpecsForTest exposes preCalculateNodeSpecs for unit testing.
-// Returns node names and IPs as parallel slices (nodeSpec is unexported).
+// It builds a contiguous index range from nextIndex for backwards-compatible test
+// inputs. Returns node names and IPs as parallel slices (nodeSpec is unexported).
 func PreCalculateNodeSpecsForTest(
 	cidr netip.Prefix,
 	clusterName, role string,
 	nextIndex, count, cpCount int,
 ) ([]string, []netip.Addr, error) {
-	specs, err := preCalculateNodeSpecs(cidr, clusterName, role, nextIndex, count, cpCount)
+	indices := make([]int, count)
+	for i := range count {
+		indices[i] = nextIndex + i
+	}
+
+	specs, err := preCalculateNodeSpecs(cidr, clusterName, role, indices, cpCount)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/svc/provisioner/cluster/talos/rolling_recreate.go
+++ b/pkg/svc/provisioner/cluster/talos/rolling_recreate.go
@@ -352,9 +352,10 @@ func (p *Provisioner) cordonAndDrain(
 	return nil
 }
 
-// createReplacementServer provisions a single new server for the role using the
-// next available node index (computed after the outgoing server was deleted, so
-// names do not collide).
+// createReplacementServer provisions a single new server for the role. The index
+// is computed after the outgoing server was deleted, so its freed slot is the
+// lowest available index and the replacement reclaims the removed node's name
+// rather than allocating a higher one (#5312).
 func (p *Provisioner) createReplacementServer(
 	ctx context.Context,
 	hzProvider *hetzner.Provider,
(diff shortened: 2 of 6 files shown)


devantler:
### 🛡️ Vulnerability Scan failure — root cause: govulncheck v1.4.0 tool panic (not this PR)

The failing `🛡️ Vulnerability Scan` is **not a vulnerability finding** and is **not caused by this PR** (which adds no dependencies — Go-logic-only). govulncheck itself crashed:

```
panic: ForEachElement called on type containing *types.TypeParam
golang.org/x/vuln/scan.(*Cmd).Wait → internal/scan.RunGovulncheck
##[error]govulncheck failed to run (exit 2) - operational error, not a scan result.
```

This is a **v1.4.0 regression**: it panics on generics under Go 1.26. Reproduced locally — `govulncheck@v1.4.0 -format json ./...` panics (exit 2), while `govulncheck@v1.3.0` runs to completion on the same module under the same Go 1.26.4. The scanner is installed via `@latest` in `devantler/govulncheck-action/action.yml:56` (`go install golang.org/x/vuln/cmd/govulncheck@latest`), so `@latest` = the broken v1.4.0 currently breaks the check on **every** open PR in the repo.

**Fix belongs in the org tooling, not here:** pin `devantler/govulncheck-action` off `@latest` to a working version (e.g. `@v1.3.0`) until x/vuln ships a fix. There is nothing to change in this PR to make the check pass.

Separately, v1.3.0 reports 4 pre-existing reachable advisories (a known no-fix docker/k8s backlog) — also unrelated to this PR.

PR #5378, opened 2026-06-21
feat(workload): support Kubescape exceptions + spec.workload.scan config

### LINKED ISSUE #5369: Support Kubescape exceptions in `ksail workload scan` (native `--exceptions` / config)
## Problem

`ksail workload scan` (Kubescape) has no way to apply **exceptions**, so it can't be used as a hard CI gate (`--compliance-threshold 100`) for a GitOps repo.

GitOps platforms enforce many security controls at **runtime** — not in the static manifests:

- admission mutation (e.g. Kyverno injecting `securityContext`, resource limits, or a default-deny `NetworkPolicy` at pod creation)
- VPA-managed resources (Deployment specs intentionally omit static requests/limits so VPA can right-size)
- `CiliumNetworkPolicy` instead of `networking.k8s.io/NetworkPolicy` (Kubescape does not recognise it as equivalent)

A static scan cannot see runtime enforcement, so it reports these as failures. Kubescape's **exceptions** mechanism is the native way to suppress such justified findings — but `ksail workload scan` exposes no way to provide them:

- there is no `--exceptions` flag (the only flags are `--framework`, `--compliance-threshold`, `--format`, `--output`, `--verbose`, `--no-render`)
- it does **not** read in-cluster `ClusterSecurityException` CRs that happen to be in the scanned directory (verified: adding namespaces to a `ClusterSecurityException` does not change the compliance score)

Net effect: the scan always reports the same runtime-enforced false-positives, so `--compliance-threshold 100` can never pass and the scan cannot gate merges.

## Proposal (native to Kubescape)

Kubescape already supports `kubescape scan ... --exceptions <file.json>`. Expose it through ksail in the same spirit as the existing `spec.workload.validation.skipKinds`:

1. **CLI passthrough** — `ksail workload scan --exceptions <file>` forwards to Kubescape's `--exceptions`.
2. **Config** — a `spec.workload.scan` block in `ksail.yaml`, so `ksail workload scan` (no args) honours it automatically (mirrors how `validation.skipKinds` already works for `ksail workload validate`).

```yaml
# ksail.yaml
spec:
  workload:
    scan:
      framework: nsa
      exceptions: .kubescape/exceptions.json   # forwarded to kubescape --exceptions
      complianceThreshold: 100
```

With this, a repo can keep a checked-in Kubescape exceptions file for its runtime-enforced/justified controls and gate CI at `--compliance-threshold 100`, failing only on genuinely-new findings.

### Why this is the clean / native option

- Uses Kubescape's own exceptions format and flag — no bespoke ksail logic to maintain.
- Symmetric with the existing `validation.skipKinds` config surface, so `ksail workload scan` (no args) stays a turnkey CI gate.
- Lets `--compliance-threshold 100` become meaningful instead of permanently red.

## Context

Surfaced while making `ksail workload scan` a hard CI gate for a Flux + Kustomize platform repo where the large majority of remaining NSA findings are runtime-enforced (Kyverno mutation, Cilium network policies, VPA-managed resources) and therefore need exceptions rather than manifest changes.

---
_Filed via Claude Code._

devantler:

Fixes #5369

## Problem

`ksail workload scan` (Kubescape) had no way to apply **exceptions**, so it couldn't be used as a hard CI gate (`--compliance-threshold 100`): runtime-enforced controls (Kyverno-injected `securityContext`, VPA-managed resources, `CiliumNetworkPolicy`, …) that a *static* manifest scan can't see always reported as failures, so the threshold could never pass.

## What this does

Exposes Kubescape's native exceptions mechanism, mirroring the existing `spec.workload.validation.skipKinds` surface:

1. **CLI** — `ksail workload scan --exceptions <file>` forwards to Kubescape's `ScanInfo.UseExceptions` (the same field its own `--exceptions` flag sets). Thin passthrough, no bespoke ksail logic.
2. **Config** — a `spec.workload.scan` block so `ksail workload scan` (no args) is a turnkey CI gate:

```yaml
spec:
  workload:
    scan:
      frameworks: [nsa]
      exceptions: .kubescape/exceptions.json   # forwarded to kubescape --exceptions
      complianceThreshold: 100
```

Precedence mirrors `validation.skipKinds`: config supplies the default, an **explicitly-set flag always wins** (via `cmd.Flags().Changed`). An unreadable `ksail.yaml` warns (doesn't silently drop the gate). `exceptions` is env-var expanded like other path fields (e.g. `${HOME}/.kubescape/exceptions.json`).

## Two engineering calls vs the issue's literal sketch (please confirm)

- **`frameworks` (list) instead of `framework: nsa`** — the underlying `--framework` flag is a `StringSlice` (multiple frameworks), so a list field keeps config↔flag parity (`frameworks: [nsa]`). Happy to rename to singular if you'd rather match the sketch exactly.
- **`complianceThreshold` is an `int32` (0-100) in config**, not the flag's `float32` — `controller-gen` rejects floats in the CRD (discouraged for cross-language serialization), and gate thresholds are naturally whole percentages. The **CLI flag keeps `float32`** for ad-hoc fractional runs; the config int is converted on apply. Added CRD `minimum: 0` / `maximum: 100` validation as a bonus.

## Validation

- `go build ./...`, `go vet`, **golangci-lint clean**.
- **959 tests pass** across `pkg/cli/cmd/workload`, `pkg/apis/...`, `pkg/client/kubescape`. New tests: exceptions→`UseExceptions` passthrough; `--exceptions` flag default; config-precedence table (config fills / flag wins / no-config / unreadable-warns); a network-free integration guard that a config-supplied out-of-range threshold is rejected (proving config flows into validation).
- `make generate` artifacts regenerated and **idempotent** (re-running produced no further drift): CRD, `ksail-config.schema.json`, deepcopy, CLI-flag + declarative-config docs, chat docs, web/ui TS types, help snapshots.
- The env-var drift guard was updated: `Workload.Scan.Exceptions` → expanded list, `Workload.Scan.Frameworks[]` → skip list (consistent with `Validation.SkipKinds[]`).


devantler:

**CI blocker is infrastructural, not a code issue.** The only failing check, `🔍 Dead Code Analysis`, was **OOM-killed** (exit 143, ~4 min in — not the 15-min timeout). Root cause: `deadcode -test ./...` builds a whole-program SSA call graph that peaks at **~16.5 GB RSS** on the ksail module (measured locally), right at the ubuntu-latest **16 GiB** ceiling, so the host kills the runner. Locally `deadcode -test ./...` reports **0 findings, exit 0** — there is no dead code here.

This is a shared reusable-workflow gap: the adjacent govulncheck job already mitigates the identical failure mode with `GOMEMLIMIT`, but the deadcode job did not. Fix: **devantler-tech/reusable-workflows#329** adds `GOMEMLIMIT: 12GiB` + `timeout-minutes: 15` to the deadcode job. Re-running this check is retry-resistant (deterministic memory exhaustion), so the durable unblock is: merge #329 → reusable-workflows release → ksail re-pins `validate-go-project.yaml` (Dependabot) → this PR greens. No change needed to #5378 itself.

devantler:

Closing as **superseded by #5377** (merged 18:32Z, closing #5369), which I picked up autonomously in parallel without spotting the in-flight PR — my mistake for not checking for an existing open PR on the issue first.

No action needed: #5377 is the canonical implementation and is a bit more polished than this one. For the record, the two converged on the same key calls independently — notably `complianceThreshold` as an **int32** in config (controller-gen rejects floats in CRDs) while the `--compliance-threshold` flag keeps float32; #5377 goes one better with `*int32` (nil = no gate) and resolves a relative exceptions path against the ksail.yaml directory. Nothing here to salvage.

PR #5402, opened 2026-06-22
feat(ui): poll plugin K8s lists for live updates

devantler:
## What

Adds **live updates (polling)** to KSail's Headlamp-plugin Kubernetes data hooks so plugin resource lists stay current — a tractable, frontend-only substitute for Headlamp's WebSocket watch multiplexer. No backend changes.

## How

- **New `web/ui/src/lib/plugins/useAsyncList.ts`** — a shared `useAsyncList<T>(fetcher, deps): [T[], Error | null]` hook that both plugin K8s surfaces now delegate to, removing the duplicated one-shot-fetch logic. It:
  - holds `[items, error]` state and runs `fetcher` in a `useEffect` keyed on `deps`, with an `active` cancellation guard;
  - **resets `error` to `null` at the start of every fetch** (fixes a stale-error bug — the old per-surface hooks never cleared a transient failure, so one network blip stuck forever);
  - polls every `PLUGIN_LIST_POLL_INTERVAL_MS = 5000` after the initial load and clears the interval on cleanup;
  - **pauses polling while the tab is hidden** (`document.visibilityState === "hidden"`) and refetches immediately on `visibilitychange` back to visible (listener added/removed in the effect).
- **`web/ui/src/lib/plugins/k8s.ts`** — `makeResourceClass`'s `useList()` now delegates to `useAsyncList` (keyed on the cluster name/namespace primitives).
- **`web/ui/src/lib/plugins/pluginLib.ts`** — `makeK8sShim`'s `useResourceList(kind, namespace)` now delegates to `useAsyncList` (keyed on kind/namespace + cluster primitives).

Both call sites keep their **exact public signatures and return types** — plugins see no API change. An absent active cluster still yields an empty list.

## Scope

Polling-based list refresh only. The full Headlamp WS multiplexer (for plugins opening raw WS connections) is explicitly **out of scope**.

## Verification

`web/ui` is MegaLinter-excluded; only `tsc` + `vite build` gate it. Both pass:
- `cd web/ui && npx tsc --noEmit` → No errors found
- `npm run build` → built successfully

No unit-test harness exists in `web/ui` (no vitest/jest config, no `test` script, no `*.test.ts`), so per the task no test was added.


devantler:

**Heads-up: `main` is currently red on the jscpd 0% copy-paste gate, and this PR is the fix.**

The `chore(deps): Bump …/vcluster` push (#5400) ran the full `✅ Validate Go Project / 🧹 Lint - mega-linter` on `main` and it failed — jscpd found two clones (0.02% over the 0% threshold), both between the merged plugin-list files:

- `web/ui/src/lib/plugins/k8s.ts [105–118]` ↔ `pluginLib.ts [180–193]`
- `web/ui/src/lib/plugins/k8s.ts [121–133]` ↔ `pluginLib.ts [196–208]`

i.e. the identical cluster-keyed `useEffect` list-fetch body in `ResourceClasses.<Kind>.useList()` and `useResourceList()`. **This PR's `useAsyncList<T>` extraction deletes both duplicated bodies**, so merging it removes the clone. It also currently blocks the **v7.73.0 release (#5401)** and the **helm/v4 bump (#5398)**, whose `mega-linter` legs fail on the same finding.

One caveat on verification: `✅ Validate Go Project` (which hosts jscpd) is path-filtered to **Go** changes, so it shows **SKIPPED** on this TS-only PR — that same gap is why the clone slipped onto `main` (the TS PRs that introduced these files never triggered jscpd; only the Go push did). So this PR won't self-verify the gate; the red clears on the next Validate-Go-Project run after merge. Possible hardening follow-up: also run jscpd (or mega-linter) on `web/ui` changes via `web-ui.yaml` so TS clones are caught on their own PR — happy to open that as a separate `ci:` PR if useful.

(Confirmed your `useAsyncList` extraction is functionally identical to a dedup I independently prototyped while triaging the red — yours supersedes it, so I'm not opening a competing PR.)

Issue #4899, opened 2026-05-26
Operator-driven component-install lifecycle


devantler:
## Problem & goal

The CLI's `ksail cluster create` does two things: (a) **provision** the bare cluster, then (b) **install components** — CNI, CSI, CDI, metrics-server, cert-manager, load-balancer, policy-engine, and the GitOps engine — according to the `Cluster` spec.

The **operator today only does (a)**: `ClusterReconciler.reconcileNormal` calls `provisioner.Create/Exists/Update` and never installs components. So a `Cluster` resource that sets `spec.cluster.cni: Cilium` (or `gitOpsEngine: Flux`, etc.) is provisioned but those components are never installed.

**Goal:** make the operator reconcile the *full* desired state, installing/maintaining the components declared in the spec, idempotently and safely within the reconcile loop.

Scope: `internal/controller`, `pkg/operator`, `pkg/svc/installer`, `pkg/svc/provisioner/cluster`.

## Building blocks that already exist

- **`pkg/svc/installer.Factory`** (`NewFactory(helmClient, dockerClient, kubeconfig, kubecontext, timeout, distribution)`) with **`CreateInstallersForConfig(cfg *v1alpha1.Cluster) (map[string]Installer, error)`**. It already picks the right installer per spec enum (CNI=Cilium/Calico/None, GitOps=Flux/ArgoCD, …).
- **`installer.Installer`** interface: `Install(ctx) error`, `Uninstall(ctx) error`, `Images(ctx) ([]string, error)`.
- **`helm.NewClient(kubeconfigPath, kubeContext) (*helm.Client, error)`** — a helm client bound to a kubeconfig **file path** + context. Helm install/upgrade is internally idempotent.
- **`pkg/svc/detector`** — already used by the CLI `update` path to detect installed Helm releases.

These are service-layer packages (no Cobra/CLI coupling), so the operator can import them directly. The CLI's *orchestration* (`pkg/cli/setup.InstallCNI` / `InstallPostCNIComponents`) is `*cobra.Command`-coupled and **will not be reused**; we re-implement a thin, reconcile-friendly runner.

## The crux: obtaining a client to the *child* cluster

Installers run against the **provisioned (child)** cluster, not the hub. The operator must obtain a kubeconfig/helm client for the child. Today the `Provisioner` interface exposes **no** way to get this. Access differs per distribution × provider:

| Distribution × Provider | How the child is reached | Feasible from operator now? |
|---|---|---|
[...]

PR #5551, opened 2026-06-29
feat(operator): uninstall components removed from the cluster spec

devantler:

## Problem

The operator's reconcile loop installs/upgrades the components a `Cluster` spec declares (CNI, CSI, metrics-server, cert-manager, load-balancer, policy-engine, GitOps), but it **never uninstalls a component the spec drops**. Flipping `policyEngine: Kyverno` → `None` (or switching CNI `Cilium` → `Calico`) left the old component's Helm release **orphaned** in the child cluster. The operator only ever called `Install`, never `Uninstall` — exactly [#4899](https://github.com/devantler-tech/ksail/issues/4899)'s open question #2 (*"Phase 1 only installs/upgrades"*).

## Change

Drive the **full desired state** instead of an install-only convergence:

- **New baseline annotation** `ksail.io/last-applied-components` — reconciler-owned, written only after a successful component apply. It records the spec the operator last installed components for. It is deliberately **distinct** from `ksail.io/last-applied-spec` (the drift baseline, which the provisioner path rewrites *before* components reconcile — reusing it would race the component diff), and is stripped from client-submitted objects (`sanitizeForWrite`) like the spec baseline so a user cannot forge it.
- **Uninstall the removed set** — on each apply, rebuild the *previous* installer set from the baseline via the same `installer.Factory`, diff its keys against the desired set, and `Uninstall` the removed components in **reverse dependency order** (GitOps first, CNI last) before converging the desired set.
  - *Determining* what to remove is best-effort: a corrupt or unbuildable baseline is logged and swallowed (the next reconcile retries) so it can never block a fresh apply.
  - An actual *uninstall failure* is folded into the returned error → surfaces as `ComponentsReady=False` and requeues.
- **Status-safe baseline write** — `recordAppliedComponents` runs after runtime status has been observed this reconcile, so a plain `Update` would clobber `cluster.Status` (the API decodes the server's stored status back into the object). It writes through a deep copy and propagates only the new annotations + `resourceVersion` back, leaving the conditions/observed status set this reconcile intact for the single status update at the end of `reconcileNormal`.

## Scope / safety

- **No CRD/schema/flag/CLI-surface changes** — annotation + operator-internal logic only, so no generated-file regeneration.
- **Happy path unchanged**: first install has no baseline → empty removed-set → behaves exactly as before. Uninstall only fires when a prior baseline contains a component absent from the desired set, and only when the generation gate re-runs component reconciliation.

## Tests

- `runUninstallers`: reverse-order teardown; error aggregation continues past a stuck uninstall.
- `removedComponentInstallers`: dropped-vs-still-desired diff; no-baseline → empty; corrupt baseline → empty (never blocks).
- Controller: baseline recorded (full reconciled spec) after a successful apply and **not** on a skip; per-component status preserved across the mid-reconcile metadata write.

Validated locally: `go build ./...`, `go vet`, `go test ./pkg/operator/... ./internal/controller/...`, and `golangci-lint run` on the changed packages all clean.

Part of #4899

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -356,6 +356,25 @@ func (r *ClusterReconciler) reconcileComponents(
 		return true
 	}
 
+	// Record the spec we just reconciled components for, so a later spec change that drops a component
+	// (e.g. policyEngine Kyverno→None) is detected next reconcile and the component uninstalled. A
+	// write failure keeps ComponentsReady False so it retries — the baseline must exist before the
+	// condition reads True, or a removal could go undetected.
+	recordErr := r.recordAppliedComponents(ctx, cluster)
+	if recordErr != nil {
+		logf.FromContext(ctx).
+			Info("record components baseline (best-effort)", "error", recordErr.Error())
+		setCondition(
+			cluster,
+			v1alpha1.ConditionComponentsReady,
+			metav1.ConditionFalse,
+			"BaselineRecordFailed",
+			recordErr.Error(),
+		)
+
+		return false
+	}
+
 	setCondition(
 		cluster,
 		v1alpha1.ConditionComponentsReady,
--- a/internal/controller/cluster_controller_test.go
+++ b/internal/controller/cluster_controller_test.go
@@ -2,6 +2,7 @@ package controller_test
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"slices"
 	"strings"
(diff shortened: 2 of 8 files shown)


devantler:

Addressed CodeRabbit's **outside-diff-range Major** (baseline factory built from the current cluster): `removedComponentInstallers` now receives a factory-builder closure and rebuilds the previous installer set with a factory for the **last-applied spec's distribution**, so a distribution change no longer computes the uninstall set against the wrong distribution. Pinned with a K3s-baseline → Vanilla-current regression test (35 pkg tests green, lint clean).

Issue #5620, opened 2026-07-01
feat(cloudinit): carry declarative apt sources, packages, and files in the bootstrap transport


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

The `cloudinitbootstrap` transport (`pkg/svc/bootstrap/cloudinit`) currently renders only a **boot script + `runcmd`** from `Config.Commands`. It cannot yet carry the *declarative* install primitives cloud-init natively supports — `packages:`, `apt: sources:` (with a signing key trusted declaratively), and arbitrary `write_files:`.

Those are exactly what the Vanilla × Hetzner (kubeadm) provisioner needs to deliver the declarative install rendered for a node — installing packages and configuring the pkgs.k8s.io apt repo through cloud-init's own modules instead of a `curl | sh` runcmd, per the settled native-Go provisioning design (native Go renders the typed config; declarative cloud-init installs via `packages:`/`write_files:`, **never** `curl|sh`).

## Proposed direction

Extend the pure `cloudinitbootstrap.Config` / `BuildUserData` (transport-agnostic; no coupling to any bootstrapper's `Install` type) to also render:

- **`Packages []string`** → cloud-init `packages:` (apt install before `runcmd`)
- **`AptSources []AptSource`** → cloud-init `apt: sources:` map, each with a `source:` line and an optional inline armored `key:` (the signing key is trusted **declaratively**, no runtime `curl|gpg`)
- **`Files []File`** → additional `write_files:` entries (e.g. a kubeadm config dropped `0600`)

Keep the existing command→boot-script behaviour byte-for-byte for command-only configs (new fields are `omitempty`, so existing output is unchanged). Deterministic YAML (sorted apt-source keys). Validate each new field (single-line/no-NUL where applicable, absolute file paths, non-empty apt name+source); a fully-empty config still returns `ErrNoCommands`.

## Acceptance criteria

- `Config` gains `Packages`, `AptSources`, `Files`; `BuildUserData` renders `packages:`, `apt: sources:`, and extra `write_files:`.
- Command-only output is unchanged (existing tests pass untouched).
- New unit tests cover rendering + determinism + validation for each field; pure (no I/O), lint-clean, no generated drift.

## Scope / non-goals

- The provisioner mapping (a bootstrapper's `Install{AptSources,Packages,Files,Commands}` → this `Config`) and the `UserDataProvider` seam extension are **later** slices — this is the transport primitive only.

Part of #5513 / #3983.

Issue #5616, opened 2026-07-01
feat(bootstrap): render declarative kubeadm node install (repo + packages + bootstrap)


devantler:
> 🤖 Generated by the Daily AI Assistant

Child of #5513 (Vanilla/kubeadm × Hetzner provisioner), part of epic #3983.

## Problem

`pkg/svc/bootstrap/kubeadm` renders the typed kubeadm **configuration** (`Render` → InitConfiguration/ClusterConfiguration/JoinConfiguration), and its `doc.go` states that the Kubernetes **components** (kubeadm, kubelet, the container runtime) are installed at first boot "by the cloud-config … `packages:`/`write_files:` — never `curl … | sh`". But no code yet renders that install. The k3s sibling has `install.go` (`Render` → the install command); kubeadm has no equivalent, so the Vanilla × Hetzner provisioner (#5513) has no install layer to compose.

## Proposal

Add a pure `RenderInstall(InstallConfig) (Install, error)` to `kubeadmbootstrap` — the **declarative** sibling of the k3s install renderer. Given the Kubernetes version (whose minor track selects the `pkgs.k8s.io` community repository) and the rendered kubeadm config, it returns a transport-agnostic `Install`:

- **`AptSources`** — the Kubernetes community repo for the version's minor track, with the `Release.key` URL so cloud-init's `apt.sources` module trusts it declaratively (no imperative `curl | gpg --dearmor`).
- **`Packages`** — `containerd` + `kubelet`/`kubeadm`/`kubectl`.
- **`Files`** — the rendered kubeadm config dropped at a fixed `/etc/kubernetes/ksail` path (root-only; it carries the bootstrap token).
- **`Commands`** — enable the runtime, `apt-mark hold` the kube* packages, then `kubeadm init` (control-plane-init) or `kubeadm join` (joiners).

Pure (no I/O/network), fully unit-tested, and independent of the topology planner (#5611/#5612) — it renders install for a single node's config.

## Out of scope (later #5513 increments)

Container-runtime CRI configuration (SystemdCgroup), CNI install, and kubeconfig fetch are post-install provisioner-lifecycle concerns (as they are for the k3s path), rendered by later slices. Wiring `Install` into the cloud-init transport (extending it with `apt.sources`/`packages`/`write_files`) and composing it in a `vanillahetzner` provisioner are the following increments.

## Acceptance criteria

[...]

Issue #5611, opened 2026-06-30
feat(bootstrap): kubeadm topology planner (Plan)


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #5513 (Vanilla (kubeadm) × Hetzner provisioner), child of #3983 (epic #4627).

## Problem
The `kubeadmbootstrap` package can `Render` a **single** node's typed kubeadm config (`InitConfiguration`/`ClusterConfiguration` or `JoinConfiguration`, #5521), but there is no topology layer that expands a cluster's node counts into the ordered per-node configs a provisioner bootstraps in sequence. The K3s sibling already has this seam — `k3sbootstrap.Plan(PlanInput) → []Node` — which the merged `k3shetzner` provisioner (#5594) composes. The Vanilla provisioner needs the same precursor.

## Proposal
Add `kubeadmbootstrap.Plan(PlanInput) → ([]Node, error)`, mirroring `k3sbootstrap.Plan`:
- expands a topology (control-plane count, agent count, cluster-wide options, shared token) into the ordered nodes — cluster-initialising control plane (`RoleServerInit`) → additional control planes (`RoleServer`) → agents (`RoleAgent`);
- every returned `Node.Config` passes validation, so `Render` never fails for a planned node;
- pure (no I/O, no network), fully unit-testable.

Unlike k3s's pre-shared token, kubeadm discovers a joining node's **API server endpoint** and **CA cert hashes** only once the first control plane is up, so those are `PlanInput` fields a provisioner injects at run time before planning the joining nodes (validated up-front when joiners are present).

## Acceptance criteria
- [ ] `Plan` expands a topology into ordered, role-correct, renderable `NodeConfig`s.
- [ ] Discovery fields (endpoint + CA hashes) are required and validated only when joining nodes are present.
- [ ] Pure + unit-tested (happy paths, every error path, slice-clone independence, Render-never-fails contract).

Issue #5513, opened 2026-06-28
feat(provisioner): Vanilla (kubeadm) × Hetzner provisioner


devantler:
> 🤖 Generated by the Daily AI Assistant

Child of #3983 (epic #4627). Depends on the remote-exec transport (and parallels #5510 for kubeadm).

## Problem
Per the maintainer's confirmed direction on #3983 (Decision 2), **"Vanilla" on a server = upstream Kubernetes via kubeadm** — Kind remains the *Docker-provider* implementation of the Vanilla distribution (Kind has no remote/server mode). No kubeadm bootstrap path exists.

## Proposal
1. A kubeadm install-command/script renderer (sibling to #5510): containerd + kubeadm install, `kubeadm init` on the first control-plane, `kubeadm join` for additional control-plane + workers, CNI install, kubeconfig fetch.
2. Make the Vanilla provisioner `ProviderAware`; on Hetzner, provision servers and drive the kubeadm bootstrap over the remote-exec transport.

## Acceptance criteria
- [ ] `ksail cluster create --distribution Vanilla --provider Hetzner` provisions a working kubeadm-based upstream Kubernetes cluster on Hetzner servers.
- [ ] Kind remains the Vanilla × Docker implementation, unchanged.
- [ ] Renderer is pure + unit-tested; provisioner lifecycle consistent with the other Hetzner paths.

devantler:
> 🤖 Generated by the Daily AI Assistant

Started this issue with the **pure config renderer** foundation slice: draft #5521 adds `pkg/svc/bootstrap/kubeadm` (`Render(NodeConfig) → kubeadm.k8s.io/v1beta4` Init/Cluster/Join documents per role), the Vanilla sibling of the k3s renderer (#5516). Shares the `server-init`/`server`/`agent` role vocabulary so the topology plan (#5519) can drive either distribution. Pure, 98.8% covered, lint/build/test clean.

Remaining for this issue (follow-up slices): the cloud-init install package list (apt repo + kubeadm/kubelet/containerd) and the K3s/Vanilla × Hetzner provisioner wiring (#5512). The `supportedProviders` validation flip lands last in #5514.

devantler:
> 🤖 Generated by the Daily AI Assistant

Closing as **done by events** — every AC is code-complete on `main`, delivered across the #3983 foundation train:

[...]

PR #5623, opened 2026-07-01
feat(bootstrap): render containerd config (systemd cgroup driver) for kubeadm nodes

### LINKED ISSUE #5622: feat(bootstrap): render containerd config (systemd cgroup driver) for kubeadm nodes

**Part of #5513 / #3983** (Vanilla/kubeadm × Hetzner provisioner).

## Problem

A kubeadm node brought up on a raw VM (the Vanilla × Hetzner path) needs its
container runtime (containerd) configured before `kubeadm init`/`join` will
produce a healthy node. The single most common failure is a **cgroup-driver
mismatch**: kubeadm configures the kubelet to use the `systemd` cgroup driver
(the default since kubeadm 1.22), but a stock containerd install defaults to
`SystemdCgroup = false` (cgroupfs). When the two disagree the kubelet fails to
manage pods and the node never reaches `Ready`. Secondarily, containerd's CRI
`sandbox_image` (the "pause" image) should match the one kubeadm/kubelet
expects, or every pod sandbox triggers an extra image pull and a mismatch
warning.

Today there is **no containerd configuration rendering anywhere** in ksail's
bootstrap or provisioner packages (`pkg/svc/bootstrap/{kubeadm,k3s,cloudinit}`,
`pkg/svc/provisioner/cluster/*`) — confirmed absent on `main`. The kubeadm
bootstrap renders only the kubeadm YAML; the container runtime is left to its
install-time defaults, which are wrong for a systemd-cgroup kubeadm node.

## Proposed direction

Add a small, pure native-Go renderer — a sibling of the existing
`kubeadmbootstrap`/`k3sbootstrap`/`cloudinitbootstrap` packages — that produces
the `/etc/containerd/config.toml` a kubeadm node needs:

- new package `pkg/svc/bootstrap/containerd` (`containerdbootstrap`);
- `RenderContainerdConfig(ContainerdConfig) (string, error)` returning a
  deterministic, managed-file-headed containerd **v2** config that enables the
  **systemd cgroup driver** (`SystemdCgroup = true` under the CRI runc runtime)
  and optionally pins `sandbox_image`;
- pure/no-I/O, fully unit-testable, delivered on the node declaratively via
  cloud-init `write_files:` (the transport capability added in #5620) — never a
  runtime `curl … | sh`, per the native-Go provisioning direction.

This is the container-runtime configuration slice the `kubeadmhetzner`/
`vanillahetzner` provisioner (a later #5513 increment) will fold into the node's
cloud-init alongside the kubeadm config (#5616) and the declarative install
(#5620). It is **orthogonal** to the topology planner (#5611) and install
renderer (#5616) — it imports none of them — so it advances #5513 without
stacking on those in-flight slices.

## Rough size

Small. One pure renderer + validation + table-driven tests. No CRD/schema/flag/
CLI/docs surface, no new dependency.

## Acceptance criteria

- [ ] `RenderContainerdConfig` renders valid containerd v2 TOML with
  `SystemdCgroup = true` on the CRI `runc` runtime.
- [ ] `sandbox_image` is pinned when supplied and omitted (containerd default)
  when empty; a malformed image reference is rejected at render time.
- [ ] Output is deterministic and carries the ksail managed-file header.
- [ ] Pure (no I/O), covered by table-driven tests; build/vet/golangci/deadcode
  all clean with no generated drift.

devantler:

Fixes #5622 · **Part of #5513 / #3983** (Vanilla/kubeadm × Hetzner provisioner).

## What

Adds `pkg/svc/bootstrap/containerd` (`containerdbootstrap`) — a pure native-Go
renderer for the `/etc/containerd/config.toml` a kubeadm (Vanilla) node needs so
its container runtime agrees with the way kubeadm configures the kubelet. It is a
sibling of the existing `kubeadmbootstrap` / `k3sbootstrap` / `cloudinitbootstrap`
renderers.

```go
func RenderContainerdConfig(ContainerdConfig) (string, error)
```

## Why

kubeadm configures the kubelet to use the **systemd cgroup driver** (the default
since kubeadm 1.22), but a stock containerd install defaults to `cgroupfs`
(`SystemdCgroup = false`). When the runtime and the kubelet disagree, the kubelet
cannot manage pod cgroups and the node **never reaches `Ready`** — the single most
common failure of a hand-rolled kubeadm-on-containerd node. There is no containerd
config rendering anywhere in ksail today (confirmed absent on `main`), so this
renderer removes that mismatch before the node ever boots. It also optionally pins
the CRI `sandbox_image` (the "pause" image) so pod sandboxes don't trigger an extra
pull and a version-mismatch warning.

## Design notes

- **containerd v2 config**, `SystemdCgroup = true` on the CRI `runc` runtime
  (`runtime_type = io.containerd.runc.v2`).
- The cgroup driver is **fixed to systemd, not configurable** — a ksail-provisioned
  kubeadm node must match kubeadm's kubelet default, and a knob that could only ever
  be `systemd` would only invite misconfiguration.
- `SandboxImage` is **optional**: pinned when supplied, omitted (containerd default)
  when empty; a malformed reference (space, control char, quote) is rejected at
  render time via `ErrInvalidSandboxImage`.
- **Pure & deterministic** (no I/O, byte-identical output), delivered on the node
  declaratively via cloud-init `write_files:` (the transport capability added in
  #5620) — never a runtime `curl … | sh`, per the native-Go provisioning direction.
- **Orthogonal**, not stacked: imports neither the topology planner (#5611) nor the
  kubeadm install renderer (#5616), so it advances #5513 independently of those
  in-flight slices. A later provisioner increment folds all three into the node's
  cloud-init.

## Dependency note

Promotes `github.com/BurntSushi/toml v1.6.0` (already vendored transitively) from an
**indirect to a direct test dependency** — the test **decodes** the rendered
document and proves it is valid TOML, rather than substring-matching. `go.sum`
unchanged; no new supply-chain surface.

## Validation

- `go test ./pkg/svc/bootstrap/containerd/...` — **10 tests pass, 100% coverage**.
- `go build -o /tmp/ksail .` — clean; `go test ./pkg/svc/bootstrap/...` — 76 pass.
- `golangci-lint run --timeout 5m` — 0 issues; `deadcode -test` — clean.
- No generated drift (`git status` shows only `go.mod` + the new package).

Opened as a draft for your review/promotion.


devantler:

Good catch — a lone backslash starts a TOML escape sequence, so it would corrupt the double-quoted `sandbox_image` string. Fixed in `6253ed04`: `validateSandboxImage` now rejects `\` alongside quotes, whitespace, and control characters, and the malformed-image test table gains a `backslash` case (`registry.k8s.io\pause:3.10`). Package tests + golangci-lint green.

PR #5652, opened 2026-07-01
fix(provider/hetzner): reconcile subnet when network already exists

### LINKED ISSUE #5651: fix(provider/hetzner): EnsureNetwork skips subnet reconciliation when network already exists

## Problem

`Provider.EnsureNetwork` in [`pkg/svc/provider/hetzner/infrastructure.go`](https://github.com/devantler-tech/ksail/blob/main/pkg/svc/provider/hetzner/infrastructure.go) is **not idempotent across a partial failure**:

```go
network, _, err := p.client.Network.GetByName(ctx, networkName)
...
if network != nil {
    return network, nil   // <-- returns early, never checks the subnet
}
// ... only the create path adds the subnet via AddSubnet(...)
```

If a first `Create` creates the network but then `AddSubnet` fails (transient hcloud API error, rate limit, context cancel), a later `Create` **retry** calls `GetByName`, finds the existing network, and returns it immediately — **the subnet is never added**. The cluster is left wedged in a partial-provisioned state: servers attach to a network with no subnet, and the retry cannot heal it.

This is a pre-existing gap in the **shared** Hetzner provider infra, so it affects **every** Hetzner provisioner (`k3shetzner`, `talos`-hetzner, and the new `kubeadmhetzner`). Surfaced by CodeRabbit on #5649 (the kubeadmhetzner provisioner seam), but the fix belongs in the shared provider, not that PR.

## Proposed direction

Make the "network already exists" branch **reconcile the subnet**: after `GetByName` returns an existing network, check whether it already carries the expected subnet (`network.Subnets`), and `AddSubnet` it if missing — reusing the same CIDR-derivation the create path uses. Extract the subnet-derivation + add into a small helper so both paths (fresh-create and adopt-existing) share it and stay in sync. Keep it idempotent (adding a subnet that already exists must be a no-op / tolerated).

## Acceptance criteria

- `EnsureNetwork` adds the missing subnet when it adopts an existing network that has none.
- Adding a subnet is a no-op when the network already has the expected subnet (idempotent, no duplicate-subnet error surfaced).
- Regression test with the fake hcloud client: network exists **without** the subnet → `EnsureNetwork` results in the subnet being present (simulating the AddSubnet-failed-then-retried path).
- No behavior change to the fresh-create path.

## Size

Small (S) — one function + a shared helper + one focused test in `pkg/svc/provider/hetzner`.

devantler:

Fixes #5651

## Problem

`Provider.EnsureNetwork` returned an existing network **early** without ensuring its server subnet:

```go
if network != nil {
    return network, nil   // subnet never checked
}
// ...only the create path added the subnet
```

If a first `Create` created the network but then `AddSubnet` failed (transient hcloud error, rate limit, context cancel), a later `Create` **retry** found the existing network and returned it **subnet-less** — a partial-provisioned state that never healed (servers attach to a network with no subnet). This is shared Hetzner provider infra, so it affected **every** Hetzner provisioner (k3s, Talos, and the new kubeadm). Surfaced by CodeRabbit on #5649.

## Change

- Extract subnet handling into an idempotent `ensureSubnet` helper called by **both** the fresh-create and adopt-existing paths.
- `ensureSubnet` skips when the network already carries the expected subnet and adds it otherwise, so a retry reconciles a missing subnet.
- No behaviour change to the fresh-create path (a freshly-created network has no subnets, so the helper adds it exactly as before).

## Tests

New `infrastructure_reconcile_test.go` (httptest mock of the Hetzner API, reusing the existing `newTestHcloudClient` pattern):

- **`ReconcilesMissingSubnetOnExisting`** — network exists with `subnets: []`, so `EnsureNetwork` issues one `AddSubnet` (heals the partial state). Mutation-checked: neutering `ensureSubnet` fails this test.
- **`SkipsSubnetWhenPresent`** — network already has `10.0.1.0/24`, so **no** `AddSubnet` call (idempotent).

## Validation

`go build` / `go vet` / `gofmt` clean; full `pkg/svc/provider/hetzner` test package green; `golangci-lint run --new-from-rev=origin/main ./pkg/svc/provider/hetzner/...` reports 0 issues. Pure library change — no CLI/schema/apis/generated-surface drift.

### DIFF
--- a/pkg/svc/provider/hetzner/infrastructure.go
+++ b/pkg/svc/provider/hetzner/infrastructure.go
@@ -58,6 +58,43 @@ func (p *Provider) EnsureNetwork(
 	return network, nil
 }
 
+// defaultServerSubnetCIDR is the historical server subnet carved out of the
+// default network range; kept so existing default-CIDR clusters reconcile
+// unchanged.
+const defaultServerSubnetCIDR = "10.0.1.0/24"
+
+// subnetMaskBits is the prefix length of the server subnet derived from a
+// custom network range.
+const subnetMaskBits = 24
+
+// ipv4Bits is the bit length of an IPv4 mask, used to recognise IPv4 ranges.
+const ipv4Bits = 32
+
+// serverSubnetCIDR derives the server subnet from the network CIDR. The
+// default network keeps its historical 10.0.1.0/24 subnet; a custom network
+// uses its first /24 (a Hetzner subnet must sit inside the network range, and
+// the full range would leave no room for further subnets). A custom range
+// that is already /24 or smaller — or not IPv4 — is used whole.
+func serverSubnetCIDR(cidr string) (string, error) {
+	if cidr == v1alpha1.DefaultHetznerNetworkCIDR {
+		return defaultServerSubnetCIDR, nil
+	}
+
+	_, ipRange, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return "", fmt.Errorf("invalid network CIDR %s: %w", cidr, err)
+	}
+
+	ones, bits := ipRange.Mask.Size()
+	if bits != ipv4Bits || ones >= subnetMaskBits {
+		return ipRange.String(), nil
+	}
+
+	firstSlash24 := net.IPNet{IP: ipRange.IP, Mask: net.CIDRMask(subnetMaskBits, bits)}
+
+	return firstSlash24.String(), nil
+}
+
 // ensureSubnet adds the server subnet to the network when it is not already present.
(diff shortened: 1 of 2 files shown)


devantler:

Addressed CodeRabbit's **outside-diff-range Major** (custom network CIDRs used whole as the server subnet): `ensureSubnet` now derives the **first /24** of a custom range via `serverSubnetCIDR` (whole range only when the network is already /24-or-smaller or non-IPv4; the default network keeps its historical `10.0.1.0/24`), and the idempotency check now compares against the derived value. Added regression tests for custom-/16 add + skip and /24 whole-range cases (285 pkg tests green, lint clean).

PR #5723, opened 2026-07-02
feat(hetzner): generate the Talos VIP machine-config for the floating-IP endpoint

### LINKED ISSUE #5718: feat(hetzner): generate the Talos VIP machine-config for the floating-IP endpoint (node-side ownership handover)

## Problem

With `floatingIPEnabled` (#5714), KSail provisions the floating IP, attaches it to the first control-plane server, and renders it as the cluster endpoint + certificate SANs — but **no node answers on the address** until the OS configures it. On Talos + Hetzner the mechanism is the native VIP support: a `machine.network.interfaces[].vip` block with `hcloud.apiToken`, which makes the elected etcd leader claim the floating IP (this is exactly the handover `EnsureFloatingIP`'s doc comment anticipates). Until KSail generates that block, an enabled cluster's stable endpoint only works if the operator supplies the VIP block as their own machine-config patch (documented limitation in #5714's slice).

## Proposal

When `FloatingIPEnabled` is set, generate a control-plane-scoped machine-config patch:

- `machine.network.interfaces` entry selecting the public interface — design decision needed: `deviceSelector` (e.g. `physical: true` / bus path) vs a hardcoded interface name; hcloud VMs are predictable but a selector is robust across server types.
- `vip: { ip: <floating-ip>, hcloud: { apiToken: <token> } }` — the token comes from the configured `tokenEnvVar`; note it becomes part of the rendered machine config (same trust surface as Talos docs prescribe for hcloud VIP; call it out in docs). Consider a dedicated read-write-floating-IP-scoped token option later.
- Ordering: the patch needs the floating IP's address, which is only known after `EnsureFloatingIP` — same regenerate-with-preserved-PKI flow as `WithEndpoint`/`WithCertSANs` (#5714).
- Once the VIP owns attachment, the explicit `AttachFloatingIPToServer` on create becomes just the initial claim (Talos reasserts it on leader changes).

## Acceptance criteria

- Enabled clusters answer on the floating IP without any user-provided patch; a control-plane roll moves the address to a surviving leader (E2E — env-gated on HCLOUD_TOKEN like the rest of the Hetzner lane, #4972).
- Disabled ⇒ no VIP block rendered (byte-identical configs).
- Unit tests for the generated patch content and the enabled/disabled gating.
- Docs: the #5714 limitation note is replaced by the native behaviour description.

Parent lane: devantler-tech/platform#2120 (option A) → #5698 (primitives, merged #5699) → #5714 (config surface + rendering). Rough size: M.

devantler:

## Why

Part of giving Hetzner clusters a stable, permanent address (platform#2120): today, if the control-plane node holding the cluster's floating IP is replaced, the address goes dead until someone reattaches it. The cluster should keep itself reachable without intervention.

## What

Talos clusters on Hetzner now claim the floating IP themselves: whichever control-plane node is currently the leader takes ownership of the address, automatically, across node replacements. Opt-in via the existing floating-IP setting; nothing changes for clusters not using it.

Fixes #5718 (next in the lane: roll this out to the production platform once released).


devantler:

**CodeRabbit review-body nitpicks resolved** (no threads exist; this comment is the resolution record; pushed in c96c754 + 4896455):
1. ✅ **Token fail-fast (the real half of the test finding):** the hcloud token is now validated *before* `ensureFloatingIPEndpoint` — previously a pure config error (unset token env var) left cloud side effects behind (floating IP ensured + attached). `TestUpdateConfigsWithEndpoint_FloatingIPEnabledTokenUnset` now also asserts **zero** hcloud assign calls.
2. ❌ **Refuted: `buildHetznerVIPPatch` struct + `yaml.Marshal`.** The file's established dynamic-patch style is string-built YAML with `%q` interpolation (`buildCertSANsPatch` is the direct precedent); converting one builder would leave two styles in the same file, and a file-wide builder refactor is behaviour-preserving work that must not be mixed into this feature PR (contract rule). Happy to file it as a separate refactor issue on request.

Issue #5726, opened 2026-07-02
feat(hetzner): resolve boot images and derive live server specs for K3s/Vanilla (flips live bring-up on)


devantler:
> 🤖 Generated by the Daily AI Assistant

Child of #3983, sibling of #5725 (whose first increment wires the live bring-up orchestration + kubeconfig persistence into `hetznerbase.RunCreate` behind a caller-supplied `BringUpPlan`). After that increment, both distro callers still stop at `ErrLiveBringUpNotImplemented` inside their `composePlan` because neither can derive full `CreateServerOpts` yet.

## Problem

Live bring-up needs fully-resolved server specs, and two pieces are missing:
1. **Boot-image resolution** — `CreateServerOpts.ImageID` must come from somewhere; nothing in `pkg/svc/provider/hetzner` resolves a stock OS image (e.g. `ubuntu-24.04`) to its ID. Needs `GetByNameAndArchitecture` (arch derived from the configured server type) on the provider seam + mock.
2. **Spec derivation & bootstrap material in the callers** — kubeadmhetzner has `DeriveServerSpecs`+`Infra` but `composePlan` never fills them; k3shetzner has no spec derivation at all. Both must generate the bootstrap SSH keypair (`sshbootstrap.GenerateKeyPair`), thread its public half into the cloud-init `ssh_authorized_keys`, decide the host-key trust policy (pin pre-generated host keys via cloud-init `ssh_keys` + `gossh.FixedHostKey` — no insecure callback), and return a complete `BringUpPlan` (specs, signer, host-key callback, remote kubeconfig path: k3s `/etc/rancher/k3s/k3s.yaml`, kubeadm `/etc/kubernetes/admin.conf`).

## Acceptance criteria

- `hetzner.Provider` resolves the default stock image for the configured server type's architecture (seam + mock + tests).
- Both callers' `composePlan` return a complete single-node `BringUpPlan`; `ErrLiveBringUpNotImplemented` is retired.
- Host keys pre-generated and pinned (no `InsecureIgnoreHostKey` anywhere).
- Unit tests cover spec derivation, keypair threading, and host-key pinning; live E2E stays with #5515/#4972.

devantler:
> 🤖 Generated by the Daily AI Assistant

[...]

Issue #5725, opened 2026-07-02
feat(hetzner): wire the bring-up engine into RunCreate (single-node, node-IP endpoint) and persist the kubeconfig


devantler:
> 🤖 Generated by the Daily AI Assistant

Child of #3983 (K3s/Vanilla on Hetzner). Follows #5720/#5721 (bring-up engine), #5712/#5713 (SSH key delivery), #5696/#5697 (sshbootstrap).

## Problem

`hetznerbase.RunCreate` composes infrastructure, node token, and user_data, then stops at `ErrLiveBringUpNotImplemented` — the shipped `BringUpNode` engine (create server → SSH → wait for kubeconfig → read it) is never called, and no kubeconfig is persisted. The blocker was the API-endpoint/TLS-SAN design decision: user_data is composed before any server IP exists.

## Design decision (endpoint ladder)

- **This slice: first-node public IPv4.** Both k3s and kubeadm automatically include the node's own IPs in the API server's serving-cert SANs (k3s: node IPs + localhost; kubeadm: default-route IP), so a single-node cluster's kubeconfig can be rewritten to the server's public IPv4 **after** creation with no SAN threading in user_data at all. This unblocks live bring-up (#5515 smoke lane) with the smallest diff.
- **Follow-up slice: floating-IP stable endpoint** (mirrors the Talos lane's `ensureFloatingIPEndpoint` and the platform #2120 direction; primitives `EnsureFloatingIP`/`AttachFloatingIPToServer` already shipped in #5699/#5719). Needs (a) `--tls-san <fip>` (k3s) / `ControlPlaneEndpoint`+`certSANs` (kubeadm) threaded into user_data — the FIP is creatable **before** composition, so the IP is known — and (b) binding the FIP on the node's interface via cloud-init (the k3s/kubeadm analogue of #5718's Talos VIP patch). Filed separately once this lands.

## Proposal

In `hetznerbase.RunCreate` (single-node path only; multi-node stays gated behind `ErrMultiNodeNotImplemented`):
1. Call `BringUpNode` with the composed spec after infrastructure + user_data are ready.
2. Rewrite the returned kubeconfig's server URL to `https://<server-public-IPv4>:6443`.
3. Persist via the house pattern (`k8s.MergeKubeconfig`, as Talos' `writeKubeconfig` does), logging the canonical path.
4. Keep the engine's failure-cleanup semantics (server deletion on bring-up failure) intact.

## Acceptance criteria

[...]

PR #5727, opened 2026-07-02
feat(hetzner): run the live bring-up from RunCreate and persist the kubeconfig

devantler:

## Why

The Hetzner cloud-provisioning epic (#3983): KSail could already build the cloud servers, but 'create cluster' didn't finish the job — it stopped before making the new cluster reachable from the user's machine, so there was no usable end-to-end path.

## What

Completes the create flow's backbone: bring the node up, wait for the cluster to be ready, and install a working connection (kubeconfig) pointing at the cluster's public address — with cleanup if anything fails midway. The final user-visible switch-on ships in the follow-up (#5726), which adds the remaining server-configuration pieces.

First increment of #5725.

### DIFF
--- a/pkg/svc/provisioner/cluster/factory_k3d.go
+++ b/pkg/svc/provisioner/cluster/factory_k3d.go
@@ -120,6 +120,7 @@ func createK3sHetznerProvisioner(cluster *v1alpha1.Cluster) (Provisioner, any, e
 
 	provisioner, err := k3shetznerprovisioner.NewProvisioner(
 		cluster.Name,
+		cluster.Spec.Cluster.Connection.Kubeconfig,
 		k3sInstallVersion(),
 		controlPlanes,
 		int(cluster.Spec.Cluster.Workers),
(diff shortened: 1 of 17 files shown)


devantler:

**CodeRabbit review-body nitpicks resolved** (no threads exist, so this comment is the resolution record; pushed in a8c12a2):
1. `rewriteKubeconfigEndpoint` now rejects a parsed-but-clusterless kubeconfig (`ErrKubeconfigNoClusters`) instead of silently persisting it with no endpoint rewritten — valid: `clientcmd.Load` tolerates a file read mid-write. Covered by `TestRunCreateClusterlessKubeconfigFails` (also asserts nothing is persisted).
2. Merge behaviour pinned: `TestRunCreateMergesIntoExistingKubeconfig` seeds the destination with an unrelated cluster and asserts both entries coexist after `RunCreate`.

devantler:

Resolution record for the CodeRabbit review-body findings (no threads exist for these):

- **Guard against a kubeconfig with no cluster entries** — already fixed in the earlier sweep: `ErrKubeconfigNoClusters` guards the parse path (kubeconfig.go) with a dedicated test.
- **Merge-into-non-empty-destination coverage** — already added in the earlier sweep (the preexisting-cluster merge test).
- **Missing cleanup-on-failure assertion for the clusterless-kubeconfig case** — valid; fixed in the just-pushed commit: the test now asserts `deleteNodesCalls == 1` like its parse-/persist-failure siblings. Package tests + lint green.

PR #5749, opened 2026-07-03
feat(cloudinit): deliver a pre-generated SSH host identity via ssh_keys

devantler:

## Why
The Hetzner live bring-up (#5726) must not trust a node's SSH host key on first contact — the dial needs a host identity it already knows, which means delivering pre-generated host keys to the node at boot.

## What
Adds host-identity delivery to the cloud-init composer (the `ssh_keys:` module, ed25519, both halves required together, validated + tested). This is the missing primitive for #5726's host-key pinning; the composePlan wiring that consumes it follows separately.

Part of #5726

### DIFF
--- a/pkg/svc/bootstrap/cloudinit/cloudinit.go
+++ b/pkg/svc/bootstrap/cloudinit/cloudinit.go
@@ -70,6 +70,14 @@ type Config struct {
 	// would create a server that never bootstraps, so a keys-only Config is
 	// still rejected with [ErrNoCommands].
 	SSHAuthorizedKeys []string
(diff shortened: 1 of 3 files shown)


devantler:

CodeRabbit review-body nitpick (cloudinit_test.go 302-331, add a NUL-in-public host-key case) — valid, fixed: `TestBuildUserDataHostKeyErrors` now covers the `\x00`-in-`ED25519Public` path alongside the newline case (57 package tests green). This comment is the resolution record for that body finding.

devantler:
@coderabbitai review

> 🤖 Generated by the Daily AI Assistant — re-requesting review of the new head commit (b3893e8) so the stale status context clears.

Issue #5755, opened 2026-07-03
feat(hetzner): multi-node bring-up — two-phase join sequencing for K3s/Vanilla × Hetzner


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #3983 (epic #4627). Sibling of #5515 (smoke tests — that issue is the *validation* lane; this is the *implementation* lane).

## Problem
The K3s × Hetzner (#5512) and Vanilla/kubeadm × Hetzner (#5513) provisioners bring up a **single control-plane node only**. `hetznerbase.Base.RunCreate` rejects every multi-node topology up front:

```go
if b.ControlPlanes > 1 || b.Agents > 0 {
    return ErrMultiNodeNotImplemented
}
```

and `bringUpFromPlan` asserts `len(plan.Specs) == 1`. The bootstrap layer is already multi-node-ready — `k3sbootstrap.Plan` expands an ordered init→servers→agents plan and threads a `ServerURL` into every joining node — but the **provisioner never supplies that `ServerURL`**, because a joining node registers against the first node's address, which is only known **at runtime** once that node is created and Hetzner assigns its private-network IP. So the composition passes an empty `ServerURL` and multi-node is refused before composition.

(Note: the current `ErrMultiNodeNotImplemented` message and several comments say "tracked by #5515", but #5515 is the env-blocked *smoke-test* issue, not the sequencing implementation. This issue is that missing tracker.)

## Proposed direction — runtime two-phase join sequencing
Split the create flow into two phases inside `hetznerbase`:

1. **Init phase (unchanged):** compose + bring up the cluster-initialising control plane (`ServerURL=""`), wait for its admin kubeconfig, and derive the **join endpoint** from the created server's *private-network* address (`https://<private-ip>:<distro-port>` — nodes join over the cluster's private network, which the firewall already permits; k3s: `:6443`).
2. **Join phase (new):** re-compose the joining nodes' cloud-init `user_data` with the resolved join endpoint (the `CreateStrategy` gains a runtime "compose joiners given a serverURL" seam), create them in bootstrap order, then persist the init node's kubeconfig (endpoint-rewritten to its public IPv4, as today).

Design decisions to make (a draft PR is where they get expressed/redirected):
- **Join address = the init CP's private-network IP** (not public) — intra-cluster join over the private network.
[...]

Issue #5515, opened 2026-06-28
test(system): smoke tests for Hetzner × K3s/Vanilla (blocked on #6361)


devantler:
> 🤖 Generated by the Daily AI Assistant

Child of #3983 (epic #4627). **Blocked** on the Hetzner system-test lane (#4972).

## Problem
The Hetzner × K3s / Vanilla paths need CI smoke coverage, but the Hetzner system-test lane currently fails on an invalid/absent `HCLOUD_TOKEN` (#4972), so these tests cannot go green until that is restored.

## Proposal
Add a smoke system test per new combo (K3s × Hetzner, Vanilla × Hetzner): create → assert reachable → delete, on the Hetzner system-test lane. Land once #4972 is resolved.

## Acceptance criteria
- [ ] A smoke system test for each implemented Hetzner combo.
- [ ] Tests green on the Hetzner lane (requires #4972 fixed first).
- [ ] Cleanup-on-failure so no leaked Hetzner servers.

---

## Blocker corrected 2026-08-22

`#4972` is **closed**, but it was closed by *pausing* the weekly schedule, not by rotating the
credential — so an actionability check that resolves the reference above finds a closed issue and
reads this as startable. It is not. The live blocker is the restoration issue:

**Unblocked 2026-09-13:** the maintainer rotated `HCLOUD_TOKEN` (now an org-wide secret for a Hetzner test project), and a green Hetzner system-test run is on record: https://github.com/devantler-tech/ksail/actions/runs/34774720413 (dispatched from the ksail#7018 fix branch, cleanup succeeded).

`System Test - Hetzner` had no successful run on record before that (six consecutive scheduled
failures 2026-05-31 → 2026-07-05), and its `schedule` trigger was removed on 2026-07-09.

Read every `#4972` reference above as `#6361`. That condition, a green Hetzner system-test run on record, was met on 2026-09-13.

PR #5757, opened 2026-07-03
feat(hetzner): multi-node bring-up for K3s × Hetzner (control plane + agents)

devantler:

## Why
KSail can bring up a K3s cluster on Hetzner today, but only a **single node** — any agents were refused outright. Multi-node is the point of a real cloud cluster, and it was the next unblocked step in the Hetzner provider epic (#3983).

## What
A K3s × Hetzner cluster with `controlPlanes: 1` and `agents: N` now brings up the control plane and joins N agents to it over the private network. Multi-control-plane (HA) is still rejected for now (a follow-up), and the Vanilla/kubeadm path is unchanged (its multi-node lands next). The existing single-node behaviour is untouched.

First increment of #5755; end-to-end join is verified by the env-gated Hetzner smoke tests (#5515) once that CI lane is restored.

Part of #5755. Part of #3983.

### DIFF
--- a/pkg/svc/provisioner/cluster/internal/hetznerbase/hetznerbase.go
+++ b/pkg/svc/provisioner/cluster/internal/hetznerbase/hetznerbase.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/devantler-tech/ksail/v7/pkg/apis/cluster/v1alpha1"
 	"github.com/devantler-tech/ksail/v7/pkg/svc/provider/hetzner"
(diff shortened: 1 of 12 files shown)


devantler:

**CodeRabbit review-body nitpick (init-node failure coverage) — addressed in b04ef049.**

Added `TestRunCreateMultiNodeComposeInitErrorFailsFast`, which sets `composeInitErr` and drives `RunCreateMultiNode`.

One correction to the suggested assertion: `ComposeInitNode` runs *before* any server is created (it composes the init spec that `bringUpInitControlPlane` then provisions), so a compose-init failure returns directly without reaching the cleanup path. The correct behavior is **fail-fast** — the test therefore asserts `createServerCalls == 0` and `deleteNodesCalls == 0` (no paid resource exists yet), rather than a triggered cleanup. `cleanUpFailedBringUp` only fires once the control plane is up, which the missing-private-IP and compose-join tests already cover.

Issue #5344, opened 2026-06-19
Extend `workload validate` and `workload scan` to validate/scan all GitOps layers in-process


devantler:
> 🤖 Drafted by Claude Code at the maintainer's request, grounded in a read-through of the `workload validate`/`scan` source and `go.mod` (`main` @ `5dfa22fa`).

## Problem & goal

`ksail workload validate` and `ksail workload scan` are the shift-left quality/security gate for every KSail-managed GitOps repo. But a GitOps repo expresses desired state at **three abstraction layers**, and only one of them is fully expanded before these commands inspect it. The result: the Helm- and operator-rendered manifests — i.e. most of what Flux actually applies — are never validated or scanned offline.

**Goal:** make `validate` and `scan` reason about all three layers **in-process**, so a single `ksail workload validate` / `ksail workload scan` gives full-fidelity coverage. Consumers should not have to bolt `flux-local` + `helm template` + `kubeconform` + a policy engine into their own CI to scan the real output — KSail already owns the Kustomize build, the Flux substitution, the schema validation, and the cluster lifecycle, so it is the natural place to consolidate this.

## The three layers, and where KSail stands today

| Layer | Expanded by | Offline-renderable? | `validate` today | `scan` today |
|---|---|---|---|---|
| **1. Kustomize overlays** | `kustomize build` (pure fn) | Yes, deterministic | ✅ built in-process (krusty) | ⚠️ see note |
| **2. Helm via `HelmRelease`** | helm-controller (`helm template`) | Yes (needs chart + values) | ❌ **not rendered** — HelmRelease validated as an opaque CR | ❌ not rendered |
| **3. Operator CRs** (CNPG `Cluster`, KEDA `ScaledObject`, Flagger `Canary`, `ExternalSecret`, cert-manager `Certificate`, …) | the operator controller, at runtime | **No offline renderer exists** | ⚠️ CR **shape** only (schema) | ⚠️ CR shape only |

Grounded in the source:

[...]

PR #5773, opened 2026-07-03
feat(validate): wire native CEL-rule validation into `workload validate`

### LINKED ISSUE #5772: Wire native CEL-rule validation into `workload validate` (--rules flag)

## Problem

The CEL rule engine foundation (`pkg/svc/gitops/.../celrules`, merged via #5707) can load a YAML rules file, compile CEL expressions, and evaluate them against decoded documents with per-rule severity — but it is **not wired into the `workload validate` command**. Today `workload validate` only runs kubeconform + Helm render + Flux substitution; there is no way for a user to run their own CEL policy rules over the rendered GitOps layers.

This is the next child of the #5344 in-process validation roadmap (native CEL-rule validation, tracked in #5606's child-2 direction: "converge on CEL, stay-native").

## Proposed direction

Wire the existing engine into the command, end-to-end:
1. Add a `--rules <file>` flag to `workload validate` (`pkg/cli/cmd/workload/validate.go`).
2. Load + compile the rules (existing `celrules` loader/engine) once, before the validation loop.
3. Evaluate each decoded document against the rules inside the per-kustomization validation path, reusing the render **provenance/attribution** already threaded through the pipeline so a violation names its source layer/HelmRelease.
4. Surface CEL violations alongside kubeconform errors in the command output, honoring per-rule severity (error fails validation, warning does not).
5. Docs: extend `docs/.../workload-validate.mdx` + regenerate the CLI-flags reference.

## Acceptance criteria

- `ksail workload validate --rules rules.yaml` runs the CEL rules over rendered manifests and fails on an `error`-severity violation, passes on `warning`-only.
- Violations are attributed to their source document/layer.
- Unit tests cover: flag parsing, rule load failure surfacing, an error violation failing validation, a warning violation not failing, and no-`--rules` behaviour unchanged.
- No external creds / live cluster needed (pure in-process, like the rest of `workload validate`).

## Rough size

Small–medium. Reuses the merged engine; the work is flag + load + eval-in-loop + violation surfacing + docs.

Part of #5344.

devantler:

## Why

`ksail workload validate` could render and schema-check GitOps manifests, but there was no way to run your own policy checks over them — even though the CEL rule engine to do so already shipped (#5707). Teams that want to enforce project conventions (naming, required labels, forbidden kinds, image-tag policy) had no in-tree, cred-free way to gate on them.

## What

Adds a `--rules <file>` flag to `workload validate`. It evaluates each rule in a YAML CEL rules file against every rendered/built document: an **error**-severity violation fails validation (attributed to the offending resource), a **warning**-severity violation is reported without failing. Pure in-process — no external credentials or live cluster, like the rest of `validate`.

Reuses the existing engine; the config-field equivalent and HelmRelease-layer (not just resource-identity) attribution are noted as follow-ups.

Fixes #5772. Part of #5344.


devantler:

Addressed the CEL skip-filter finding (both the inline comment on `celvalidate.go` and the `⚠️ Outside diff range` finding on `validate.go:739-749` — same root cause) in `c250c218`.

**Fix:** `evaluateCELDocuments` now takes the skip-kinds set and filters each document *before* `engine.Evaluate`, reusing the exact `opts.SkipKinds` list kubeconform already applies (populated from `--skip-secrets`, `--skip-kinds`, and `spec.workload.validation.skipKinds`). Both call sites (`validateSilent`, `validateFileSilent`) now pass it, so CEL honors the same exclusions as kubeconform on both paths.

**Tests:** added paired positive/negative coverage — a default-skipped Secret and a `--skip-kinds ConfigMap` are not evaluated, while the same resources with skipping disabled still fail (rule name + resource attribution asserted).

devantler:

**Both failing checks (`🧪 Test` + `📊 Code Coverage`) have the same single cause — a stale tool-surface snapshot, not a logic bug.**

```
--- FAIL: TestToolSurfaceSnapshot/workload_read
    pkg/toolgen/toolsurface_snapshot_test.go:81
```

The new `--rules` flag on `workload validate` changed the MCP tool surface, and `pkg/toolgen`'s snapshot wasn't regenerated (the help snapshot was, the toolgen one was missed). The Coverage job runs the same suite, so it fails on the same test — one fix greens both.

**Fix (one command + commit):**
```
UPDATE_SNAPS=true go test ./pkg/toolgen/...
# commit the updated pkg/toolgen/**/__snapshots__/*.snap
```

I diagnosed this but am holding off on pushing it myself: this branch is being co-driven (the `--skip-kinds`/`--skip-secrets` CEL fix landed here), and the local ksail submodule worktree isolation regressed to a stray `core.worktree` mid-session — so a safe isolated regen+push isn't available to me right now without risking a collision. Flagging it so it can be greened in one step.

devantler:

**Re: CodeRabbit ⚠️ Outside-diff finding — "CEL should honor skip filters" (`validate.go:739-749`)** — verified against current code; **already addressed, refuting as a false positive on the current diff.**

Both CEL call sites now pass the resolved skip set into the helper:
- `validate.go:747` and `:873` → `evaluateCELDocuments(..., opts.SkipKinds, ...)`.
- `opts.SkipKinds` is populated from **all three** exclusion sources before CEL runs: `--skip-secrets` appends `"Secret"` (`validate.go:218-219`), `--skip-kinds` appends the user list (`:226`), and the `ksail.yaml` `spec.workload.validation.skipKinds` equivalents are appended (`:227-229`).
- `evaluateCELDocuments` builds a skip-set and `continue`s past any document whose `kind` is in it **before** `engine.Evaluate` (`celvalidate.go:74-91`).

This is covered by dedicated tests in `validate_celrules_test.go`: a Secret skipped by the default `--skip-secrets` is not evaluated by CEL, and with `--skip-secrets=false` the same Secret **is** evaluated and fails its rule. The finding appears to have been anchored to an earlier revision of the branch before the skip list was threaded through; no code change needed.

(Separately: pushed `71ee160` regenerating the `pkg/toolgen` tool-surface snapshot for the new `--rules` flag, which was the Test/Coverage failure.)

PR #5785, opened 2026-07-04
feat(operator): implement Connector for nested KWOK clusters

### LINKED ISSUE #5778: operator: implement Connector for Kind & KWOK (DinD) child kubeconfig access

Part of #4899 (operator-driven component-install lifecycle). Decomposes the next Phase-2 tranche.

## Problem

The operator installs a `Cluster`'s components against the **child** cluster, obtaining child access via the `Connector` interface (`pkg/svc/provisioner/cluster`, `Kubeconfig(ctx, name) ([]byte, error)`). The **Secret-published** distributions already implement it and are done:

- vCluster + Kubernetes — reads `vc-<name>` Secret, rewrites server to the in-cluster Service address.
- k3k (K3s + Kubernetes) — reads `k3k-<name>-kubeconfig` Secret.
- Talos (nested + Kubernetes) — reads `talos-<name>-kubeconfig` Secret in `ksail-<name>`.

The **DinD-based** Kubernetes-provider distributions — **Kind** and **KWOK** — do **not** yet implement `Connector`, so `InstallComponents` type-asserts `Connector`, finds it absent, and **skips component install** for them (logged). Their child kubeconfig is written as a **file inside the DinD pod**, not published as a hub Secret, so it is not operator-reachable today.

## Proposed direction

Give Kind and KWOK the same operator-readable-Secret shape the other nested distros already use:

1. **Publish** the child kubeconfig as a Secret in the hub during provisioning — after the DinD cluster is up, extract its kubeconfig from the DinD pod and `Apply` a `<distribution>-<name>-kubeconfig` Secret (idempotent upsert), with the API-server address rewritten to the operator-reachable endpoint (the Gateway API `TCPRoute`/NodePort the DinD API server is exposed through — same exposure the status observer already relies on).
2. **Implement `Connector.Kubeconfig`** on the Kind and KWOK provisioners to read that Secret, rewrite/verify the server address, and return `clustererr.ErrKubeconfigNotReady` when the Secret is not yet published (so the operator requeues). This read side mirrors the vCluster/k3k/Talos implementations (share via `internal/nested` where possible).

## Acceptance criteria

- [ ] Kind and KWOK provisioners publish a `<distribution>-<name>-kubeconfig` Secret (idempotent) with the API server rewritten to the operator-reachable endpoint.
- [ ] Both implement `Connector`; `InstallComponents` no longer skips them (the "no operator-reachable kubeconfig" log path is not hit for Kind/KWOK).
- [ ] Unit tests for the **read** side follow the vCluster pattern (fake host clientset seeded with the Secret): asserts server rewrite, and `ErrKubeconfigNotReady` when the Secret is missing/empty.
- [ ] Unit tests for the **publish** side (fake clientset / exec) assert the Secret is created with the rewritten server and upserts idempotently.

## Notes / sequencing

- The **read** side is fully unit-testable now (vCluster test pattern). The **publish + end-to-end** path is only verifiable once the kind-in-kind dev hub has the Gateway API CRDs and DinD exposure wired — so land the code with unit coverage; E2E validation follows as the environment becomes testable (same constraint #4899 records for the DinD distros).
- Cloud (EKS/GKE) and Hetzner (SSH-fetched) Connectors are separate later tranches, not in scope here.

Scope: `pkg/svc/provisioner/cluster/{kind,kwok}`, `pkg/svc/provisioner/cluster/internal/nested`, `pkg/operator`.

devantler:

## Why

The operator installs a cluster's components into the child cluster only for distributions that can hand back an operator-reachable kubeconfig (the `Connector` capability). The nested **KWOK** (DinD) distribution didn't implement it, so KWOK-on-Kubernetes clusters silently had their component installs skipped.

## What

Adds the KWOK Connector, completing the Kind + KWOK pair started by #5779. KWOK now publishes its child kubeconfig at create time and serves it back to the operator, so the operator installs components into nested KWOK clusters like the other nested distributions. Internal change only — no new flags, config, or CRDs.

Fixes #5778
Part of #4899

### DIFF
--- a/pkg/operator/components.go
+++ b/pkg/operator/components.go
@@ -42,7 +42,7 @@ var installOrder = []string{
 //
 // Child access is obtained through the provisioner's optional clusterprovisioner.Connector
 // capability: provisioners that can hand back an operator-reachable kubeconfig (the nested
-// Kubernetes-provider distributions — vCluster, k3k, Talos, and Kind — publish one) get their
+// Kubernetes-provider distributions — vCluster, k3k, Talos, Kind, and KWOK — publish one) get their
 // components installed; others are a no-op until they implement Connector.
 //
 // The returned applied is false when installation was skipped (no Connector), so the reconciler can
--- a/pkg/svc/provider/kubernetes/provisioner_config.go
+++ b/pkg/svc/provider/kubernetes/provisioner_config.go
@@ -1,0 +1,35 @@
+package kubernetes
+
+import (
+	"github.com/devantler-tech/ksail/v7/pkg/apis/cluster/v1alpha1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// DinDProvisionerConfig holds the host-cluster wiring shared by the DinD-based nested provisioners
+// (Kind, KWOK): both run their nested cluster inside a Docker-in-Docker pod on the same host
+// cluster, so they take the same host clients, Kubernetes provider, and nested-cluster identity.
+// It is built once (see the factory's buildDinDProvisionerConfig) and embedded in each
+// distribution's provisioner config so the two never drift and the assembly lives in one place.
+type DinDProvisionerConfig struct {
+	// KubeconfigPath is the path to the nested cluster's kubeconfig.
+	KubeconfigPath string
+	// HostClientset is the Kubernetes clientset for the host cluster, used to publish and read the
+	// nested cluster's kubeconfig Secret for the operator Connector contract.
+	HostClientset kubernetes.Interface
+	// K8sProvider is the Kubernetes infrastructure provider.
+	K8sProvider *Provider
+	// DynamicClient is the dynamic client for Gateway API resources.
+	DynamicClient dynamic.Interface
+	// RestConfig is the REST config for port-forwarding to the DinD pod.
+	RestConfig *rest.Config
+	// ClusterName is the nested cluster name (used for namespace, labels).
+	ClusterName string
+	// Distribution is the distribution name (for labels).
+	Distribution string
+	// GatewayClassName is the Gateway class for API exposure (empty = no gateway).
+	GatewayClassName string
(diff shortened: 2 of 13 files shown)


devantler:

Addressed the review-body findings in `f294ed8e`:

- **🧹 Nitpick — decouple the KWOK naming prefixes:** applied. Split `kwokDistributionPrefix` (kubeconfig-Secret distribution prefix) from `kwokContextPrefix` (kwokctl context prefix) so the two — currently the same value but separate concepts — can diverge without silently affecting each other.

Also resolved the `🧹 Lint - mega-linter` (jscpd) failure at the root cause: the KWOK connector duplicated Kind's DinD config literal and kubeconfig-publish wrapper. Extracted the shared host-cluster wiring into `kubernetesprovider.DinDProvisionerConfig` (built once by `buildDinDProvisionerConfig`, embedded in both provisioner configs) and the extract+publish flow into `nested.PublishConnectorKubeconfig`, so the two DinD distributions share one implementation. jscpd now reports 0 clones; golangci-lint, gofmt, and 1465 unit tests green.

devantler:

**Re: 🧹 Nitpick — params struct for `PublishConnectorKubeconfig`'s five string args** (`internal/nested/kubeconfig.go` L78-95)

Considered and **keeping the positional signature** — the suggestion is sound in isolation, but it conflicts with a *required* check on this PR. The two call sites (`kind`/`kwok` `publishConnectorKubeconfig`) are intentionally shaped so the only diverging argument — `kindKubeconfigKey` vs `kwokKubeconfigKey` — sits at **position 3**, which is precisely the jscpd duplicate-code break added in `f294ed8e` (mega-linter flagged the Kind↔KWOK duplication and hard-fails on it).

A named params struct would move that divergence to the first struct field and leave the remaining five (`KubeconfigPath`/`ContextName`/`Namespace`/`SecretName`/`Labels`) as a byte-identical contiguous block across both call sites — reintroducing the jscpd clone I just resolved. Trading a required green check for a trivial transposition-safety gain isn't worth it here. Both call sites are verified correct.

PR #5787, opened 2026-07-04
chore(compliance): regenerate THIRD_PARTY_LICENSES and automate its refresh

### LINKED ISSUE #5716: chore(compliance): THIRD_PARTY_LICENSES is stale (last regenerated 2025-05-13) — regenerate and automate refresh

## Problem

`THIRD_PARTY_LICENSES` at the repo root declares **Last regenerated: 2025-05-13** — over a year of dependency changes ago. It still names removed deps (e.g. `github.com/ProtonMail/go-mime`, `github.com/ProtonMail/gopenpgp/v2` are being swapped out by the omni/client 1.9.0 bump in #5711) and misses newer ones. CI's `go-licenses check` gates license *types* but nothing regenerates the *inventory*, so the compliance metadata silently drifts on every bump. Surfaced by CodeRabbit on #5711 (scoped there to the ProtonMail swap; the real gap is the file has no automated refresh at all).

## Proposal

1. Regenerate the inventory now (`go-licenses report`/save per the file's own provenance header) for both modules (root + `desktop/`).
2. Automate the refresh so it can't go stale again — preferred: fold it into the existing `📤 Auto-Commit Generated Changes` self-heal in ci.yaml (it already regenerates snapshots/generated files on any PR); alternative: a scheduled job or a CI check that fails when the inventory drifts from `go mod graph`.
3. Update the provenance header to reflect the automated flow.

## Acceptance criteria

- `THIRD_PARTY_LICENSES` matches the current go.mod/go.sum (root + desktop).
- A dependency bump PR that changes the module graph gets the inventory refreshed automatically (or fails a check telling it to).
- No hand-edited generated content.

Rough size: S–M (one PR).

devantler:

## Why

The third-party license inventory had not been regenerated for over a year, so the compliance metadata shipped with every release names removed dependencies and misses new ones — and nothing prevented it drifting further.

## What

Regenerates the inventory (now covering both Go modules), checks in the previously-missing generator so regeneration is reproducible, and adds a CI self-heal that refreshes the file automatically on any dependency change. A new dependency that ships no license file now fails CI until its license is manually verified and recorded.

Fixes #5716


devantler:
I believe Dependency Review is doing the same thing as this license action and related tooling. Please verify, and if I am correct we should be able to remove it in favor of the GHAS native Dependency Review action and it's license checks.

devantler:

Verified against the [dependency-review-action docs](https://github.com/actions/dependency-review-action) — **you're right about the gate, but the action can't cover this PR's artifact**, so the split I've shipped is:

**What Dependency Review does cover:** an allow/deny **license gate on PR-introduced dependency changes** — the same role as CI's existing `license-check` job (`go-licenses check`). Filed **#5788** to adopt the native action and retire that custom job after a short soak (its allowlist + the documented `--ignore` risk-accepts port over via `allow-licenses` / `allow-dependencies-licenses`).

**What it can't cover (why this PR stays):**
- It produces **no third-party notice file** — nothing generates or refreshes `THIRD_PARTY_LICENSES` (the stale-inventory problem #5716 is about; MIT/BSD/Apache attribution obligations make the notice file worth keeping for a distributed binary).
- It is **fail-open on undetected licenses** (docs: *"If we can't detect the license for a dependency we will inform you, but the action won't fail"*), and its data source is the dependency graph/ClearlyDefined rather than the module's own LICENSE files — the generator here reads source truth and **hard-fails** on any unverified Unknown module.

So: this PR resolves the inventory gap (#5716); #5788 executes your direction for the gate — native Dependency Review in, custom `license-check` out.

devantler:

CodeRabbit review-body nitpicks drained in b2ce8ebe: keyed `verification` literals; sibling packages of one repo now share a single verification via `verifyAll` (also keeps `verifiedUnknown` within funlen); 15-min run-level subprocess timeout; `dependency.module` documented as the go-licenses-emitted import path; `writeUnknownSection` fails loud on the (checkUnknowns-guarded) unverified path; ci.yaml added to the `licenses` change filter. Regeneration verified byte-identical — no inventory churn.

devantler:
We do not need the third party license file, as the found dependency are submitted via dependency submission from GHAS. So I belive this functionality can be fully removed!

devantler:

Closing per your direction above: with the inventory covered by GHAS dependency submission, the notice file — and with it this PR's generator, `make licenses`, and the `verify-licenses` job — is not needed. The replacement is #5789: native Dependency Review license gate in, custom `license-check` job out, `THIRD_PARTY_LICENSES` deleted (it also closes #5716 and #5788).

PR #5809, opened 2026-07-05
feat(mirror): add the Phase 2 intercept tunnel frame codec

### LINKED ISSUE #5808: feat(mirror): Phase 2 intercept — tunnel frame codec (multiplexing wire format)

**Part of #4521** (local-remote service mirroring — Phase 2 *intercept*).

## Problem
Phase 1 (mirror-only) is complete: inbound traffic is captured read-only and replayed to the local process (`workload mirror`, last increment #5794's `--to` live replay). Phase 2 — *intercept* — must **return the local process's responses back into the cluster**, which the read-only mirror path deliberately cannot do (`pkg/svc/mirror/doc.go` Phase 2 note; `inject.go` "the reverse tunnel [is] the next … increment").

Intercept steers a subset (or all) of a Deployment's connections to the local process and pipes each connection's responses back. Many concurrent intercepted connections must share the **one** bidirectional byte stream the exec channel gives us (SPDY stdin+stdout) — so Phase 2 needs a **connection-multiplexing framing protocol** over that single duplex stream before any steering or agent work can begin.

## Proposed direction
Ship the foundational piece first: a small, self-framing **frame codec** in `pkg/svc/mirror` (matching the flat-package convention of `replay.go`/`session.go`), independent of any cluster or elevated privilege and fully unit-testable — exactly the increment shape of the Phase 1 primitives.

Wire format (big-endian, 9-byte header): `StreamID uint32 | Type uint8 | Length uint32 | Payload[Length]`. Frame types: `Open` / `Data` / `Close`. Control frames (Open/Close) carry no payload; Data payload is bounded by a constant so a malformed/hostile length cannot allocate unbounded memory (same defensive stance as `maxPendingSegments` in the replay parser).

`WriteFrame(w, frame)` / `ReadFrame(r)` with sentinel errors for oversized payloads, unknown frame types, and control frames carrying a payload; a clean `io.EOF` at a frame boundary vs `io.ErrUnexpectedEOF` mid-frame.

## Acceptance criteria
- [ ] `pkg/svc/mirror` gains a frame codec: `Frame`, `FrameType` (Open/Data/Close), `WriteFrame`, `ReadFrame`, with a bounded max payload constant.
- [ ] Round-trip, multi-stream interleaving, truncation (mid-header and mid-payload), oversized-length rejection, unknown-type rejection, and payload-on-control-frame rejection are unit-tested (table-driven, no cluster).
- [ ] `go build`, `go test ./...`, `golangci-lint run` green; docs/`doc.go` note the Phase 2 tunnel foundation.

Follow-on Phase 2 children (not this issue): the mux **session** exposing frames as `net.Conn`-like streams; connection **steering** (iptables/NET_ADMIN in the tap container); the in-cluster **intercept agent**.

devantler:

## Why
The `workload mirror` intercept phase (#4521 Phase 2) needs to send the local process's replies back into the cluster — something the read-only mirror path deliberately can't do. Doing that means carrying many intercepted connections at once over the single bidirectional channel KSail already has (the exec stream), which requires a multiplexing wire format before any traffic-steering work can start.

## What
Adds that foundation: a small, self-contained frame codec (open/data/close frames tagged by connection) with encode/decode and hardening against malformed input. Pure Go, fully unit-tested, no cluster or elevated privileges — the same increment shape as the earlier mirror pieces. No user-facing behaviour or flags change yet; later increments (the connection multiplexer, traffic steering, the in-cluster agent) build on it.

Fixes #5808
Part of #4521

### DIFF
--- a/pkg/svc/mirror/doc.go
+++ b/pkg/svc/mirror/doc.go
@@ -59,7 +59,15 @@
 //     cluster.
 //
 //   - Phase 2 — intercept: steer a subset (or all) of the Deployment's traffic to
-//     the local process and return its responses.
+//     the local process and return its responses. Unlike mirror mode, responses
+//     must flow back into the cluster, so intercept needs a reverse tunnel: the
+//     one bidirectional exec byte stream (SPDY stdin+stdout) must carry many
+//     concurrent intercepted connections at once. The foundation increment is
+//     the multiplexing wire format — [Frame], [WriteFrame], and [ReadFrame] —
+//     a self-framing, length-prefixed codec that tags each chunk with a
+//     StreamID so the mux/demux (a later increment) can present each
+//     intercepted connection as its own stream. Steering (iptables/NET_ADMIN in
+//     the tap container) and the in-cluster intercept agent follow.
 //
 //   - Phase 3 — environment & volume projection: run the local process with the
 //     target pod's env vars and mounted volumes/secrets for cluster-equivalent
--- a/pkg/svc/mirror/tunnel.go
+++ b/pkg/svc/mirror/tunnel.go
@@ -95,9 +95,13 @@ func checkFrameLength(frameType FrameType, length uint32) error {
 // (unknown type, oversized Data payload, or a control frame with a payload) is
 // rejected before anything is written.
 func WriteFrame(writer io.Writer, frame Frame) error {
-	// len is non-negative and, for a valid Data frame, well under 2^32;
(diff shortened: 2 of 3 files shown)


devantler:

Fixed in 58d3325e: `WriteFrame` now bounds `len(frame.Payload)` against `MaxTunnelPayload` on the un-narrowed `int` before converting to `uint32`, so an oversized payload can no longer wrap into range and desync the header from the written bytes. 66 pkg tests + lint green.

PR #5811, opened 2026-07-05
feat(gke): implement the operator Connector capability

### LINKED ISSUE #5810: Implement the operator Connector capability for GKE clusters

**Part of #4899** (Phase 4 — cloud Connectors, tranche 1).

## Problem

The operator installs components into a child cluster only when its provisioner implements the optional `clusterprovisioner.Connector` capability (`pkg/operator/components.go`). All nested Kubernetes-provider distributions now implement it (vCluster, k3k, Talos, Kind, KWOK), but the cloud provisioners do not — a GKE `Cluster` CR is provisioned and then component installation is silently skipped (`ComponentsReady=Unknown`).

## Proposal

Implement `Kubeconfig(ctx, name)` on `gkeprovisioner.Provisioner`:

1. Resolve location (existing `resolveLocation`) → `client.GetCluster`.
2. Gate readiness: status `RUNNING` + non-empty `Endpoint` + `MasterAuth.ClusterCaCertificate`; otherwise return `clustererr.ErrKubeconfigNotReady` so the reconciler requeues.
3. Build a minimal kubeconfig: `server=https://<Endpoint>`, CA from the base64 `ClusterCaCertificate`, bearer token minted from an `oauth2.TokenSource` (injectable option; defaults to Google ADC with the cloud-platform scope — workload identity / mounted SA key in the operator pod). Short-lived tokens are fine: the operator calls `Kubeconfig` per reconcile and installers consume it immediately.

GKE is the right first cloud tranche because `pkg/client/gke` already exposes `GetCluster` returning the full `containerpb.Cluster` and has the `WithClusterManager` fake seam; EKS's eksctl wrapper exposes no endpoint/CA and needs a separate design (aws-sdk `DescribeCluster` + STS presign) — follow-up tranche.

## Acceptance criteria

- `gkeprovisioner.Provisioner` satisfies `clusterprovisioner.Connector`; no factory changes needed (capability is type-asserted).
- Unit tests via the fake cluster-manager + a static token source: ready path yields a parseable kubeconfig with rewritten server/CA/token; `PROVISIONING`/empty-endpoint → `ErrKubeconfigNotReady`; missing cluster → `ErrClusterNotFound`; token/CA errors propagate.

Rough size: S–M.

devantler:

## Why

A GKE cluster managed by the operator is provisioned but then silently gets no components installed (CNI, GitOps engine, …) because the GKE provisioner doesn't expose child-cluster access — the last remaining gap class in the operator lifecycle epic now that all nested distributions have Connectors.

## What

Implements the operator's Connector capability for GKE: it builds a ready-gated, operator-usable kubeconfig straight from the GKE API (cluster endpoint + CA + a short-lived Google-credentials token), so the operator can install the declared components and correctly requeues while the cluster is still provisioning. Fully unit-tested against the existing fake SDK seam; no new dependencies.

Fixes #5810 · Part of #4899

### DIFF
--- a/pkg/svc/provisioner/cluster/gke/connector.go
+++ b/pkg/svc/provisioner/cluster/gke/connector.go
@@ -7,6 +7,7 @@ import (
 
 	"cloud.google.com/go/container/apiv1/containerpb"
 	"github.com/devantler-tech/ksail/v7/pkg/svc/provisioner/cluster/clustererr"
+	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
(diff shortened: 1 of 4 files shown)


devantler:

Both review findings addressed in c1e7094: the default ADC token source is now resolved once per provisioner under a mutex (a failed discovery is retried, not cached), and the repeated static test token source is folded into a `staticTokenSource()` helper. 29 pkg tests + lint green.

Issue #5441, opened 2026-06-24
feat(cluster): scaffold and grow multi-cluster, multi-provider GitOps repos


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

`ksail cluster init` scaffolds **one** cluster for **one** provider, and `ksail tenant create` scaffolds a tenant. But a real platform is inherently **multi-cluster / multi-provider** (e.g. a local Docker dev cluster + a prod cloud cluster sharing one base), and ksail has no first-class way to scaffold or grow that shape. Consumers fall back to hand-maintained Kustomize/Flux boilerplate and "copy-the-prod-folder" recipes.

This is drawn from a real consumer — the `devantler-tech/platform` repo (Talos+Docker local, Talos+Hetzner prod, shared base, Flux GitOps). Its README literally asks: *"If you know of a different way to manage kustomize and flux kustomizations that results in less boilerplate code, please let me know."*

## Evidence — what the platform hand-maintains today

In `devantler-tech/platform`:

- **Sentinel + per-cluster `replacements:` boilerplate.** `k8s/clusters/base/` defines shared Flux Kustomizations with `__CLUSTER__` / `__PROVIDER__` sentinels; each environment overlay (`k8s/clusters/local/kustomization.yaml`, `k8s/clusters/prod/kustomization.yaml`) repeats a near-identical ~80-line `replacements:` block to rewrite those sentinels. Adding a cluster means copying that block.
- **Per-cluster bootstrap kustomization that differs by one line.** `k8s/clusters/local/bootstrap/kustomization.yaml` vs `…/prod/bootstrap/kustomization.yaml` are identical except for the provider path (`providers/docker/bootstrap` vs `providers/hetzner/bootstrap`).
- **Per-cluster Flux timeout patches.** Each overlay re-declares strategic-merge patches that override `spec.timeout` on the four Flux Kustomizations (`bootstrap → infrastructure-controllers → infrastructure → apps`).
- **A documented "fork + edit N inputs" / "cp -R" onboarding.** `docs/TEMPLATING.md` enumerates the ~6 inputs a fork must edit; adding an environment is described as *"`cp -R k8s/clusters/prod k8s/clusters/<env>` … `cp ksail.prod.yaml ksail.<env>.yaml` … edit."*

None of this is provider- or app-specific business logic — it's the cross-environment **wiring** that ksail already understands conceptually (it owns `distribution`, `provider`, `kustomizationFile` per config).

[...]

PR #5863, opened 2026-07-06
feat(environment): surface declared environments as a read model

### LINKED ISSUE #5862: feat(environment): surface declared environments (read model for #5441 item 3)

## Problem

`#5441` item 3 (declarative `environments × providers` expansion) needs to reconcile the `clusters/` tree from the set of environments a workspace declares — but ksail has **no read-side model of that set**. Items 1–2 shipped the imperative, one-at-a-time primitives (`cluster add-environment`, `project init --multi-cluster`); nothing yet *enumerates* the environments already declared. Every declarative step (list, validate a `--from` against the real set, later reconcile/expand) first needs to answer "which environments does this workspace declare, and with what distribution/provider?"

## Design decision (item 3 source of truth)

Environments are **already declared** by the per-environment `ksail.<name>.yaml` root configs at the workspace root (the same convention `add-environment --from <env>` resolves). Rather than introduce a duplicative `spec.environments` list — a second source of truth that would drift from those files — item 3 **derives the declared set from the existing `ksail.<name>.yaml` roots**. This mirrors the "derive from what already exists, no new config field" pattern used for unmanaged-cluster discovery (#5654) and OIDC federation (#4602). No CRD/schema change.

## Proposed direction (this child — the read-side model)

Add `environment.DeriveEnvironments(repoRoot, load) ([]Environment, error)` to `pkg/svc/environment`: enumerate `ksail.<name>.yaml` roots (excluding the base `ksail.yaml` and any non-DNS-1123 name), load each via an injected `ConfigLoader` to read its distribution + provider, and return the set sorted by name. A config that fails to load is skipped (one broken file must not hide the rest). Consume it immediately in `add-environment` so a mistyped `--from` reports the **available environments** instead of only that the file was missing.

Model-first, mirroring #5654's model→CLI-parity sequencing.

## Follow-up children (not this PR)

- `project list-environments` CLI command + web-UI surface (consume the model).
- Declarative **reconcile**: compute the diff between declared configs and the `clusters/` overlays and generate/update the missing overlays + Flux Kustomization dependency wiring + per-cluster timeout knobs (the core of item 3).

## Rough size

S (this child — one pure discovery function + a friendly error, table-driven tests, no schema/CLI-surface change). Part of #5441.

devantler:

## Why

A real platform is multi-cluster/multi-environment, and #5441 tracks giving ksail a first-class way to scaffold and grow that shape. Its declarative step (item 3) must reconcile the `clusters/` tree from the environments a workspace declares — but nothing yet *enumerates* that declared set. This is the read-side foundation it needs.

## What

Adds a discovery model that lists the environments a workspace declares (via its `ksail.<name>.yaml` root configs — the existing convention, no new config field), reporting each one's distribution and provider. Uses it so a mistyped `cluster add-environment --from` now tells you which environments are actually available.

First, small increment of the #5441 item-3 epic; the `list-environments` command and the declarative reconcile follow as separate children.

Fixes #5862
Part of #5441

### DIFF
--- a/pkg/cli/cmd/project/add_environment.go
+++ b/pkg/cli/cmd/project/add_environment.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	v1alpha1 "github.com/devantler-tech/ksail/v7/pkg/apis/cluster/v1alpha1"
 	"github.com/devantler-tech/ksail/v7/pkg/cli/annotations"
@@ -164,7 +165,7 @@ func resolveAddEnvironmentParams(
 
 	srcCfg, err := loadSourceConfig(cmd, srcName)
 	if err != nil {
-		return addEnvironmentParams{}, err
+		return addEnvironmentParams{}, enrichSourceConfigError(cmd, repoRoot, err)
 	}
 
 	distribution := srcCfg.Spec.Cluster.Distribution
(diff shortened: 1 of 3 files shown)


devantler:

Addressed — added a `slog.Default().Debug` trace (`354c47e`) when a config is skipped, so a silently-absent environment entry now leaves a troubleshooting lead. The skip-on-load-error itself stays by design (one broken file must not hide the rest, per the documented contract + tests).

Issue #5851, opened 2026-07-05
Intercept steering increment 3: wire the tunnel over an exec stream + workload intercept CLI


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #5839 (traffic steering) / epic #4521 (local-remote service mirroring).

## Problem
Increments 1 (#5841: iptables REDIRECT builder + `ksail-steer` container + inject/wait seams) and 2 (#5847: the conn↔stream `Pump`, `ForwardRedirected`, `ServeIntercepted`, and the `TunnelSession` mux) built the **intercept library** — but nothing composes it. A `TunnelSession` needs a byte transport, and there is no CLI entry point. So the steering machinery is unreachable end-to-end.

## Proposed direction (decompose — ship smallest first)
1. **Exec-stream transport (this increment's first child — foundation).** A `pkg/svc/mirror` helper that opens a remotecommand exec session to a container and adapts its stdin/stdout into an `io.ReadWriteCloser` suitable as the transport for `NewTunnelSession`. Pure library, unit-testable with `io.Pipe` stubs behind an injectable executor seam — no live cluster. Unblocks both the agent side and the CLI.
2. **Steer-container agent command.** The `ksail-steer` container currently idles (`sleep infinity`); give it a command that installs the `SteeringRedirect` iptables rule and runs `ForwardRedirected` against the tunnel, with reversible teardown (`DeleteArgs`) on exit.
3. **`workload intercept` CLI command.** `NewInterceptCmd()` in `pkg/cli/cmd/workload/intercept.go` (registered under `groupDevLoop`, mirroring `mirror.go`): resolve → inject tap+steer → wait → open exec transport → `NewTunnelSession` → `ServeIntercepted` against the local dev process → block until Ctrl-C → cleanup. Flags: `--namespace/--container/--port/--local-port/--context/--steer-image/--wait-timeout`.

## Acceptance criteria (whole increment 3)
- `ksail workload intercept <deployment> --port <svc> --local-port <dev>` redirects the workload's inbound TCP into the developer's local process over the tunnel, and restores the pod on exit.
- Each sub-child is a focused, unit-testable PR; the exec-stream transport lands first with tests proving bytes flow both ways through a `TunnelSession` over a piped exec stub.
- Live-cluster end-to-end validation stays integration-only (like the mirror/capture path).

Starting with child 1 (exec-stream transport) now.

devantler:
> 🤖 Generated by the Daily AI Assistant

[...]

Issue #5839, opened 2026-07-05
Traffic steering for workload intercept: NET_ADMIN steering agent + iptables REDIRECT into the tunnel session


devantler:
> 🤖 Generated by the Daily AI Assistant

**Problem:** The Phase-2 intercept foundations are merged — the length-prefixed frame codec and the multiplexed `TunnelSession` (parity-split stream IDs, bounded per-stream receive budgets) — but nothing yet *steers* a remote Deployment's inbound traffic into that tunnel, so intercept mode cannot exist: traffic still terminates at the remote container, and the local process never sees a connection.

**Proposed direction** (informed by the merged P1/P2 code):
- **A separately-injected steering agent, never an upgrade of the tap.** The `ksail-tap` ephemeral container's NET_RAW-only hardened context is the security posture of mirror-only mode; steering runs as a second ephemeral container (`ksail-steer`) with `NET_ADMIN` added (drop-ALL otherwise, read-only rootfs, seccomp RuntimeDefault), so mirror and intercept stay independently runnable.
- **iptables NAT `REDIRECT`** in the pod netns: `-t nat -I PREROUTING -p tcp --dport <service-port> -j REDIRECT --to-port <intercept-port>` (same-netns redirect; original destination recoverable via `SO_ORIGINAL_DST`). TPROXY is rejected for now — it needs policy routing + more privileges for no benefit inside a single pod netns.
- **Conn→stream pump:** the agent accepts each redirected `net.Conn` and opens a `TunnelSession` stream per connection over the existing exec channel (stdin/stdout), piping bytes bidirectionally; the local ksail side accepts streams and dials the local process, so responses flow back through the same frames.
- **Reversibility is a hard requirement:** ephemeral containers cannot be removed, so the *rules* must be — the agent deletes its iptables rules on session close AND on an idle/heartbeat watchdog, so a dropped ksail session never leaves the service black-holed.

**Rough size:** M, shipped as unit-testable increments like the prior P2 children: (1) steering-rule builder + `ksail-steer` container spec + inject/lifecycle seams; (2) the conn↔stream pump against an in-memory TunnelSession; (3) `workload intercept` CLI wiring (live-cluster E2E deferred like the cloud lanes).

Part of #4521.

devantler:
> 🤖 Generated by the Daily AI Assistant

The full traffic-steering capability this issue scoped is merged. All three rough-size increments landed as focused, unit-tested PRs:

[...]

PR #5871, opened 2026-07-06
feat(mirror): add the in-pod steering agent entrypoint

devantler:

**Why:** `ksail workload intercept` (route a cluster deployment's live traffic to your local dev process) needs an agent inside the pod that redirects the inbound traffic and pumps it back over the tunnel. Previous increments built every piece except the one that composes them, so the intercept path still can't run end-to-end.

**What:** Adds that composition — the steering agent's in-pod entrypoint: install the traffic-redirect rule, forward the redirected connections over the tunnel, and undo the rule on exit (even when interrupted). It's the next-to-last building block before the `workload intercept` command itself. Pure library, fully unit-tested; nothing user-facing yet.

**Next:** one more child wires this behind the `ksail workload intercept` CLI (and picks how the agent is shipped into the container).

Part of #5851 / #5839 / #4521.


devantler:

**CodeRabbit outside-diff finding — `pkg/svc/mirror/agent.go` `removeSteeringRule` teardown observability: fixed.**

Valid point — a failed NAT REDIRECT delete on an ephemeral pod has no future cleanup path, so a silent swallow leaves a dangling rule invisible. Rather than deferring a logger dependency to the CLI wiring (`#5851`), `removeSteeringRule` now **returns** its build/exec error and `RunSteerAgent` **propagates** it via a named return — but only when the forwarding result was otherwise nil, so a real forwarding error is never clobbered. Added `TestRunSteerAgent_SurfacesTeardownFailure` (`-D` fails → the returned error surfaces). build + vet + `-race` + lint-v2.12.2 all green.

devantler:

Resolved the teardown-error-dropping finding (and its companion test suggestion) in `a670a67f`:

- `RunSteerAgent` now joins a teardown failure onto any forwarding error via `errors.Join`, so a delete-rule failure that coincides with a forwarding failure no longer vanishes — the dangling REDIRECT rule stays observable in either case (previously it surfaced only when forwarding succeeded).
- Added `TestRunSteerAgent_JoinsForwardingAndTeardownFailures` (a `failingListener` drives the genuine forwarding-failure path) asserting both errors surface.

build + vet + `-race` (9 tests) + golangci-lint all green.

Issue #5908, opened 2026-07-07
ci: jscpd reds `main` on every push — track & decompose the residual clone clusters


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem
The mega-linter `COPYPASTE`/jscpd check has `threshold: 0.0%` (zero tolerance) and runs **whole-codebase on `push` (main) events** (incremental on PRs). So any clone anywhere reds the `CI - KSail` → `CI - Required Checks` aggregate on `main`, even though PRs still merge cleanly. `main` is currently red with **36 clones** (latest full run). This is long-standing tribal knowledge with only two narrow child issues filed — this umbrella makes it trackable and decomposable.

## Clone clusters (from the latest main run)
- **CSS self-duplication** — `docs/src/styles/custom.css` (8 clones, internal repetition in one file).
- **Parallel cloud provider/connector files** — `pkg/svc/provider/{aws,gcp,azure}/provider.go`, `pkg/svc/provisioner/cluster/{gke,aks}/connector.go`, `pkg/svc/provider/{omni,kubernetes}` — near-identical by architecture.
- **Talos generators** — `pkg/fsutil/generator/talos/csrapprover/manifest.go`, `generator.go`, `configmanager/{kind,k3d}/helpers.go`.
- **Docker registry / mirror** — `pkg/client/docker/registry_*.go`, `pkg/cli/setup/mirrorregistry/talos.go`.
- Already-tracked pairs: exectransport↔session (#5878), clusterapi↔clusterdiscovery (#5876); hetzner constructors (#5748, PR in flight).

## Proposed directions (maintainer to steer)
1. **Decompose & dedup** the clusters above into small `refactor:` children (share-don't-clone), oldest-first — extends the existing #5878/#5876 approach.
2. **Scope the check** where duplication is architectural and not worth collapsing (e.g. per-provider connectors that are intentionally parallel) via jscpd ignore patterns, keeping 0.0% meaningful for genuinely-avoidable clones.
3. Combination of the two.

Rough size: umbrella epic; each cluster is an S–M child.

## Acceptance
- `main` full-codebase jscpd run is green (either 0 clones or a justified, documented ignore-scope).
- Each residual cluster is either deduped or explicitly scoped with a rationale.

devantler:
> 🤖 Generated by the Daily AI Assistant

[...]

Issue #5878, opened 2026-07-06
refactor(mirror): dedup the exec-subresource URL builder shared by exectransport.go and session.go


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

`pkg/svc/mirror/exectransport.go` and `pkg/svc/mirror/session.go` each build a pods/exec subresource `*url.URL` with the same `RESTClient().Post().Resource("pods").Name(...).Namespace(...).SubResource("exec").VersionedParams(...)` chain — jscpd flags them as a copy-paste clone (9 lines / 82 tokens). It sits just under the whole-repo copy-paste threshold today, so it is not currently red, but any further duplication in the mirror package tips the jscpd gate over (it did while a sibling feature was in flight).

## Proposed direction

Extract a small shared helper in the `mirror` package (e.g. `execRequestURL(client, namespace, pod, container, command)`) and call it from both sites. Behaviour-preserving; keep both callers exact.

## Rough size

XS — one helper + two call sites, no behaviour change.

PR #5967, opened 2026-07-08
fix(ci): clear jscpd + govulncheck findings reddening main's required checks

devantler:

## Why
`main`'s `CI - Required Checks` has been red for ~24h on two independent, whole-repo scans (jscpd 0.0% duplication threshold, govulncheck), which cascades to every open ksail PR since both checks scan the full tree regardless of what a PR touches.

## What
- **jscpd:** dedupes 3 of the outstanding clone clusters (CSS section dividers, cipher.go's encrypt/decrypt setup, startstop.go's shared long-description text) — behavior-preserving, tests unchanged. The exectransport/session cluster is separately tracked as #5878 (fix in #5952); larger cross-package provider/generator clusters remain as follow-on children of #5908.
- **govulncheck:** bumps the Go toolchain to 1.26.5 (fixes 2 new reachable stdlib advisories) and risk-accepts one unpatchable transitive advisory (`x/crypto/openpgp`, pulled in via Helm's chart-provenance code through the embedded kubescape library — no ksail call path, no upstream fix) in `.govulncheck-allow.txt`, consistent with the file's existing policy.

Part of #5908.


devantler:

@coderabbitai Confirmed and fixed (6263af70) — you were right that the prior comment's "ksail never calls Helm's provenance/PGP verification APIs itself" framing was wrong. A fresh `govulncheck -show verbose` run shows real symbol-level reachability: `pkg/client/kubescape.Client.ScanDirectory` → `kubescape/v3/core.Kubescape.Scan` → x/crypto/openpgp's armor/packet/clearsign decoders, via kubescape's own internal policy-scanning engine — independent of Helm's chart-provenance download path the old comment cited.

The allowlist entry itself is still correct to keep (GO-2026-5932 has `Fixed in: N/A` — it's a permanent "package is unmaintained, migrate off it" advisory with no patch to bump to, and ksail doesn't control kubescape's internal dependency choice), and `ScanDirectory` only ever processes a local, non-attacker-controlled manifest directory. I corrected the comment to document the real reachability chain accurately instead of removing the entry, which would just re-break the required govulncheck gate on an unpatchable finding.

devantler:

@coderabbitai Good catch, fixed (0870e204) — renamed `getClusterJSON` to `runGetJSON` (no longer cluster-specific) and reused it in `ListNodegroups` too, matching your suggested diff. Also applied the same consolidation to `aws.Provider.fetchNodegroups` (was still doing its own nil-check + translate instead of `provider.FetchOrTranslate`). Local `jscpd -c .jscpd.json .` confirms 0 clones.

devantler:

Fixed: added `imagePullPolicy: Always` to the cert-approver container in the manifest (and updated the package doc comment to note this deliberate local addition, since the file is otherwise a content-identical copy of upstream's standalone-install.yaml).

PR #5968, opened 2026-07-08
feat(workload): extract CRD schemas from rendered Helm/Kustomize output

### LINKED ISSUE #5906: feat(workload): extract CRD schemas from rendered (Helm/Kustomize) output for --include-crd-schemas

## Problem

`workload validate --include-crd-schemas` (introduced in #5905, `Fixes #5904`) derives kubeconform schemas from `kind: CustomResourceDefinition` manifests found by walking the **raw source tree**. CRDs that are only produced by **Helm/Kustomize rendering** (e.g. a chart that ships its CRDs in `templates/`, or a kustomize component that emits a CRD) are therefore **not** converted, so custom resources of those CRDs are still silently skipped by validation.

This is a deliberately-scoped boundary of the first increment (raw-tree files only, default-off flag, graceful degradation — a CRD it can't convert is warned + skipped, the run never fails). CodeRabbit flagged it on #5905 as a 🟡 Minor functional-correctness note; capturing it here as the planned follow-up.

## Proposed direction

Run CRD schema materialization on the **rendered** manifest set that the validate pipeline already produces (the `buildValidateRenderer` output that feeds `validatePath`), so CRD discovery uses the rendered output as the source of truth rather than the raw tree. Thread the rendered content into `crdschema.Materialize` (or a helper that calls it) while keeping the existing raw-tree behaviour and the default-off `--include-crd-schemas` flag.

## Rough size

M — reuses the existing `pkg/svc/crdschema` primitive and the renderer already in the validate path; the work is wiring CRD discovery to consume rendered output + tests (rendered-CRD catches an invalid CR / passes a valid one, both flag states). Feature-flag-first: extends the existing default-off flag, no new surface.

Part of #5344 · follow-up to #5904 / #5905

devantler:

## Why

`workload validate --include-crd-schemas` only finds CustomResourceDefinitions by walking the raw source tree, so a CRD a Helm chart ships under `templates/` — never copied into the GitOps repo itself — was invisible, and custom resources of that kind were silently skipped instead of validated.

## What

Extends CRD-schema discovery to also scan every kustomization's **rendered** output (reusing the same renderer the real validation pass already uses, so the double render is cache-assisted, not doubled cost) alongside the existing raw-tree walk. Both sources feed the same schema location, so a CRD from either place is validated. Off by default, opt-in via the existing `--include-crd-schemas` flag — no new surface.

Fixes #5906
Part of #5344


devantler:

Valid finding — `addRenderedCRDSchemas` and the main validation pass each render every kustomization independently when `--include-crd-schemas` is set, so `gitopsRenderer.expand` runs twice per kustomization (`ChartCache` only dedupes repeat Helm templating, not the kustomize/Flux expansion itself).

Deferring the fix out of this PR: threading the rendered stream through both passes is a real plumbing change to `validate.go` + the renderer interface, not a local tweak, and this PR is already a focused feature addition. Filed as devantler-tech/ksail#5969 to track the follow-up.

PR #5998, opened 2026-07-10
fix(eks): report control-plane endpoint in EKS provider status

devantler:

**Why:** `ksail cluster info` shows the cluster API endpoint, but EKS clusters silently showed nothing there while GKE and Azure clusters did — the EKS provider never populated the endpoint field. It's a small parity gap in a shipped command.

**What:** The EKS provider now reads the control-plane endpoint from the AWS SDK (eksctl's summary omits it) via a narrow, credential-free describer seam that resolves lazily — so the eksctl-only create/delete/scale paths still never touch AWS credentials. Behaviour is unchanged for every other provider. Unit-tested end-to-end through the `cluster info` path with a fake describer; no live AWS credentials needed.

Part of #4328 (completes the endpoint half of "cluster info reports EKS status").

## Summary by CodeRabbit

* **New Features**
  * Cluster status now includes the AWS EKS control-plane endpoint when available.
  * Endpoint details are retrieved for the selected cluster and region.

* **Bug Fixes**
  * Improved handling when endpoint information is unavailable.
  * Clearer error reporting when cluster endpoint details cannot be retrieved.
  * Region selection is consistently applied when checking AWS provider status.

### DIFF
--- a/pkg/cli/cmd/cluster/export_test.go
+++ b/pkg/cli/cmd/cluster/export_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/devantler-tech/ksail/v7/pkg/svc/clusterdiscovery"
 	clusterdetector "github.com/devantler-tech/ksail/v7/pkg/svc/detector/cluster"
 	"github.com/devantler-tech/ksail/v7/pkg/svc/provider"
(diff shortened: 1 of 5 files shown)


devantler:

@coderabbitai review

Issue #6040, opened 2026-07-11
[intercept] steering agent and its REDIRECT rule are orphaned when the client dies without a graceful close — pod traffic stays black-holed


devantler:
> 🤖 Generated by the Daily AI Engineer

Part of #4521; found by #5971's live-cluster validation pass (Kind/Docker, 2026-07-11). Sibling of #6038/#6039; complements the graceful-Ctrl-C fix in PR #6031.

## Problem
When the intercept client process dies without tearing the exec channel down gracefully (plain SIGINT on current `main` — no signal handler until #6031 — or kill -9, network drop, laptop sleep), the in-cluster steering agent keeps running and its REDIRECT rule stays installed: observed live, the agent process and the rule were still present 35+ seconds after client death, with the workload's traffic still redirected to a dead tunnel (connection refused for all callers). The agent's stdin read never observes EOF because the kubelet/containerd exec stream is not reliably closed when the client connection drops. The agent's own teardown IS sound — SIGTERM to the agent removed the rule and restored traffic immediately — but nothing triggers it on client death.

## Proposed direction
Give the tunnel a liveness mechanism so the agent can self-terminate when the client is gone: e.g. a client→agent keepalive ping frame every ~10s with an agent-side read deadline (~30s) that triggers the existing teardown, or an equivalent exec-channel liveness check. Ephemeral containers cannot be removed, so self-termination + rule removal is the only cleanup path.

## Rough size
M — protocol keepalive frame + agent read-deadline + tests both sides (client ships in lockstep with the version-pinned steer image, so the frame addition is skew-safe for release builds).

## Acceptance criteria
- [ ] Killing the client uncleanly (SIGKILL) results in the agent removing its REDIRECT rule and exiting within the liveness window (observed E2E: workload traffic restored without manual intervention).
- [ ] Normal sessions are unaffected (keepalive invisible to traffic; clean Ctrl-C path per #6031 still tears down immediately).

Issue #6039, opened 2026-07-11
[intercept] steering agent listens on 127.0.0.1 but iptables REDIRECT delivers remote traffic to the pod IP — intercepted traffic can never reach the agent


devantler:
> 🤖 Generated by the Daily AI Engineer

Part of #4521; found by #5971's live-cluster validation pass (Kind/Docker, 2026-07-11), reachable once #6038's corruption is worked around.

## Problem
The steering agent binds its interception listener to `127.0.0.1:<intercept-port>` (`loopbackHost` in `pkg/cli/cmd/steeragent`), reasoning "the REDIRECT rule retargets traffic here". That is not what REDIRECT does: for traffic arriving from outside the pod, `-j REDIRECT` rewrites the destination to **the address of the interface the packet came in on** (the pod eth0 IP), not to 127.0.0.1. Result on a live cluster: the rule is installed (`-A PREROUTING -p tcp --dport 8080 -j REDIRECT --to-ports 19000`), the agent listens on `127.0.0.1:19000`, and every intercepted connection is **refused** — observed live: `ss -tln` shows `127.0.0.1:19000`, and in-cluster curls to the pod IP and Service both fail with connection refused while the intercept session is up. The workload's inbound traffic is black-holed instead of forwarded — the feature's core path cannot work for any remote caller (the only callers that matter).

## Proposed direction
Bind the agent's interception listener to `0.0.0.0:<intercept-port>` (or the pod IP + loopback), and re-establish least-exposure with a guard the agent already has the capability to install (CAP_NET_ADMIN): e.g. `-I INPUT -p tcp --dport <intercept-port> -m conntrack ! --ctstate DNAT -j DROP` so only REDIRECTed connections (ctstate DNAT) reach the listener, never direct hits on `<podIP>:<intercept-port>` from other pods. Remove the guard in the same teardown that removes the REDIRECT rule. Update the `loopbackHost` reasoning comment.

## Rough size
S–M — agent listen-address change + guard rule install/teardown + tests; E2E-verified via a locally built steer image (`Dockerfile.steer` + `kind load`).

## Acceptance criteria
- [ ] With an intercept session running against a real cluster, an in-cluster curl to the workload's Service/pod IP is answered by the developer's local process (observed E2E).
- [ ] A direct connection to `<podIP>:<intercept-port>` from another pod is refused/dropped while the session runs.
- [ ] Teardown removes both the REDIRECT rule and the guard rule (pod traffic restored, observed E2E).

Issue #5971, opened 2026-07-09
[intercept] Graduate `workload intercept` out of the experimental gate after a live-cluster validation pass


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #4521 (local-remote service mirroring). Follow-up to #5882.

## Problem
`ksail workload intercept` is fully wired end-to-end (client CLI #5883, exec-stream transport #5852, hidden `steer-agent` entrypoint #5901, default steering image #5944, default-image + derived-command activation #5946) and unit-tested throughout, but it still ships behind `experimental.Guard` (`--experimental` required). Per `AGENTS.md`'s feature-flag lifecycle rule, a validated feature should graduate to stable rather than carry the gate forever — but per the "Verify it WORKS E2E, not by guts" rule, graduation itself needs a real observed pass, not just green unit tests.

## Proposed direction
On a throwaway cluster (Kind/K3d), run `ksail workload intercept` end-to-end against a real Deployment: confirm inbound traffic is actually redirected to a local process via the tunnel, and that the pod is restored (iptables rule removed) on exit/Ctrl-C. Once that live pass is clean, graduate the command by deleting the single `experimental.Guard(cmd)` call in `pkg/cli/cmd/workload/intercept.go`, regenerate the `--help`/toolgen/docs snapshots (`UPDATE_SNAPS=true go test ./pkg/...` + `make generate`), and update the `AGENTS.md` "Feature-flag / experimental gating" reference note (currently says gated "until the default steering-agent image (#5882) ships" — that part is now done; the remaining gate condition is this live validation).

## Rough size
S — mostly a live-cluster verification pass (always clean up the cluster) plus a small mechanical PR to drop the guard and regenerate snapshots once that pass is clean.

## Acceptance criteria
- [ ] A live-cluster run of `ksail workload intercept` is observed to actually redirect traffic and restore the pod on exit (documented in the PR, per the E2E-verify-don't-guess rule).
- [ ] `experimental.Guard` removed from `NewInterceptCmd`; `--help`/toolgen/docs snapshots regenerated.
- [ ] `AGENTS.md`'s experimental-gating reference note updated to reflect the graduation.

devantler:
> 🤖 Generated by the Daily AI Engineer

[...]

PR #6041, opened 2026-07-11
fix(intercept): fail loud on a corrupted steering tunnel and silence the agent banner

### LINKED ISSUE #6038: [intercept] steer-agent stdout noise corrupts the tunnel protocol, and the client masks the corruption as a clean exit 0

Part of #4521; found by #5971's live-cluster validation pass (Kind/Docker, 2026-07-11).

## Problem
`ksail workload intercept` exits 0 within ~2s on a real cluster without ever running the steering agent — a silent no-op. Two stacked defects:

1. **Protocol corruption:** the ksail binary links `github.com/derailed/k9s/cmd`, whose package `init()` does `fmt.Printf("Fail to init k9s logs location …")` — **to stdout** — whenever the k9s state dir is not creatable (k9s v0.51.0 `cmd/root.go:57`). The steer container's filesystem is read-only, so every in-container ksail process prints this line at startup. `ksail steer-agent`'s stdout **is** the tunnel channel, so the client demux reads `Fail ` as a frame header and tears the session down with `tunnel frame has an unknown type: 32` (byte 5 = ASCII space).
2. **Failure masking:** that protocol error is recorded in `TunnelSession.loopErr` but never surfaced: `ServeIntercepted` returns `nil` on any `AcceptStream` failure, `defaultRunInterceptSession` discards `transport.Close()`'s terminal stream error, and `ExecTransport.run` normalises the teardown-triggered `context.Canceled` to nil — so the exec channel is cancelled before it ever dials, the agent never starts, and the CLI exits 0.

Observed evidence: client exits 0 in <3s, no REDIRECT rule installed, no steering agent process; `kubectl exec <pod> -c ksail-steer -- sh -c 'ksail --version 2>/dev/null'` prints the k9s line (stdout-only capture); instrumented client shows the demux error above; `XDG_STATE_HOME=/dev/shm` (writable tmpfs) silences the noise and the session then stays alive with the REDIRECT rule installed.

## Proposed direction
- Make the intercept path fail loudly: surface `TunnelSession.Err()` through `ServeIntercepted` (or check it after), and stop discarding the transport's terminal stream error — a corrupted tunnel must be a non-zero, diagnosable failure, never exit 0.
- Silence the known noise source for the shipped image: derive the default steer invocation with a writable `XDG_STATE_HOME` (e.g. `env XDG_STATE_HOME=/dev/shm`) and/or bake the k9s state dir into `Dockerfile.steer` so `InitLogLoc` succeeds.
- (Upstream, separate decision: k9s printing to stdout from `init()` is an upstream defect — a candidate for an upstream PR moving it to stderr; ask-gated per contract.)

## Rough size
S–M — client error-surfacing + derived-command/Dockerfile hardening + unit tests pinning both.

## Acceptance criteria
- [ ] A corrupted tunnel (non-frame bytes on the channel) makes `workload intercept` exit non-zero with the protocol error in the message.
- [ ] With the default steer image, no non-frame bytes reach the tunnel (live pass shows the session staying up and the REDIRECT rule installed).
- [ ] Unit tests pin both the error surfacing and the noise-free derived command.

devantler:

## Why
A live-cluster validation pass for `workload intercept` (the graduation gate, #5971) found the command exiting **0 while doing nothing** — a silent no-op. The steering agent's own startup banner was corrupting the tunnel it runs over, and the client swallowed the failure as success. That is the worst kind of bug: it looks like it worked. This is the first of the three real defects that pass surfaced.

## What
Two changes so a broken tunnel can never again read as success:
- The intercept client now **fails loudly** — a tunnel torn down by a protocol/codec error becomes a diagnosable non-zero exit instead of a silent nil.
- The shipped steering image no longer emits the stray banner onto the tunnel channel, so the healthy path stays healthy.

Verified end-to-end against a real cluster (the session now stays up and installs its redirect rule) and with unit tests pinning both the fail-loud path and the noise-free image.

Two follow-on defects the same pass found are filed separately (#6039 listener/redirect mismatch, #6040 orphaned-agent leak); graduation (#5971) stays gated until those land.

Fixes #6038
Part of #4521

### DIFF
--- a/Dockerfile.steer
+++ b/Dockerfile.steer
@@ -26,4 +26,14 @@ ARG TARGETPLATFORM
 # command exec'd over the tunnel is simply `ksail steer-agent …`.
 COPY ${TARGETPLATFORM}/ksail /usr/local/bin/ksail
 
+# Point the XDG state dir at a writable tmpfs. ksail links k9s's cmd package,
+# whose init() prints "Fail to init k9s logs location …" to STDOUT when it cannot
+# create its state dir (the default ~/.local/state is unwritable under this
+# read-only-rootfs steer container). For `ksail steer-agent` STDOUT *is* the
+# tunnel channel, so that banner would inject non-frame bytes and corrupt the
+# steering tunnel — the intercept client rejects it as an unknown frame type.
+# /dev/shm is a writable tmpfs present in every pod, so InitLogLoc succeeds and
+# emits nothing.
+ENV XDG_STATE_HOME=/dev/shm
+
 ENTRYPOINT ["/usr/local/bin/ksail"]
--- a/pkg/svc/mirror/pump.go
+++ b/pkg/svc/mirror/pump.go
@@ -152,10 +152,11 @@ func pumpUntilDone(ctx context.Context, left, right io.ReadWriteCloser) {
 // stream the steering agent opened, dials the developer's local process, and
 // pumps the two together. A failed dial closes the stream — the tunnel's
 // "connection refused", which makes the agent close the redirected cluster
-// connection. It blocks until the context is cancelled or the session closes
-// (both return nil — inspect [TunnelSession.Err] for why a session ended);
-// cancellation also tears down every active pump, and all pumps are drained
-// before it returns.
+// connection. It blocks until the context is cancelled or the session ends: a
+// clean end (context cancelled or the caller closed the session) returns nil,
+// while a session the demux loop tore down on a protocol/codec error returns
+// that error. Cancellation also tears down every active pump, and all pumps are
+// drained before it returns.
 func ServeIntercepted(ctx context.Context, session *TunnelSession, dial LocalDialer) error {
 	pumps := &sync.WaitGroup{}
 	defer pumps.Wait()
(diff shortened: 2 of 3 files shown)


devantler:

@coderabbitai run pre-merge checks

Docstring-Coverage fix pushed (8b0097b): the new test's rationale was an in-body comment; it is now a proper leading doc comment, and `sessionTeardownErr` already carried one — every function in the diff is now documented.

devantler:

@coderabbitai run pre-merge checks

Docstring coverage re-fix: the check measures the whole touched file (as on #6034), so every pre-existing helper and test in `pump_test.go` now carries a doc comment (commit pushed); `pump.go`'s functions were already fully documented.

devantler:

@coderabbitai Docstring Coverage still reports 0.00% (05:02Z re-run), but every function declaration in the two touched Go files now carries a leading doc comment — `pump.go` was already fully documented and commit `cc8f08cf` documented all 16 pre-existing helpers/tests in `pump_test.go`. Please list the exact declarations the check currently counts as undocumented (file + function name) so the remaining gap can be fixed without guesswork.

devantler:

@coderabbitai run pre-merge checks

Fresh retry after the review-limit window reopened. The current summary still reports Docstring Coverage 0.00%, but the latest CodeRabbit analysis independently confirmed every declaration in `pump.go` and `pump_test.go` has a leading doc comment and identified the result as a likely stale/check-side anomaly.

Issue #6070, opened 2026-07-12
fix(update): floating IP attached and endpoint synced but VIP never becomes reachable (dead API endpoint in cluster secrets)


devantler:
> 🤖 Generated by the Daily AI Engineer

## Problem

The #6034 floating-IP reconcile path ran live for the first time today on the platform prod cluster (v7.166.1, merge_group deploys 08:32Z and 10:20Z 2026-07-12). It created/attached the floating IP and synced the endpoint into cluster secrets — but the floating IP never became reachable, so **the endpoint now stored in cluster secrets is dead**. Any subsequent kubeconfig export (or DR rebuild consuming the synced endpoint) targets an IP nothing answers on.

## Live evidence (prod, 2026-07-12 ~10:45Z)

- 08:32:03Z `✓ Synced cluster secrets and endpoint (178.105.204.76) from 49.13.53.183` — sync ran BEFORE the attach and picked up a different, pre-existing floating IP (itself unreachable; possibly a leaked earlier creation).
- 08:32:04Z `✓ Floating IP 5.75.215.250 attached to prod-control-plane-2`, then `Regenerating configs with endpoint IP 5.75.215.250...`, all nodes config-applied in place (no reboot), run green.
- 10:20Z run (next queue entry): `✓ Synced cluster secrets and endpoint (5.75.215.250) from 49.13.53.183`.
- `curl -k https://5.75.215.250:6443/version` → connection timeout (2+ hours after attach). Same for 178.105.204.76. Direct CP `49.13.53.183:6443` answers 401 as expected.
- CP-2's serving cert DOES carry `IP Address:5.75.215.250` in its SANs — so the SAN side took effect, but the address is never bound/claimed on the node's interface (Hetzner floating IPs require the server to configure the IP; a Hetzner-side attach alone does not bind it).

## Proposed direction

1. **Root-cause the binding gap:** verify what the in-place config push actually contains for the VIP (Talos `vip` on the interface needs the hcloud API token for the claim controller; a missing/invalid token or absent interface config leaves the IP attached-but-unbound). Reproduce in the fake-hcloud E2E by asserting the *reachability contract*, not just the attach call.
2. **Fail-closed endpoint sync:** never sync a floating IP into cluster secrets before it is verified reachable (e.g. TCP/TLS probe on `<ip>:6443`) — today the sync happily recorded a dead endpoint, which is worse than the pre-#6034 no-op.
[...]

PR #6071, opened 2026-07-12
fix(update): never persist an unreachable floating-IP endpoint

devantler:

## Why

The first live floating-IP reconcile on the prod platform cluster recorded a dead API endpoint into the synced cluster secrets and the kubeconfig — the IP was attached in the cloud but never became reachable, so kubeconfig exports and new-node joins would target an address nothing answers on (#6070).

## What

Cluster updates now verify an endpoint actually answers before persisting it: an unreachable floating IP is no longer written into synced secrets or the kubeconfig — the directly reachable control-plane address is used instead, with a visible warning. Once released and picked up by the platform, the next deploy self-heals the dead endpoint currently recorded on prod.

Part of #6070 (the unclaimed-VIP root cause and duplicate floating-IP surfacing remain on the issue).

### DIFF
--- a/pkg/svc/provisioner/cluster/talos/endpoint_probe.go
+++ b/pkg/svc/provisioner/cluster/talos/endpoint_probe.go
@@ -1,0 +1,62 @@
+package talosprovisioner
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+const (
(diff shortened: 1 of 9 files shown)


devantler:

@codex review

Issue #6078, opened 2026-07-12
Honor custom AWS credential environment mappings for EKS lifecycle


devantler:
## Problem

`spec.provider.aws` exposes `profileEnvVar`, `accessKeyIdEnvVar`, `secretAccessKeyEnvVar`, and `sessionTokenEnvVar` so users can point KSail at non-standard environment variable names. The EKS lifecycle currently shells out to `eksctl` without translating those configured variables to the canonical AWS variables inherited by the child process.

As a result, a configuration such as `profileEnvVar: KSAIL_AWS_PROFILE` can create or inspect a cluster with the wrong default AWS identity even though the KSail configuration explicitly selected another profile. The same drift can affect static credentials. This was confirmed while implementing #6077; it is separate from post-create context selection and should not widen that fix.

## Proposed direction

- Resolve each configured AWS credential environment variable without mutating the parent process environment.
- Pass an explicit child-process environment to every `eksctl` lifecycle invocation, mapping configured values to `AWS_PROFILE`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`.
- Keep region precedence aligned with `lifecycle.ResolveAWSRegion` and avoid placing credentials in command arguments, logs, errors, or test output.
- Cover default names, custom names, missing optional values, and isolation between concurrent invocations.

## Acceptance criteria

- EKS create/read/delete/update operations honor all configured AWS credential environment-variable names.
- The child environment contains only the resolved canonical mappings needed by `eksctl`; the parent environment is unchanged.
- Tests prove a custom profile and custom temporary credentials reach the `eksctl` runner without secret disclosure.
- Existing canonical AWS environment variables continue to work unchanged.

## Size

Medium. This likely needs an environment-aware runner option shared across the EKS client commands plus focused lifecycle tests.

Part of #4328.

> 🤖 Generated by the Daily AI Engineer (Codex automation).

PR #6079, opened 2026-07-12
fix(eks): resolve post-create kubeconfig context

### LINKED ISSUE #6077: Resolve the actual eksctl context before GitOps bootstrap

## Description

`ksail cluster create` synthesizes an incomplete EKS kubeconfig context before the post-create Helm/GitOps install. KSail currently assigns `<name>.eksctl.io`, while eksctl writes an identity- and region-qualified context such as `<identity>@<name>.<region>.eksctl.io`. `HelmClientForCluster` then exact-validates the synthetic context and can fail before Flux or ArgoCD bootstrapping begins.

## Steps to Reproduce

1. Configure AWS/EKS with an empty `spec.cluster.connection.context` and a writable kubeconfig path.
2. Let eksctl create the cluster and write its normal qualified context.
3. Continue into KSail's post-create component installation.
4. Observe KSail validate `<name>.eksctl.io`, which is absent from the kubeconfig.

## Expected Behavior

KSail resolves the actual context that eksctl wrote and passes it to post-create Flux/ArgoCD installation. An explicitly configured context is never overwritten.

## Actual Behavior

`runClusterCreationWorkflow` assigns the synthetic distribution context before `handlePostCreationSetup`; EKS does not implement `KubeconfigRefresher`, so the exact Helm client lookup can target a context that does not exist.

## Acceptance Criteria

- [ ] Preserve a non-empty user-configured context unchanged.
- [ ] For EKS with an empty context, read the configured kubeconfig after creation and match the exact `@<name>.<region>.eksctl.io` suffix.
- [ ] Prefer a matching `current-context`; otherwise accept exactly one matching context.
- [ ] Fail closed with an explicit-context hint when no match exists or multiple matches remain ambiguous.
- [ ] Pass the resolved context into both Flux and ArgoCD post-create paths.
- [ ] Preserve existing context resolution for every non-EKS distribution.
- [ ] Add failing-first unit tests for the qualified context, explicit context, unique/current selection, no-match, ambiguity, and non-EKS behavior.

## Additional Context

Parent: #4328 (AC4). This internal correctness slice is independent of the still-required live AWS/OIDC smoke run. Keep the README/support-matrix EKS markers unchanged until that live verification succeeds.

devantler:
> 🤖 Generated by the Daily AI Engineer (Codex automation).

## Summary

- pass KSail's canonical configured kubeconfig path to `eksctl create cluster`
- resolve eksctl's identity-qualified EKS context before Flux or Argo CD setup
- prefer the matching current context, accept a unique match, and fail closed on zero or ambiguous matches
- preserve explicit EKS contexts and retain the existing deterministic context behavior for every non-EKS distribution
- resolve the effective AWS region consistently and support profile-delegated regions
- create and canonicalize the kubeconfig output directory before invoking the external process

## Why

KSail previously synthesized `<name>.eksctl.io`, but eksctl writes contexts such as `<identity>@<name>.<region>.eksctl.io`. Cluster creation could succeed and then GitOps bootstrap would fail because Helm targeted a context that did not exist. A custom kubeconfig path could also drift because KSail read one path while eksctl wrote another.

## Validation

- failing-first Flux and Argo CD regression reproduced the synthetic-context failure
- `go test ./pkg/client/eksctl ./pkg/svc/provisioner/cluster/eks ./pkg/svc/provisioner/cluster ./pkg/cli/lifecycle ./pkg/cli/cmd/cluster -count=1`
- `golangci-lint run --fix --new-from-rev=HEAD` across affected package trees: 0 issues
- `go build -o /tmp/ksail-maint .`
- repository-wide `go test ./...`: affected packages passed; two unrelated environment/process-sensitive packages failed under the fully parallel host run and both passed when rerun serially with the live `HCLOUD_TOKEN` removed

## Scope

Custom AWS credential environment-variable mappings are a separate pre-existing concern and are tracked in #6078 rather than widening this fix.

Fixes #6077
Part of #4328


devantler:

Sweep note: CI is green and this draft is CLEAN, but no review has been requested at any head yet — securing the green-review gate so it's promotable.

@codex review

PR #6080, opened 2026-07-12
fix(cluster): resolve the actual eksctl context before GitOps bootstrap

### LINKED ISSUE #6077: Resolve the actual eksctl context before GitOps bootstrap

devantler:

## Why

EKS cluster creation could fail right before Flux/ArgoCD bootstrapping: KSail assumed a synthetic kubeconfig context name while eksctl writes an identity- and region-qualified one, so the post-create install targeted a context that doesn't exist.

## What

After EKS creation, KSail now reads the context eksctl actually wrote from the kubeconfig (preferring current-context, accepting a unique match, and failing closed with a clear hint when none or several match). Explicitly configured contexts and all non-EKS distributions behave exactly as before.

Fixes #6077. Part of #4328 (AC4).

### DIFF
--- a/pkg/cli/cmd/cluster/export_test.go
+++ b/pkg/cli/cmd/cluster/export_test.go
@@ -81,6 +81,19 @@ func ExportResolveClusterContext(kubeconfigPath, clusterName string) (string, er
 	return resolveClusterContext(kubeconfigPath, clusterName)
 }
 
+// ExportResolveCreatedContext exports resolveCreatedContext for testing.
+func ExportResolveCreatedContext(
+	ctx *localregistry.Context,
+	clusterName string,
+) (string, error) {
+	return resolveCreatedContext(ctx, clusterName)
+}
+
+// ExportResolveEKSCreatedContext exports resolveEKSCreatedContext for testing.
+func ExportResolveEKSCreatedContext(kubeconfigSpec, name, region string) (string, error) {
+	return resolveEKSCreatedContext(kubeconfigSpec, name, region)
+}
+
 // ExportEnsureConfiguredContextResolvable exports ensureConfiguredContextResolvable for testing.
 func ExportEnsureConfiguredContextResolvable(clusterCfg *v1alpha1.Cluster) error {
 	return ensureConfiguredContextResolvable(clusterCfg)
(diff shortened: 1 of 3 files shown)


devantler:

Closing as a duplicate: #6079 (opened minutes earlier by the sibling instance) resolves the same #6077 eksctl-context bug with the same fail-closed semantics and additionally fixes the kubeconfig-path passing and region/profile resolution around it. Deferring to #6079 to avoid a duplicate lane on one concern.

PR #6115, opened 2026-07-14
refactor(cluster): converge unmanaged synthesis onto a shared clusterdiscovery seam

### LINKED ISSUE #5876: refactor(cluster): converge clusterapi unmanaged synthesis onto the shared clusterdiscovery seam

## Problem

#5869 added `clusterdiscovery.DiscoverUnmanaged` (kubeconfig-context enumeration + managed-set dedup) so `ksail cluster list` surfaces unmanaged clusters. But `pkg/cli/clusterapi` still carries its own near-duplicate of that logic — `unmanagedClusters` / `contextIsManaged` in `resources.go` — which predates the shared seam and produces `v1alpha1.Cluster` objects instead of `clusterdiscovery.Cluster`. Two copies of the enumerate-and-dedup rule can drift (e.g. a fix to context detection lands in one but not the other).

## Proposed direction

Extract the shared primitive — "given a kubeconfig config + a managed-name predicate, return the unmanaged context names (sorted, deduped)" — into `clusterdiscovery`, and have BOTH `DiscoverUnmanaged` (CLI) and `clusterapi`'s `unmanagedClusters` consume it. The two surfaces keep their own result shapes (`clusterdiscovery.Cluster` vs a `v1alpha1.Cluster` with the `ksail.io/unmanaged` annotation + Ready=False condition), but the enumeration/dedup logic lives in one place. Behaviour-preserving; both surfaces keep their existing tests green.

## Rough size

S — one extracted function + two call sites rewired + no behaviour change.

Part of #5654. Follows #5869.

devantler:

## Why

The CLI's `cluster list` and the web-UI/desktop model each carried their own copy of the same "list unmanaged kubeconfig clusters" rule. Two copies can silently drift — a fix to one surface (e.g. context detection) can miss the other, so the two views disagree about which clusters are unmanaged.

## What

Extracts the shared enumerate-and-dedup step into one place so both surfaces stay aligned by construction. Behaviour is unchanged (existing tests pass; the shared step is unit-tested).

Fixes #5876

### DIFF
--- a/pkg/cli/clusterapi/resources.go
+++ b/pkg/cli/clusterapi/resources.go
@@ -3,7 +3,6 @@ package clusterapi
 import (
 	"context"
 	"fmt"
-	"sort"
 
 	v1alpha1 "github.com/devantler-tech/ksail/v7/pkg/apis/cluster/v1alpha1"
 	"github.com/devantler-tech/ksail/v7/pkg/k8s"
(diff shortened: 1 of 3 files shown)


devantler:

@codex review

PR #6138, opened 2026-07-14
fix(release): give the release build the disk headroom it needs

### LINKED ISSUE #6137: fix(release): releases are silently dropped when the runner runs out of disk

## Problem

**Releases are being silently dropped because the release runner runs out of disk.**

`v7.169.3` never shipped. Its CD run
([29336552552](https://github.com/devantler-tech/ksail/actions/runs/29336552552)) died 32 minutes into
`🏗️ Run GoReleaser`:

```
⨯ release failed after 32m45s
  build failed: exit status 1
  github.com/go-openapi/errors: mkdir /tmp/go-build.../b4429/: no space left on device
  go.uber.org/dig/internal/digerror: mkdir /tmp/go-build.../b4648/: no space left on device
  ...
```

The job already frees a lot up front (`dotnet`, `android`, `ghc`, the CodeQL tool cache, all Docker
images and build cache) — and it *still* exhausts the disk. The build has simply outgrown the runner:
GoReleaser cross-compiles every OS/arch and builds and signs the container images, with Go's build
scratch in `/tmp` on the root filesystem.

The blast radius is bigger than one lost release:

- The tag is cut before the build, so a failed release **leaves an orphaned tag** (`v7.169.3` exists
  with no release behind it) — #6136.
- The evergreen cask branch is written with that version, so the tap is offered a cask whose download
  URL 404s (homebrew-tap#1184 fails `brew audit` for exactly this reason).
- Nobody is told. The failure looks like one red CD run among many.

This is also latent, not historical: the `v7.170.0` build is running the same path right now.

## Proposed direction

- Give the build the headroom it needs — reclaim more of the preinstalled toolchains, and keep Go's
  build scratch and cache off the small root filesystem.
- **Print free space before and after the cleanup.** Right now a disk failure surfaces as a confusing
  compiler/linker error deep in a 30-minute build; the run should say plainly how much room it had.
- Consider whether every artifact GoReleaser produces still needs to be built in one job — splitting
  the matrix would remove the ceiling rather than raise it.

## Rough size

S to raise the headroom; M if the release job is split.

devantler:

## Why

**A release can be silently dropped because the build machine runs out of disk.** That is what happened
to v7.169.3: it died 32 minutes into the build, so the version was never published — yet the tag was
already cut, and the Homebrew cask was already written against it. The tap is still holding a cask for
a release that does not exist.

The failure is also hard to recognise: running out of disk shows up as a compiler error deep in a long
build, so it reads like a broken build rather than a full disk.

## What

- Give the build the room it needs: reclaim more of the pre-installed software we never use, and keep
  the compiler's scratch space off the small system disk (it picks whichever disk actually has the most
  room, checked at run time rather than assumed).
- **Print the free space before and after**, so the next time this bites it says so plainly instead of
  looking like a build failure.

Worth knowing: this raises the ceiling rather than removing it. The single release job builds every
platform and both container images at once, so it will creep back toward the limit — splitting that job
is the durable fix and is noted on the issue.

Fixes #6137

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -71,13 +71,16 @@ jobs:
           df -h / /mnt 2>/dev/null || df -h /
           echo "::endgroup::"
 
+          # Do NOT add /usr/local/lib/node_modules here: that directory *is* npm, and GoReleaser's
+          # `before` hook runs scripts/stage-webui.sh, which builds the web UI with `npm ci`. This job
+          # has no setup-node step, so removing it would break the release at the hook — the very
+          # failure this change exists to prevent.
           sudo rm -rf \
             /usr/share/dotnet \
             /usr/local/lib/android \
             /opt/ghc \
             /usr/local/.ghcup \
             /opt/hostedtoolcache/CodeQL \
-            /usr/local/lib/node_modules \
             /usr/share/swift \
             /usr/local/share/boost \
             /usr/local/share/powershell
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -59,11 +59,62 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
+      # GoReleaser cross-compiles every OS/arch and builds+signs the images in one job. That outgrew the
+      # runner's root filesystem: v7.169.3 died 32 minutes in with `mkdir /tmp/go-build…: no space left
+      # on device`, which surfaces as a compiler error and silently dropped the release (#6137).
+      # So: reclaim more of the preinstalled toolchains, keep Go's scratch off the root filesystem, and
+      # PRINT the headroom — a disk failure must never again look like a build failure.
       - name: 🧹 Cleanup node
         run: |
-          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL
+          set -euo pipefail
+          echo "::group::Disk before cleanup"
+          df -h / /mnt 2>/dev/null || df -h /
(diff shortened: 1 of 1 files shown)


devantler:

@codex review

devantler:

**The finding is correct, and it was a serious bug in this PR — thank you.** Verified at source before acting:

- This PR is what added `/usr/local/lib/node_modules` to the reclaim list (it was not in the pre-existing cleanup).
- That directory **is** the runner's npm installation.
- `.goreleaser.yaml` runs `bash scripts/stage-webui.sh` as a `before` hook, and that script calls `npm --prefix web/ui ci` / `run build`.
- The `goreleaser` job has **no** `setup-node` step (only the separate VS Code extension job does).

So the change would have broken the release at the GoReleaser hook — the exact silent-release-drop failure it exists to prevent. A disk fix that deletes a toolchain the build needs is worse than no fix.

**Fixed in `bb80757`** by removing that one path from the list and leaving a comment stating the constraint so it is not re-added. I deliberately did **not** take the "delete it and add `setup-node`" route: the reclaim there is a few hundred MB, `setup-node` would put most of it straight back in the tool cache, and it adds a network dependency to the release path — while the multi-GB wins (dotnet, android, ghc, CodeQL, swift, boost, powershell) are untouched and still land. `actionlint` and YAML parse green.

**One correction for the record:** this comment states the change was committed as `7c0a59b` and a follow-up PR was created. Neither exists — there is no such commit in the repository and no such PR. The fix above is my own, authored at the root cause. Flagging it because a reported-but-unlanded remediation is easy to take on trust.

devantler:

@codex review

PR #6172, opened 2026-07-16
fix(config): reserve a Hetzner server slot for the Talos snapshot build

### LINKED ISSUE #6171: fix(config): validation permits maxNodesTotal == serverLimit, leaving no slot for the Talos snapshot build

## Problem

KSail's capacity validation accepts a Hetzner config whose reachable node total **equals** `provider.hetzner.serverLimit`. On the Talos + Hetzner path that configuration is not actually valid: KSail itself needs to create a **temporary server** to build the Talos snapshot, and at the cap there is no slot for it.

`pkg/apis/cluster/v1alpha1/validation.go:466` only rejects the strict overflow:

```go
if reachableTotal > serverLimit {   // `==` passes
```

So a config sitting exactly at the cap validates clean, deploys fine for weeks, and then fails the first time a snapshot must be built.

## Evidence

The deploy path ensures the autoscaler config secret on **every** deploy, and that step builds the Talos snapshot whenever one is **absent** — independent of what changed. So the failure is latent and version-triggered:

- While the pinned Talos version's snapshot exists → the lookup hits → no temp server → everything green. **This is why it hides.**
- **Any Talos version bump** invalidates the lookup → every deploy must build → needs a slot → fails:

```
Building Talos snapshot (version: v1.13.6, schematic: df884f4f…)
✗ failed to apply updates: failed to ensure autoscaler config secret:
  looking up snapshot image for autoscaler secret: failed to ensure Talos snapshot:
  talos snapshot build failed: creating the temporary server failed:
  server limit reached (resource_limit_exceeded)
```

Observed twice on a cluster configured at the cap, each time triggered by a routine Talos bump (v1.13.5, then v1.13.6) — so with a bot proposing Talos updates this is a **recurring trap, not a one-off**. Because the temp server fails at *creation*, nothing is orphaned; and normal autoscale decay frees a slot and makes the symptom vanish while the config stays broken — which is why it appears to "resolve itself" and then returns.

Worth stressing: the blast radius is not just a failed deploy. Any recovery/heal path that runs the same ensure step fails identically, so a GitOps consumer can be left with **no working deploy path at all** until a slot frees.

## The validator currently pins the wrong behaviour

`pkg/apis/cluster/v1alpha1/autoscaler_test.go:830` is named:

> `capacity guard: maxNodesTotal equal to serverLimit is valid despite large pools`

That test asserts exactly the configuration that deadlocks. Any fix must reframe it, not just add a case beside it.

## Expected

KSail should reserve headroom for the servers **it creates itself**. A config that can reach the account cap leaves KSail no room to do its own work, so validation should require the reachable total to stay **strictly below** `serverLimit` on paths where KSail may need a temporary server (Talos snapshot builds), rather than merely not exceed it.

Open design question worth deciding rather than assuming: whether the reserve should be unconditional, or applied only when the provisioning path can actually build a snapshot (Talos + Hetzner). A blanket `<` would also constrain users who never trigger a snapshot build — so scoping it to the paths that need it is probably right, but that's a call to make deliberately.

## Acceptance criteria

- [ ] A Hetzner + Talos config whose reachable total equals `serverLimit` is **rejected at validation time**, with an error naming the reserved slot and why (snapshot build), not discovered later as a mid-deploy `resource_limit_exceeded`.
- [ ] The reserve is scoped to the paths that can create a temporary server (decision recorded in the issue/PR).
- [ ] `autoscaler_test.go:830` is reframed to assert the new contract; table cases cover `reachable == limit` (rejected), `reachable == limit - 1` (valid), and an uncapped-pool case.
- [ ] The `serverLimit` JSON-schema description states the reserved slot, so the constraint is discoverable from config completion.

## Size

~S–M. The validation change is small; the scoping decision and the test reframe are the substance.

devantler:

## Why

KSail accepted a Talos + Hetzner cluster config that can grow to use **every server the account allows** — leaving no room for the temporary server **KSail itself boots** to build the Talos snapshot the autoscaler needs.

That config cannot work, but it fails late and confusingly: it looks fine for weeks (while the pinned Talos version's snapshot already exists), then the next Talos bump makes every deploy try to build one, and each dies mid-apply with an opaque `resource_limit_exceeded` — recovery deploys included. This wedged our own prod today, and it did the same thing on the previous Talos bump, so it's a recurring trap rather than a one-off.

## What

Validation now reserves one server slot on the Talos + Hetzner path, so this surfaces as a **config error that names the reserve** instead of a runtime failure with no obvious connection to node counts. Scoped to Talos on purpose — the k3s and kubeadm Hetzner paths build no snapshot and keep the full limit.

## ⚠️ Behaviour change — please weigh this before promoting

**Configs that deploy successfully today will start failing validation.** A Talos+Hetzner cluster sitting exactly at its server limit works fine for as long as its snapshot exists; after this, `ksail cluster create/update` refuses it up front and asks for one slot of headroom.

That's the intent — the config is a landmine that only detonates on the next Talos bump — but it does mean a working deploy can start erroring on a KSail upgrade, so the semver call is yours. **Our own `platform` prod config is affected** (it's exactly at the cap, which is what caused today's outage); its capacity needs fixing regardless.

Static `ksail workload validate` is **not** affected — I verified the cluster-spec validation doesn't run on that path — so consumers' manifest CI won't break.

Fixes #6171

### DIFF
--- a/pkg/apis/cluster/v1alpha1/autoscaler_test.go
+++ b/pkg/apis/cluster/v1alpha1/autoscaler_test.go
@@ -827,9 +827,14 @@ func TestValidateAutoscalerConfig(t *testing.T) {
 			// the baseline. With baseline 6 + poolCapacity 10, the old logic computed
 			// 6 + min(10,10) = 16 and wrongly rejected, even though MaxNodesTotal(10)
 			// equals serverLimit(10). reachableTotal = min(16, 10) = 10 → valid.
-			name: "capacity guard: maxNodesTotal equal to serverLimit is valid despite large pools",
+			//
+			// Distribution is deliberately NOT Talos: only the Talos Hetzner path builds
+			// a snapshot, so only it reserves a slot (#6171). Every other distribution is
+			// entitled to the full serverLimit, which is what this case pins.
+			name: "capacity guard: maxNodesTotal equal to serverLimit is valid for a non-Talos distribution",
 			cluster: &v1alpha1.ClusterSpec{
 				Provider:      v1alpha1.ProviderHetzner,
+				Distribution:  v1alpha1.DistributionK3s,
 				ControlPlanes: 3,
 				Workers:       3,
 				Autoscaler: v1alpha1.AutoscalerConfig{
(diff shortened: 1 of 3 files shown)


devantler:

@codex review

PR #6250, opened 2026-07-18
fix(security): drop allowlist entries for unreachable Docker archive advisories

devantler:

### Motivation
Three Docker `cp` / container-archive advisories sat on ksail's vulnerability allowlist as accepted risk. They are not reachable from ksail today, so the entries suppress nothing — but a standing risk acceptance would silently wave those advisories through the moment ksail did start using that code path. Removing them means a real finding there shows up as a finding.

### Description
Drops the `GO-2026-5617`, `GO-2026-5668` and `GO-2026-5746` entries and their explanatory block from the allowlist, and adjusts the header text so it does not carry a stale count.

### Testing
The vulnerability gate runs on this PR and passes. It prints every reachable advisory, and none of the three removed entries appears in that list — which is the evidence that they were suppressing nothing and that removing them does not open a gap.

> **Correction to this PR's original description.** It claimed these advisories were *reachable* and that the allowlist was hiding real findings. That was wrong: the gate's own reachability output does not list them, so the removed comment block ("unreachable in practice") was accurate. The change is still right, for the narrower reason above — dead risk-acceptance entries are worth removing — but it is a cleanup, not the recovery of suppressed findings.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52c3914604832b9b3c8882d35f7fb0)


devantler:

devantler:
@cursor review

devantler (commented):
> 🤖 Generated by the Agentic Engineer — static review of this ownership-unverified draft at `1c1e619f7990e498c3eae7fa959c12fd611aad3e`; the branch was left untouched.

## Review finding

### [P1] Make the allowlist deletion reach the vulnerability scanner

This change cannot surface the three advisories through the current KSail CI path. At this head, [`ci-go`](https://github.com/devantler-tech/ksail/blob/1c1e619f7990e498c3eae7fa959c12fd611aad3e/.github/workflows/ci.yaml#L270-L273) invokes the reusable validator only when the caller event is a `push`, but the pinned validator gates [`govulncheck`](https://github.com/devantler-tech/actions/blob/84813ae61fb21741e0a6193b5c14c6941a448577/.github/workflows/validate-go-project.yaml#L355-L366) on `github.event_name == pull_request`. On this PR the outer job is skipped; after merge the reusable job sees `push` and skips the scanner. That is why this head can report successful required checks without evaluating the deleted entries, and the stated security outcome remains absent. Please couple this deletion to a KSail workflow path that actually executes the pinned `govulncheck` job, then refresh the exact-head evidence before promotion.

devantler:

## Conflict resolved — and the central claim here has never actually been tested

Merged `main` at `97107c84`. Only the header block conflicted (main rewrote it on 2026-07-22); git kept the deletion automatically, so the branch now differs from `main` by **exactly** the 12-line docker `cp`/archive block — 12 allowlist entries down to 9. Nothing else.

### Why this PR was not ready to judge before now

The Testing section records that `rg` finds no matches for the three IDs. That proves the **lines were deleted**; it says nothing about the thing this PR changes — whether the govulncheck gate still passes without them.

It could not have said anything, because **the gate never ran on this PR**. At head `1c1e619f79` the only workflow runs were `CI - KSail` and two CodeQL runs — no Go validation run exists at that sha at all. This is *not* the prefix-filtered check-name illusion that produced #6563/#6564: I looked at the runs, not the check names.

That gap is closed on `main` now. `ci-go` gained a `pull_request` arm gated on the `govuln-allowlist` filter (`.govulncheck-allow.txt`), precisely so a risk acceptance is validated by the scanner it applies to — the org-required run takes `scan-default-branch: false` and skips an allowlist-only PR. So this rebase is what makes the change measurable for the first time.

### What the next CI run means — either result is a real answer

The block being removed states its own advisories are unfixed on the `docker/docker` v1 line (patched only in `moby/moby/v2`, which ksail does not import), and the gate fails on any **reachable** advisory not listed.

- **Gate red** → the entries were load-bearing, the premise ("an incorrect risk-acceptance entry") is wrong, and the right terminal state is to close this PR with that evidence rather than weaken the gate.
- **Gate green** → they were genuinely stale or no longer reachable, and this is a correct tightening worth merging.

I have deliberately not predicted which. The point of the rebase is that the scanner decides instead of the PR body asserting.

devantler:



@coderabbitai review

devantler:

**Readiness**

- **Programmatically tested** — the `✅ Validate Go Project / 🛡️ Vulnerability Scan` gate ran at this head and passed. That matters here because the changed file *is* that gate's input: at the previous head no Go validation run existed at all, so the gate had never scanned the allowlist it governs. The rebase onto current `main` is what made it measurable.
- **Reviewed** — `cr@97107c84`, finding-free, current head.
- **Tried and evaluated as a user** — read the gate run's own output rather than inferring from a green tick. It enumerates every reachable advisory: `GO-2025-3521`, `GO-2025-3547`, `GO-2026-4883`, `GO-2026-4887`, `GO-2026-5064`, `GO-2026-5338`, `GO-2026-5622`, `GO-2026-5932`, `GO-2026-5939`. **None of the three removed entries appears**, which is the direct evidence that they suppressed nothing and that removing them opens no gap.

That evidence also corrected this PR: it was opened claiming the three advisories were *reachable* and being hidden. They are not — the comment block it deleted was accurate. The change is still right, as a removal of dead risk acceptance that would otherwise silently pre-approve those advisories if the `docker cp` path ever became reachable. Title and body have been corrected to say so.

PR #6255, opened 2026-07-18
fix(webui): enforce same-origin policy for pod exec WebSocket

devantler:
### Motivation
- The pod exec WebSocket previously used an upgrader with `CheckOrigin` returning `true`, which allowed browser pages from arbitrary origins to open WebSocket connections to a local, unauthenticated UI and drive `kubectl`-style exec into pods using the user's kubeconfig.
- This created a cross-origin security boundary break enabling a malicious webpage to execute commands in cluster workloads when the local UI was running.

### Description
- Replace the permissive upgrader with the default gorilla/websocket policy by changing the upgrader in `pkg/webui/api/exec_handler.go` from `websocket.Upgrader{CheckOrigin: func(_ *http.Request) bool { return true }}` to `websocket.Upgrader{}` so browser-initiated connections must be same-origin while non-browser/no-origin clients are still allowed.
- Add regression coverage in `pkg/webui/api/exec_test.go` that verifies a same-origin `Origin` header succeeds and that a cross-origin `Origin` (e.g. `https://attacker.example`) is rejected with `403 Forbidden`.
- The change preserves behavior for non-browser clients and for same-origin local UI/desktop flows while closing the cross-origin attack vector.

### Testing
- Ran `gofmt` to apply formatting changes on the modified files and verified the working tree diff for correctness; no formatting issues remain.
- Attempted unit tests: ran `go test ./pkg/webui/api -run 'TestExec(WebSocketBridge|RejectsCrossOriginWebSocket)' -count=1`, but the run was blocked by environment module fetch restrictions (failed to download `mvdan.cc/sh/v3` from the proxy), so the tests could not be executed to completion in this environment.
- Also attempted `GOPROXY=direct go test ...` which was similarly blocked by network/registry fetch restrictions; test code and assertions are present and intended to pass when dependencies are available in an unrestricted environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a531022448c832bae3d1b1d2afabc57)

### DIFF
--- a/pkg/webui/api/exec_handler.go
+++ b/pkg/webui/api/exec_handler.go
@@ -92,9 +92,11 @@ func (s *Server) handleExec(writer http.ResponseWriter, request *http.Request) {
 		command = []string{"/bin/sh"}
 	}
 
-	// CheckOrigin allows any origin: the endpoint is already behind the auth guard (when OIDC is on)
-	// and the read-only check above, and the desktop connects cross-origin (wails:// → loopback).
-	upgrader := websocket.Upgrader{CheckOrigin: func(_ *http.Request) bool { return true }}
+	// Use gorilla/websocket's default origin policy: requests without an Origin header are allowed,
+	// and browser requests must be same-origin (Origin host equals Host). The local UI is
+	// unauthenticated and bound to loopback, so allowing arbitrary browser origins would let a
+	// malicious page drive pod exec through a victim's local kubeconfig.
+	upgrader := websocket.Upgrader{}
 
 	conn, err := upgrader.Upgrade(writer, request, nil)
 	if err != nil {
--- a/pkg/webui/api/exec_handler.go
+++ b/pkg/webui/api/exec_handler.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/url"
+	"strings"
 	"sync"
 	"time"
 
@@ -92,11 +95,7 @@ func (s *Server) handleExec(writer http.ResponseWriter, request *http.Request) {
 		command = []string{"/bin/sh"}
 	}
 
-	// Use gorilla/websocket's default origin policy: requests without an Origin header are allowed,
-	// and browser requests must be same-origin (Origin host equals Host). The local UI is
-	// unauthenticated and bound to loopback, so allowing arbitrary browser origins would let a
-	// malicious page drive pod exec through a victim's local kubeconfig.
-	upgrader := websocket.Upgrader{}
+	upgrader := websocket.Upgrader{CheckOrigin: execCheckOrigin}
 
 	conn, err := upgrader.Upgrade(writer, request, nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


devantler:

devantler:
@cursor review

pkg/webui/api/exec_handler.go
```diff
@@ -92,9 +92,11 @@ func (s *Server) handleExec(writer http.ResponseWriter, request *http.Request) {
...
 	}
 
-	// CheckOrigin allows any origin: the endpoint is already behind the auth guard (when OIDC is on)
-	// and the read-only check above, and the desktop connects cross-origin (wails:// → loopback).
-	upgrader := websocket.Upgrader{CheckOrigin: func(_ *http.Request) bool { return true }}
+	// Use gorilla/websocket's default origin policy: requests without an Origin header are allowed,
+	// and browser requests must be same-origin (Origin host equals Host). The local UI is
+	// unauthenticated and bound to loopback, so allowing arbitrary browser origins would let a
+	// malicious page drive pod exec through a victim's local kubeconfig.
+	upgrader := websocket.Upgrader{}
```
devantler: > 🤖 Generated by the Agentic Engineer

[P1] Anchor the origin check to the loopback listener, not the request Host

Reviewed head: `dddaa654199073de9dc5bcdcfe3d5d2b67377b6a`. Gorilla's default policy only requires the `Origin` host to equal the request's `Host`. Both values can be `attacker.example` during DNS rebinding even though that hostname has rebound to `127.0.0.1`, so this upgrade succeeds and the unauthenticated local handler can still exec through the victim's kubeconfig. The new test covers a mismatched hostile Origin, but not the matched hostile Host/Origin pair. Please validate Host/Origin against the configured loopback UI origin (independent of request headers) and add a regression with `Host: attacker.example:<port>` plus `Origin: http://attacker.example:<port>` that must return 403.

devantler: > 🤖 Generated by the Agentic Engineer

Confirmed and fixed at `9a39c9248c350d4dd36bd5edfbb9e8df588737f4`.

The finding is correct: gorilla's default `CheckOrigin` only requires the Origin host to equal the request Host, and DNS rebinding puts **both** under the attacker's control, so a self-consistent hostile pair passed. Reproduced it as a failing test before fixing — with the previous head's `websocket.Upgrader{}`, a dial carrying `Host: attacker.example` + `Origin: http://attacker.example` was **accepted**.

The check is now anchored on the loopback identity rather than on request headers (`execCheckOrigin`): an absent Origin is still allowed (non-browser clients; browsers always send it on an upgrade), the fixed `wails://` desktop origin is trusted, and every other origin must resolve to a loopback host. A browser sets Origin from the page's real URL, so only a page actually served from loopback can present one.

Scope check: the exec route is registered only when the service implements `ExecService`, which is `pkg/cli/clusterapi` alone — the operator never registers it — so this costs the non-loopback backend nothing.

Two regression tests added: `TestExecRejectsRebindingOriginMatchingHost` (the case you asked for; fails against the previous head) and `TestExecAllowsLocalhostOrigin`. The second surfaced a **DevEx regression the previous head had introduced**: the default policy rejects `Origin: http://localhost:PORT` against a `127.0.0.1`-bound listener, so anyone reaching the UI by the `localhost` spelling would have lost the terminal. That now works again. `golangci-lint` output is identical to the pre-change baseline.

devantler:



CI is green at this head and the same-origin fix has been reworked to anchor on the loopback identity
rather than the request Host. Requesting a review.

@coderabbitai review

devantler:

## Readiness at `9a39c9248c350d4dd36bd5edfbb9e8df588737f4`

**Programmatically tested.** All checks green at this head (0 failing). `go test ./pkg/webui/api/ -run Exec`
passes 6/6 locally.

**Reviewed.** CodeRabbit at this head — `Review completed`, and its summary comment (updated 18:12:00Z,
after the 18:09:37Z request, naming this head) reports *"No actionable comments were generated in the
recent review."* Unresolved threads: 0. Non-thread findings: 0.

**Tried and evaluated as a user — with an ablation, because a passing security test proves nothing until
you have seen it fail.** I ran the exec suite on this head, then reverted just the one line that installs
the policy (`CheckOrigin: execCheckOrigin` → `func(_ *http.Request) bool { return true }`) and re-ran:

| Run | Result |
|---|---|
| At head | 6/6 pass |
| Policy ablated | **exactly 2 failures** — `TestExecRejectsCrossOriginWebSocket`, `TestExecRejectsRebindingOriginMatchingHost` |
| Restored | 6/6 pass |

Predicted two failures and got exactly two, so the tests pin the behaviour rather than passing
incidentally, and the DNS-rebinding case is genuinely closed: a page on a name rebound to `127.0.0.1`
presents a matching `Origin`/`Host` pair, which the previous `CheckOrigin` accepted and this one rejects.

I also checked the origin parser against the shapes that usually defeat this kind of check —
`Origin: null` (sandboxed iframe), a suffix spoof (`http://127.0.0.1.attacker.com`), a prefix spoof
(`http://localhost.attacker.com`), and IPv6 loopback (`http://[::1]:8080`). The first three are rejected
because `url.Parse` yields a hostname that neither equals `localhost` nor parses as a loopback IP; the
last is accepted via `Hostname()` stripping the brackets. The everyday path is unchanged — the SPA served
from the loopback listener, and the desktop shell's `wails://` origin, both still connect.

Promoting on that basis.

PR #6258, opened 2026-07-18
fix(ui): disable local kubeconfig export

devantler:
### Motivation
- Prevent credential disclosure via the unauthenticated local UI: the local `clusterapi.Service` implementation exported full kubeconfig bytes (including `AuthInfo`) through a GET endpoint that was reachable when the UI ran bound to loopback.

### Description
- Remove the local kubeconfig export implementation by deleting `pkg/cli/clusterapi/kubeconfig.go` so the local service no longer implements `api.KubeconfigProvider` and the `/api/v1/clusters/{ns}/{name}/kubeconfig` route is not registered for the unauthenticated local backend.
- Replace the previous unit tests with a regression test `TestLocalServiceDoesNotExposeKubeconfigProvider` in `pkg/cli/clusterapi/kubeconfig_test.go` that asserts the local service does not satisfy `api.KubeconfigProvider`.
- Clarify capability documentation comments in `pkg/webui/api/service.go` and `web/ui/src/api.ts` to state that credential-bearing kubeconfig exports must only be implemented by backends with an authentication boundary.

### Testing
- Ran `git diff --check` and staged/committed the change successfully; the commit message is `fix(ui): disable local kubeconfig export`.
- Attempted `go test` for the affected packages: `go test -v ./pkg/cli/clusterapi -run TestLocalServiceDoesNotExposeKubeconfigProvider -count=1` and `go test ./pkg/webui/api -run TestKubeconfigEndpointUnregisteredWithoutProvider -count=1`, but both test runs timed out in this environment and did not produce full results.
- No other automated test failures were observed from the quick repository checks performed here (file diffs and `git` checks completed successfully).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a531acd18c0832b8e0a657538ae17e7)


devantler:

devantler:
@cursor review

devantler:

## Parked on a named blocker: this PR's Go test has never executed

I picked this up to drive to merge. The surface reads ready — 21 checks green, 13 skipped, 0 review
threads, 0 non-thread findings, and a Cursor Bugbot green at head `9ed47ac6a4`. I did not promote it,
because the readiness that matters is not present.

**The regression test added here has never run — anywhere.**

* The PR description records that the author's own runs "timed out in this environment and did not
  produce full results".
* CI did not cover for that: `✅ Validate Go Project` is **SKIPPED** on this PR.

That skip is not specific to this PR. `ci-go` gates its pull-request arm on
`needs.changes.outputs.govuln-allowlist`, a filter matching only `.govulncheck-allow.txt`, rather than
on `code` (`**/*.go`, `go.mod`, `go.sum`). So Go validation is skipped on *every* pull request that
changes Go code — confirmed on merged #6550 (3 Go files) and #6528 (2 Go files), both SKIPPED. Filed
as #6563.

So `TestLocalServiceDoesNotExposeKubeconfigProvider` is unproven, and the deletion of
`pkg/cli/clusterapi/kubeconfig.go` has not been compiled against the current tree by any test job.
This branch is also **327 commits behind `main`**, so even its stale green describes a tree that no
longer exists.

## The change itself looks right

Removing the credential-bearing export from the unauthenticated local backend, and stating in
`service.go` / `api.ts` that such exports belong only to a backend with an authentication boundary, is
a sound boundary to draw. This is not an objection to the design — it is that nothing has verified it.

## What unblocks it

1. #6563 lands, so Go validation actually runs on pull requests.
2. This branch is brought up to date with `main`.
3. The suite runs green at the new head, and a fresh review is secured there (every push stales the
   current Bugbot green).

Leaving it open and parked rather than closing it — the work is worth keeping, it just cannot be
called tested yet.

devantler:

Updated this branch onto current `main`, because its checks were **not measuring the change**.

Diagnosed by run rather than by check name: at the previous head there was **no Go validation run at all** — only `CI - KSail`, the Web UI/Desktop workflows and CodeQL. `✅ Validate Go Project` showed as `skipping`, which is not the check-name-prefix illusion; the workflow genuinely never ran. So the Go regression guard this PR adds had **never executed**, while an all-green rollup plus a current-head green review made the PR read as ready to promote.

The heads here are ~26 days old and predate the `ci-go` `pull_request` arm, so the rebase is what makes the change measurable — the same fix that unblocked ksail#6250 earlier today.

This does stale the previous green review, which has to be re-secured at the new head. That is the right trade: the old green sat over a guard nothing had run.

devantler:



@coderabbitai review

devantler:

**Readiness record** at head `751bf14f148e6718af14ed70dbedddd9e588b58d`.

**1. Programmatically tested.** This branch previously sat at a head that predated the `ci-go` `pull_request` arm, so **no Go validation ran on it at all** — the guard it adds had never executed while the PR still read as all-green. After `update-branch`, diagnosis *by run* (not by check name) shows `✅ Validate Go Project` present at this head with its `🧪 Test` job `completed/success` — not `skipped` — and the log carries `ok github.com/devantler-tech/ksail/v7/pkg/cli/clusterapi 1.040s`, so the package holding the new `TestLocalServiceDoesNotExposeKubeconfigProvider` compiled and ran.

**2. Reviewed.** CodeRabbit green at this head: `No actionable comments were generated in the recent review`, naming `981c7993…751bf14f`, corroborated by the head status `Review completed`. Zero unresolved threads, zero non-thread findings.

**3. Tried and evaluated as a user.** The committed test asserts the *mechanism* (the type assertion fails). I additionally exercised the **real** local backend through the **real** server — `api.Server{Service: clusterapi.NewService()}` — and observed what a local web UI user actually receives, then ablated the same check against this PR's merge-base `981c7993`:

| Observation | Base `981c7993` | Head `751bf14f` |
|---|---|---|
| `GET /api/v1/config` → `capabilities.kubeconfigDownload` | `true` | `false` |
| `GET /api/v1/clusters/default/prod/kubeconfig` | **200** (served kubeconfig bytes) | **404** (route not registered) |

The ablation fired on both conjuncts, so the change is causally responsible for both. It also confirms the exposure was real rather than theoretical: before this change the unauthenticated, loopback-bound local API **did** serve credential-bearing kubeconfig bytes on request. The route is now absent altogether, not merely hidden in the SPA — so the fix does not depend on the frontend honouring the capability flag.

(The scratch acceptance check was run in a throwaway worktree and is not part of this diff.)

Promoting and merging on this evidence.

PR #6260, opened 2026-07-18
fix(webui): require JSON content type for API writes

devantler:
### Motivation
- Prevent cross-origin simple requests (e.g. `Content-Type: text/plain` + `mode: no-cors`) from being decoded and causing mutating side effects such as cloud cluster provisioning via the local UI API.
- Harden the web UI REST surface now that cloud-capable distributions/providers are exposed through the local create endpoint by enforcing an explicit media-type contract for JSON writes.

### Description
- Added `requireJSONContentType` and `errUnsupportedJSONContentType` and enforce them from `decodeJSON` and `decodeCluster`, returning HTTP 415 for non-`application/json` requests. (file: `pkg/webui/api/server.go`)
- Updated `writeDecodeError` to map unsupported media types to `415 Unsupported Media Type`. (file: `pkg/webui/api/server.go`)
- Added regression test `TestCreateClusterRejectsSimpleCrossOriginContentType` that simulates a cross-origin-style `text/plain` POST to `/api/v1/clusters` and asserts the request is rejected and no cluster is created. (file: `pkg/webui/api/server_test.go`)
- Adjusted the API test helper to set `Content-Type: application/json` for tests that supply a body so existing tests continue to exercise normal JSON request handling. (file: `pkg/webui/api/server_test.go`)

### Testing
- Ran the new unit test: `go test ./pkg/webui/api -run TestCreateClusterRejectsSimpleCrossOriginContentType -count=1` and the package tests: `go test ./pkg/webui/api -count=1`, both succeeded.
- Built the repo binary: `go build -o /tmp/ksail-maint .` which succeeded locally.
- Attempted `go test ./...`; several tests that fetch remote Kubernetes JSON schemas failed in this environment with HTTP 403 from `raw.githubusercontent.com`, and one editor-related snapshot test differed due to environment-specific editor lookup; these failures are environment/network related and not caused by the change to JSON content-type handling.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5326a1ee3c832b917cbb753507183d)


devantler (commented):

## Review finding

**[P1] Guard non-JSON POST mutations too**

Reviewed head: `69d4805c4a23a37322ecc1399ad4a4487118cd05`

The JSON media-type check closes the demonstrated `POST /api/v1/clusters` path, but it does not close the cross-origin mutation class described by this PR. `POST /api/v1/clusters/{namespace}/{name}/apply` accepts an arbitrary manifest body, while the restart and reconcile routes are bodyless POSTs. All three remain browser-simple requests and reach mutating services when OIDC is disabled, because the handler chain has no Origin/CSRF guard. An arbitrary page can therefore still apply manifests or trigger resource actions against a reachable local UI without a preflight.

Please enforce the boundary for every state-changing route (for example, a shared Origin/token middleware rather than a JSON-decoder-only check) and add negative tests proving cross-origin-style simple requests cannot invoke `ApplyManifests`, `RestartResource`, or `ReconcileResource`.

devantler:



First review request on this PR. It has been open since early August with all checks green, no unresolved threads, and no conflict with base — no review had ever been requested, so the review gate rather than the change itself was holding it.

@coderabbitai review

devantler:

## This PR owns the content-type concern — #6256 duplicates it partially

**#6256** (*constrain SOPS decrypt to age metadata*) also adds an `application/json` requirement to
`decodeJSON`, so the two PRs collide on `pkg/webui/api/server.go` and `server_test.go`. Both branch
from `main`, so whichever merges second conflicts.

**This PR is the one that should keep the concern.** `main` has five body-decoding call sites —
three via `decodeJSON`, two via `decodeCluster` (`handleCreateCluster` and the update handler). This
PR guards all five by calling `requireJSONContentType` from both decoders; #6256 guards only the
three `decodeJSON` sites and leaves the cluster write endpoints accepting any content type. Routing
the 415 through `writeDecodeError` with the `errUnsupportedJSONContentType` sentinel also keeps the
status mapping in one place rather than writing it inline at the call site.

I have recommended on #6256 that it drop the content-type change and stay scoped to its SOPS title.

**What this PR still needs:** it is `DIRTY` against `main` and has had no review requested at
`69d4805c4a`. The conflict wants resolving before a review is spent, since any resolution moves the
head and stales the result.

devantler:



@coderabbitai review

devantler:

Readiness record for `dcd17eba5dfefdb5aaaf5869ed005a6c06ae5867`.

**Programmatically tested** — `go test ./pkg/webui/api -count=1` passes; all required checks green
(0 non-green), 0 unresolved threads, 0 non-thread review findings, CLEAN against base.

**Reviewed** — CodeRabbit green at this head: summary comment updated `08:04:25Z` (after the
`08:01:20Z` request), naming the range ending `dcd17eba5dfefdb5aaaf5869ed005a6c06ae5867`, reporting
no actionable comments. No rate-limit or service marker.

**Tried and evaluated as a user** — ran the built binary (`ksail open web --port 18261`) and drove
the real endpoint:

| Request to `POST /api/v1/clusters` | Result |
|---|---|
| `Content-Type: text/plain` (the cross-origin simple-request shape) | **415** `content type must be application/json` |
| `Content-Type: application/x-www-form-urlencoded` | **415** — the other simple content type is closed too |
| `Content-Type: application/json` | passes the gate, reaches validation (422 on my deliberately invalid body) |
| `Content-Type: application/json; charset=utf-8` | **accepted** — the media type is parsed, not string-compared |

Side effect asserted, not just the status code: `GET /api/v1/clusters` after the rejected writes
lists only the pre-existing unmanaged kubeconfig entry, so no cluster was provisioned. `GET`
still returns 200, so the gate does not touch the read path.

The charset case is the one worth calling out — a naive `== "application/json"` comparison would
have rejected it and broken ordinary browser clients while still passing every other check here.

Promoting and merging.

Issue #6271, opened 2026-07-19
chat CLI-probe tests fail under full-suite load: fixed 3s/5s process timeouts


devantler:
> 🤖 Generated by the Daily AI Engineer

## Evidence
On a loaded machine, `go test ./...` fails `pkg/cli/cmd/open/chat`'s process-spawning tests while the same package passes in isolation, at the same commit:

- Full suite (2026-07-19, this host): `TestDiagnoseCLIStartupFailure/captures_stderr_from_failing_process` and `/captures_multiline_stderr` fail at 3.00s; `TestVerifyCopilotCLI/wraps_a_pre-flight_failure_with_the_CLI_output` and `/returns_nil_when_the_CLI_exits_zero` fail at 5.00s.
- Same worktree, `go test ./pkg/cli/cmd/open/chat/` alone: 127 passed.
- Pristine `origin/main` worktree, same package alone: 127 passed.

The exact 3.00s/5.00s failure durations match fixed timeouts: the tests spawn real processes and give them a hard deadline, which a runner under full-suite parallel load can miss without any product defect. An earlier note (2026-07-18) misread this as a deterministic host-environment failure; the isolation/main controls above show it is load-dependent timing.

## Audience and impact
Anyone running the repo's documented pre-PR validation (`go test ./...`) on a busy machine gets false reds in an unrelated package — this run it cost a false-red diagnosis on an EKS-installer PR, and it degrades trust in local validation (the failure class where people start ignoring red).

## Expected behaviour
The chat CLI-probe tests pass under full-suite load: either the probe deadlines scale (generous ceiling with early-exit on success), the tests serialize their process spawns, or the deadline is enforced on observed progress rather than wall-clock.

## Acceptance criteria
- [ ] `go test ./...` on a loaded machine (e.g. `-parallel` default, stress alongside) passes the four named subtests reliably (10/10 repetitions).
- [ ] Isolated-package behaviour and covered failure modes are unchanged (the tests still fail when the probed process genuinely hangs or errors).
- [ ] No blanket timeout inflation that would slow the isolated happy path materially.

Rough size: S.

devantler:
> 🤖 Generated by the Daily AI Engineer

**Refining the hypothesis on this issue — "flaky under full-suite load" looks wrong.**

Observed again today while validating an unrelated branch:

- `go test ./...` → `TestDiagnoseCLIStartupFailure` FAILS (`expected "Error: missing config", actual ""`).
[...]

PR #6276, opened 2026-07-19
fix(hetzner): select the VIP interface by bus path so the floating IP binds

### LINKED ISSUE #6070: fix(update): floating IP attached and endpoint synced but VIP never becomes reachable (dead API endpoint in cluster secrets)

devantler:

## Why

The Hetzner floating IP that is meant to give the prod Kubernetes API a stable address has never actually worked. KSail attached the IP and recorded it as the cluster endpoint, but nothing on the node ever claimed it — so the recorded endpoint answered nothing, and any kubeconfig exported from it pointed at a dead address.

The cause is a wrong assumption about how Talos names network interfaces. The config targeted an interface that does not exist on these servers, so the instruction was silently ignored.

## What

Targets the public network card by its hardware slot instead of by name, which is what the servers actually expose. Confirmed against both live production control planes before and after.

This unblocks devantler-tech/platform#2120 (stable API endpoint), which has been waiting on this. The IP will bind on the next deploy after a KSail release is pinned; no manual step is needed.

Fixes #6070

### DIFF
--- a/pkg/fsutil/configmanager/talos/configs.go
+++ b/pkg/fsutil/configmanager/talos/configs.go
@@ -354,27 +354,39 @@ func (c *Configs) WithHetznerVIP(vip, hcloudAPIToken string) (*Configs, error) {
 	})
 }
 
+// hetznerPublicNICBusPath is the PCI bus path of the public NIC on a Hetzner Cloud
+// server. The public interface is always the first virtio device (slot 01); a cluster
+// private network attaches as a later device (slot 07 on the two-NIC servers KSail
+// provisions), so the bus path distinguishes the two where `physical: true` cannot.
+const hetznerPublicNICBusPath = "0000:01:00.0"
+
 // buildHetznerVIPPatch builds a control-plane-scope patch that configures the Talos
 // virtual (shared) IP on the public interface, with hcloud API management so the
 // elected leader reassigns the floating IP to itself.
 //
-// The public interface is addressed by name (eth0) rather than a deviceSelector:
-// KSail always attaches its servers to a cluster private network, so the VMs carry
-// two virtio NICs and a `physical: true` selector would match both — while Hetzner
-// Cloud VMs predictably enumerate the public NIC first under Talos' ethN naming
-// (design decision on devantler-tech/ksail#5718). dhcp stays enabled so declaring
-// the interface does not drop its primary address configuration.
+// The public interface is addressed by a deviceSelector on its PCI bus path rather
+// than by link name. Addressing it as `eth0` (the original design on
+// devantler-tech/ksail#5718) assumed Talos uses ethN naming — it does not: Talos
+// applies predictable interface names, so a Hetzner control plane enumerates its NICs
+// as enp1s0 (public) and enp7s0 (private) and carries no `eth0` at all. The patch
+// therefore matched no link, the VIP was never claimed on the node, and the floating
+// IP stayed attached-but-unbound with a dead API endpoint (devantler-tech/ksail#6070).
+// A `physical: true` selector cannot be used instead: KSail always attaches its servers
+// to a cluster private network, so it would match both NICs. dhcp stays enabled so
+// declaring the interface does not drop its primary address configuration.
 func buildHetznerVIPPatch(vip, hcloudAPIToken string) Patch {
 	content := fmt.Sprintf(
 		"machine:\n"+
 			"  network:\n"+
 			"    interfaces:\n"+
-			"      - interface: eth0\n"+
+			"      - deviceSelector:\n"+
+			"          busPath: %q\n"+
 			"        dhcp: true\n"+
 			"        vip:\n"+
 			"          ip: %q\n"+
 			"          hcloud:\n"+
 			"            apiToken: %q\n",
+		hetznerPublicNICBusPath,
 		vip,
 		hcloudAPIToken,
 	)
--- a/pkg/fsutil/configmanager/talos/configs.go
+++ b/pkg/fsutil/configmanager/talos/configs.go
@@ -354,11 +354,16 @@ func (c *Configs) WithHetznerVIP(vip, hcloudAPIToken string) (*Configs, error) {
 	})
 }
 
-// hetznerPublicNICBusPath is the PCI bus path of the public NIC on a Hetzner Cloud
+// HetznerPublicNICBusPath is the PCI bus path of the public NIC on a Hetzner Cloud
 // server. The public interface is always the first virtio device (slot 01); a cluster
 // private network attaches as a later device (slot 07 on the two-NIC servers KSail
 // provisions), so the bus path distinguishes the two where `physical: true` cannot.
-const hetznerPublicNICBusPath = "0000:01:00.0"
+//
+// Exported because the update path reads it too: drift detection has to recognise a
+// VIP that is present but addressed the OLD (name-based) way, so clusters built by an
+// earlier release are migrated rather than reported as already configured. Keep the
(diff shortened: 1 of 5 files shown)


devantler:

**Validation and user-evaluation evidence** (`89b72313b7202db697ca6d36b30cc75496c6f3e6`).

**Root cause, established against live prod (read-only), not inferred:**
- `talosctl --context prod -n 49.13.53.183 get links` → `enp1s0` (bus `0000:01:00.0`, carries public `49.13.53.183`) and `enp7s0` (bus `0000:07:00.0`, private `10.0.1.1`).
- `talosctl get links eth0` → `NotFound` on **both** control planes (Talos v1.13.5). The old patch therefore matched no link, which is why the VIP was never claimed.
- Bus paths are identical on cp-1 and cp-2, so the selector is stable across the pool.

**RED/GREEN proof:** with the original `interface: eth0` rendering restored (valid YAML, so the config still parses), the new assertion fails with `Should be empty, but was eth0`; with the fix, 5/5 pass. The two packages together: **859 tests pass**, `golangci-lint` clean.

**Tried as a user:** rendered the real control-plane machine config through `WithHetznerVIP` and inspected the artifact that lands on nodes —
```
- deviceSelector:
    busPath: 0000:01:00.0
  dhcp: true
  vip:
    ip: 192.0.2.10
    hcloud:
      apiToken: test-token
```
then had Talos itself accept it: `talosctl validate --mode cloud` → *is valid for cloud mode*.

**Honest limit of that evidence:** `talosctl validate` also accepts the *broken* `eth0` form — schema validation cannot see that a link does not exist, which is exactly why this shipped green and stayed broken for a week. The binding itself is only observable after a release is pinned and a deploy runs; #6071's reachability probe means a still-unreachable IP degrades to the node address with a warning rather than persisting a dead endpoint. I will verify the actual bind post-release on devantler-tech/platform#2120.

**Not in scope here:** the apparent second floating IP (point 4 of #6070) is filed separately as #6277.

Full-suite note: `pkg/cli/cmd/open/chat` failed once during `go test ./...` — unrelated to this change, reproduces on a clean tree under the same conditions, and passes here on re-run; characterization refined on #6271.

devantler:

@codex review

devantler:

**P1 confirmed and fixed** in `0e5bdcb92`. This is a genuinely important catch — verified against the code rather than taken on trust:

`hasHetznerFloatingIPConfig` matched the VIP by **address only**, never by how its device is addressed. A cluster created before #6070 carries the HCloud VIP on `interface: eth0`, satisfies `hasHetznerFloatingIPEndpoint`, and hits the `vip.HCloud() != nil && vip.IP() == expectedIP` branch — so `allControlPlanesHaveHetznerFloatingIPConfig` returned true, `mergeFloatingIPChanges` emitted no reconcile, and the new selector was never pushed. The fix would have reached **new clusters only**, leaving every already-deployed cluster with the exact dead endpoint this PR exists to repair.

That is the no-op-in-production class our own contract calls out: validates green, changes nothing on the target's current state.

**Fix:** require the VIP-bearing device to carry the expected `busPath` selector, so the legacy form reads as drift and is migrated in place. `HetznerPublicNICBusPath` is now exported so the build and update paths share one constant rather than drifting apart.

**RED/GREEN proof** (`TestAllControlPlanesHaveHetznerFloatingIPConfig_LegacyEth0FormIsDrift`):
- **GREEN** — current form configured, legacy form drift. 860 tests pass across both affected packages.
- **RED** — reverting only the selector requirement (restoring the pre-fix `return true`) fails the test, so it is pinned to this behaviour and not to something incidental.
- The legacy fixture is derived **source-minus-delta** from the real configured config, matched by pattern rather than a literal — my first attempt used a literal that silently no-opped against the re-encoder's own indentation, which would have "proved" drift detection against an unmodified fixture. Both fixture preconditions are asserted so that failure mode cannot return.

devantler:

@codex review

devantler:

**User-evaluation at head `0e5bdcb92d`** (re-done after the P1 fix, which changed
behaviour for already-deployed clusters).

Rendered the patch an operator actually gets:

```
machine:
  network:
    interfaces:
      - deviceSelector:
          busPath: "0000:01:00.0"
        dhcp: true
        vip:
          ip: "…"
```

Then verified the selector against the **live prod control planes** (read-only,
`talosctl get links`), on both nodes:

| link | busPath |
|---|---|
| `enp1s0` (public) | `0000:01:00.0` — the selector matches |
| `enp7s0` (private) | `0000:07:00.0` — correctly not matched |
| `eth0` | **absent on both nodes** |

So the selector resolves to exactly the public NIC and nothing else, and the
absent `eth0` confirms the original root cause rather than assuming it.

Floating-IP suite green, including
`TestAllControlPlanesHaveHetznerFloatingIPConfig_LegacyEth0FormIsDrift` — the
regression test for the P1: a pre-#6070 cluster now reads as drift and gets
reconciled instead of being reported already-configured.

Issue #6008, opened 2026-07-10
Track Claircore manifest URI SSRF exposure


devantler:
## Description

GitHub Dependabot alerts #165 and #166 report a runtime-scope SSRF vulnerability in `github.com/quay/claircore`: unauthenticated manifest input can reference internal services or cloud metadata endpoints.

KSail currently resolves `github.com/quay/claircore v1.5.35` in both the root and desktop modules. The dependency is reachable transitively from `pkg/client/kubescape` through Kubescape's `claircore/osrelease` path. The alerts mark versions through v1.5.52 as vulnerable and currently report no patched version.

This issue tracks one root cause for both module alerts; do not create separate root/desktop fixes.

## Risk

If a KSail/Kubescape scan path passes attacker-controlled manifest URIs into the vulnerable Claircore flow, the scanner process may make requests to internal or cloud-metadata endpoints. Reachability from KSail's exposed scan inputs must be established rather than assumed from dependency presence alone.

## Proposed direction

1. Trace the KSail workload-scan inputs into Kubescape/Claircore and document whether remote manifest URIs are reachable.
2. If reachable, add the narrowest boundary validation or remote-fetch restriction available without weakening scan coverage.
3. Upgrade both Go modules together as soon as a fixed Claircore release is available.
4. Keep the mitigation temporary and remove it after the patched dependency is adopted.

## Acceptance criteria

- [ ] Reachability from user-controlled KSail inputs is proven or ruled out with a focused test.
- [ ] Any reachable URI path rejects internal/link-local metadata targets or disables remote retrieval at the appropriate boundary.
- [ ] Root and desktop modules use the first patched Claircore release when available.
- [ ] Focused Kubescape client tests and the normal Go validation pass.
- [ ] Dependabot alerts #165 and #166 close after the fix/upgrade.

## Current blocker

GitHub currently reports no patched Claircore version. The issue is still actionable for reachability analysis and a scoped mitigation; only the final dependency bump is externally blocked.

devantler:
> 🤖 Generated by the Daily AI Engineer

## Reachability analysis (AC 1): NOT REACHABLE — ruled out with evidence

**Verdict: the SSRF sink is not linked into the ksail binary, and no user-controlled input could reach it even if it were.**

**Import chain** (`go mod why github.com/quay/claircore`):
```
pkg/client/kubescape → kubescape/v3/core/core → claircore/osrelease → claircore
```

[...]

PR #6448, opened 2026-08-02
fix(deps): adopt patched claircore and re-establish the SSRF verdict

### LINKED ISSUE #6198: chore(deps): claircore SSRF advisory (2 medium alerts) — indirect, no patched version yet

## Evidence

Dependabot alerts #165/#166 (medium): claircore `<= 1.5.52` — *"Unauthenticated attackers can submit manifests with URIs pointing to internal services or cloud metadata endpoints"* (SSRF in claircore's manifest handling). ksail pins `github.com/quay/claircore v1.5.35` **indirect** via the scan toolchain; `first_patched_version` is **null**, so no bump exists today and dependabot cannot propose one.

## Reachability assessment

ksail never runs a claircore service — the vulnerable surface is claircore's own manifest-ingestion endpoints, not the library paths a scanner consumer exercises. Risk to ksail users is accordingly assessed low; this issue exists so the alert is owned, not ignored.

## Acceptance criteria

- [ ] When upstream publishes a patched claircore, bump it (dependabot will propose it; drive that PR to green/merge) and confirm alerts #165/#166 auto-resolve.
- [ ] If no patch ships within ~a quarter, re-assess: dismiss the alerts with the reachability rationale (recorded here) or pin an upstream fix.

Blocked on: upstream patched release (none as of 2026-07-17).

devantler:

## Why

A security advisory against the `claircore` library has been open on KSail since 17 July, tracked as
#6198 and parked on the belief that no patched release existed. A patched release **does** exist —
the `desktop` module has already been running it since 23 July, and its half of the alert closed
automatically at that point. Only the root module was still holding the affected version, so the
alert stayed open with nobody acting on it.

## What

Root moves onto the same patched `claircore` release `desktop` already runs, which takes it outside
the advisory's affected range and should close the remaining alert.

The repository guards this dependency deliberately: it refuses any version change until the
"can an attacker actually reach this?" verdict is re-established, and it forbids one module from
inheriting another module's verdict. That verdict was re-established against the root module's own
dependency graph — the vulnerable component is still not reachable from KSail — and the guard was
retargeted from a hard-coded list of package names to the property it is actually protecting, so it
cannot quietly go stale the next time this happens.

No product behaviour changes; this is a dependency and its safety guard.

Fixes #6198

**Supersedes automation's half-fix:** Dependabot proposed the same version bump in #6342 on 23 July
and it has sat open since. A dependency bot can move the version but cannot re-establish the safety
verdict the repo requires alongside it, so that PR could never complete on its own. This delivers
both halves; #6342 becomes redundant once this lands.


devantler:

## Verification record

**Condition 3 (tried and evaluated as a user).** This change's user-facing surface *is* the guard, so
I exercised the guard rather than tracing it. All runs uncached (`-count=1`), on head `e950d623`.

| Check | Result |
|---|---|
| **GREEN** — full Claircore suite on this branch | `ok` (3.8s) |
| **RED** — `main`'s guard against this branch's patched `go.mod` | **FAIL**, root only: `uses github.com/quay/claircore "v1.5.53"; re-establish the #6008 SSRF reachability verdict before updating the audited version "v1.5.35"` |
| **Ablation A** — drop a package root genuinely links (`osrelease`) from root's audit | **FAIL** naming that exact package; **desktop's subtest still passed** ⇒ per-module independence holds |
| **Ablation B** — admit a package root does *not* link into both audits | **FAIL** on the retargeted tripwire: `root audit admits desktop-audited package … that root does not link; remove it rather than inheriting desktop's verdict` |

Ablation B is the one that matters for the retarget: `TestClaircoreLinkedPackagesStayInert` **did not
fire** there — only the tripwire did. An over-broad audit inheriting another module's verdict is
structurally invisible to the linkage check, which is precisely the gap the retarget closes.

I also measured the audit against reality: both modules link **exactly** the 9 audited packages —
the allow-list is neither over- nor under-broad.

**The fix itself is confirmed against a live oracle, not reasoned.** GitHub's own Dependabot state:

| Alert | Manifest | State | Affected range |
|---|---|---|---|
| 165 | `desktop/go.mod` | **fixed** | `<= 1.5.52` |
| 166 | `go.mod` (root) | **open** | `<= 1.5.52` |

Same advisory, two manifests. GitHub already closed 165 when `desktop` adopted `v1.5.53` — the exact
release this PR moves root onto. That is why #6198 sat parked as "no patched version exists" for 16
days despite `first_patched_version` still reading `null`: the field was never the evidence, the
sibling manifest was.

Ablations restored and the tree verified clean after each.

devantler:



@coderabbitai review

devantler:

## Readiness — self-promoting

| Condition | Evidence |
|---|---|
| **1. Programmatically tested** | 71/71 checks green, 0 failures. Plus, locally and uncached: guard suite GREEN, `main`'s guard RED against this `go.mod`, and both ablations firing on the right guard. |
| **2. Reviewed** | `cr@e950d6232b` — CodeRabbit summary at exactly this head, *"No actionable comments were generated"*, no finding sections, no quota marker. 0 unresolved threads. |
| **3. Tried and evaluated as a user** | The guard **is** this change's user surface and I exercised it directly (record above). The fix itself is confirmed against a live oracle rather than reasoned: GitHub already closed the sibling manifest's alert at this exact release. |

`mergeStateStatus: CLEAN`. Merging.

PR #6479, opened 2026-08-04
fix(cluster): detect and repair Flux artifact-verification drift on update

devantler:

## Why

A cluster can be configured to require signed infrastructure artifacts and still run completely unverified, with nothing anywhere reporting a problem.

Signature verification is configured under `spec.workload.flux.verify`. KSail applies it by patching the OCIRepository that the flux-operator generates — which means it is not one of the fields `cluster update` compares. So configuring verification produced no detected change, `cluster update` had no work to do, and the setting never reached the running cluster. The config read as correct, validation passed, and the artifact source was still accepting anything.

This was found live on a production cluster: the policy had been configured and merged, deploys had run since, and the live resource still carried no verification at all.

## What

`cluster update` now compares verification on the live resource against what the configuration asks for, and repairs the difference — the same way it already handles a rotated registry credential. Because it compares against live state rather than reacting to a config edit, it also repairs a cluster that has drifted for any other reason, and it stays quiet when everything already matches.

A cluster still being built for the first time is deliberately left alone: there is nothing to repair before the resource exists.

Part of devantler-tech/platform#2922

### DIFF
--- a/pkg/cli/cmd/cluster/orchestrator.go
+++ b/pkg/cli/cmd/cluster/orchestrator.go
@@ -932,6 +932,10 @@ func (o *updateOrchestrator) computeUpdateDiff(
 	// Check for registry credential drift (a rotated token behind an unchanged spec)
 	checkRegistryCredentialDrift(o.cmd, o.ctx, diffEngine, diff)
 
+	// Check for signature-verification drift (spec.workload.flux.verify configured
+	// but not applied to the live OCIRepository)
+	checkFluxVerifyDrift(o.cmd, o.ctx, diffEngine, diff)
+
 	promoteUnsupportedInPlaceChanges(updater, diff)
 
 	return currentSpec, diff, nil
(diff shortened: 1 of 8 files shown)


devantler:

Closing: lost the race to #6454, which landed the same fix on `main` while this was in flight. Standing down rather than duplicating — #6454's version is the better of the two, in particular because it shares a single `verifyMatches` between the detector and the patcher, making detector/patcher disagreement structurally impossible instead of merely test-pinned (this PR pinned it with a test).

Cross-lane claim arbitration does not currently work (devantler-tech/monorepo#2302): each instance writes its own branch namespace, so both pushes succeed and both sides believe they won.

The one gap the differential comparison surfaced is filed as #6480 — `checkFluxVerifyDrift` is wired into the Updater diff path but not into `computeSpecOnlyDiff`, which serves `ksail cluster diff` and the non-Updater (VCluster) provisioners. Nothing else here is worth salvaging.

PR #6491, opened 2026-08-05
fix(cluster): check Flux verify drift on the spec-only diff path too

### LINKED ISSUE #6480: Flux verification drift is invisible to `cluster diff` and to non-Updater provisioners

## Evidence

#6454 added `checkFluxVerifyDrift` and wired it into **one** of the two diff paths:

| Path | `checkWorkloadTagDrift` | `checkFluxDistributionVersionDrift` | `checkFluxVerifyDrift` |
|---|---|---|---|
| Updater path (`orchestrator.go` ~927) | ✅ 927 | ✅ 930 | ✅ 937 |
| `computeSpecOnlyDiff` (`orchestrator.go` ~1021) | ✅ 1021 | ✅ 1024 | ❌ **absent** |

Both sibling drift checks are present in `computeSpecOnlyDiff`, which is what makes the omission read as an oversight rather than deliberate scoping.

`computeSpecOnlyDiff` is reachable from two production call sites:

- `pkg/cli/cmd/cluster/diff.go:128` — `handleDiffRunE`, i.e. the user-facing **`ksail cluster diff`**
- `pkg/cli/cmd/cluster/orchestrator.go:128` — `runWithoutUpdater`, for provisioners with no `Updater` (e.g. **VCluster**)

`mergeVersionDrift` does not cover it either — it merges component version *dimensions* through a resolver and never calls the verify check.

1. **Anyone previewing an update.** `ksail cluster diff` never shows a pending verification repair, so the plan and the apply disagree: diff reports nothing, then `cluster update` applies a verify change. A preview that omits a security-relevant repair is worse than no preview, because it is trusted.
2. **Non-Updater provisioners (VCluster).** Verification drift is never detected on those clusters at all — the original devantler-tech/platform#2922 defect, still open for that provisioner class.

## Expected behaviour

`checkFluxVerifyDrift` runs in `computeSpecOnlyDiff` too, so `cluster diff` previews the same verification repair that `cluster update` would apply, and non-Updater provisioners detect it.

## Acceptance criteria

- [ ] `computeSpecOnlyDiff` invokes `checkFluxVerifyDrift` alongside the two existing drift checks.
- [ ] A test pins that a configured-but-unapplied verify block surfaces through the `cluster diff` path, not only the update path — failing before the wiring and passing after.
- [ ] The Flux-engine and bootstrap guards still suppress it (no change emitted for ArgoCD, or when the OCIRepository does not exist).

## Size

Small — one call, plus the test that proves the path.

## Provenance

Found by differential review: this run independently implemented the same fix as #6454 and lost the race. #6454's implementation is otherwise the better of the two — notably it shares one `verifyMatches` between the detector and the patcher, which makes detector/patcher disagreement structurally impossible rather than test-pinned. This call-site asymmetry is the one gap the comparison surfaced. Reasoned from the source at `origin/main` (`56bb93baa`); not reproduced against a live VCluster.

devantler:

## Why

`ksail cluster diff` is a preview people trust before applying a change. It was not showing
pending **artifact-verification** repairs: the check that detects them ran on the update path but
not on the path `cluster diff` uses. So the preview reported nothing, and the following
`cluster update` then applied a security-relevant change nobody was shown.

The same gap meant clusters on provisioners without an updater (VCluster) never detected
verification drift at all.

A preview that silently omits a security repair is worse than no preview, because it is trusted.

## What

The verification drift check now runs on both diff paths, alongside the two sibling checks that
were already on both. Preview and apply now agree, and the affected provisioners detect the drift.

No behaviour changes for anyone already covered: the check keeps its own guards, so it stays
silent for non-Flux clusters and for clusters with no verification configured.

Fixes #6480

### DIFF
--- a/pkg/cli/cmd/cluster/export_test.go
+++ b/pkg/cli/cmd/cluster/export_test.go
@@ -485,6 +485,16 @@ func ExportIsDowngrade(current, target string) bool {
 // ErrMetricsServerDisableUnsupported exports the sentinel error for testing.
 var ErrMetricsServerDisableUnsupported = errMetricsServerDisableUnsupported
 
+// ExportComputeSpecOnlyDiff exposes computeSpecOnlyDiff for unit testing. This is the diff
+// path behind `ksail cluster diff` and behind every provisioner with no Updater, so it is
+// the path whose drift-check coverage has to be assertable independently of the update path.
+func ExportComputeSpecOnlyDiff(
+	cmd *cobra.Command,
+	ctx *localregistry.Context,
+) *clusterupdate.UpdateResult {
+	return computeSpecOnlyDiff(cmd, ctx)
+}
+
 // ExportHandlerForField reports whether a registered handler exists for the given field name.
 func ExportHandlerForField(cmd *cobra.Command, clusterCfg *v1alpha1.Cluster, field string) bool {
 	r := newComponentReconciler(cmd, clusterCfg, "test-cluster")
--- a/pkg/cli/cmd/cluster/flux_verify_reconcile_test.go
+++ b/pkg/cli/cmd/cluster/flux_verify_reconcile_test.go
@@ -1,10 +1,13 @@
 package cluster_test
 
 import (
+	"bytes"
+	"strings"
 	"testing"
 
 	"github.com/devantler-tech/ksail/v7/pkg/apis/cluster/v1alpha1"
 	"github.com/devantler-tech/ksail/v7/pkg/cli/cmd/cluster"
+	"github.com/devantler-tech/ksail/v7/pkg/cli/setup/localregistry"
 	specdiff "github.com/devantler-tech/ksail/v7/pkg/svc/diff"
 	"github.com/devantler-tech/ksail/v7/pkg/svc/provisioner/cluster/clusterupdate"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


devantler:

**RED/GREEN proof, and the control on the ablation.**

Two tests were added on the `cluster diff` path: one that the verify check runs for a
Flux-with-verification cluster, one that it stays silent for ArgoCD and for a cluster with no
verify block.

- **GREEN** — with the wiring: package passes (`go test ./pkg/cli/cmd/cluster/`, 11.4s).
- **RED** — removing only the `checkFluxVerifyDrift(...)` call (5 lines added vs `main` becomes 4;
  the comment stays) fails `TestSpecOnlyDiffChecksFluxVerifyDrift`, and **`go vet` still exits 0 on
  the ablated tree**, so the failure is the assertion rather than a build break. The guard test keeps
  passing in both states, which is what it is for.

The check is detected by its own observable effect — one of its two distinctive warnings, or a
`FluxVerifyField` change on the diff when the cluster query succeeds — rather than by requiring a
live cluster. The property under test is *which diff path invokes the check*, and that has to be
provable without one. Both warning strings are unique to this check; every sibling drift warning
names the distribution version, the sync ref, or registry credentials instead.

**Live context that motivated the priority.** Read from prod today, the root source is still
unverified — `ocirepository/flux-system`, `spec.verify: null`, tracking mutable `latest`,
`Ready=True`. Two further findings recorded on
[platform#2922](https://github.com/devantler-tech/platform/issues/2922): the `FluxInstance` CRD has
no `verify` key under `spec.sync`, so the operator cannot declare it; and server-side-apply
ownership shows `flux-operator` owns only `interval/provider/ref/secretRef/url`, never `f:verify`,
so a patch from another manager is not reconciled away. This PR is the preview/non-Updater half of
that chain, not the whole of it.

**Adjacent gap, deliberately not fixed here** (one concern per PR): `checkRegistryCredentialDrift`
is also on the Updater path only. Same shape as this bug. Filing it separately rather than widening
this diff.

devantler:

## User evaluation — exercised the real binary, with a control

Ran the built CLI as a user would, not just the unit tests. Two binaries from this branch, differing in exactly the six-line change:

- **fixed** — `401d98bd` (this PR's head)
- **control** — `d492c071` (its parent)

Same `ksail.yaml`, same `ksail cluster diff` invocation, same host, no cluster reachable.

| Config | control | fixed |
|---|---|---|
| Flux + `verify.provider: cosign` | **0** | **1** |
| ArgoCD + `verify.provider: cosign` | — | **0** |
| Flux, `verify` unconfigured | — | **0** |

(count = occurrences of the verify check's own warning)

On the fixed binary the preview now emits a fourth drift line alongside its two siblings:

```
⚠ Cannot query current GitOps sync ref for drift detection: ...
⚠ Cannot query current Flux distribution version for drift detection: ...
⚠ Cannot compare Flux artifact verification for drift detection: ...
```

The control emits only the first two — so `cluster diff` genuinely never reached `checkFluxVerifyDrift` before this change, and does now. That is the user-visible symptom #6480 describes: a preview that silently omitted a repair the following `cluster update` would apply.

Both negative directions were exercised too, so wiring the check into a second path did not make it fire where it must stay silent: the Flux-engine guard and the not-configured guard each still return 0 on the fixed binary.

devantler:



@coderabbitai review

devantler:

## Readiness record

**Programmatically tested** — 70/70 checks green at `cc6f55a6`; 0 review threads; `mergeStateStatus: CLEAN`.

Note on this PR's earlier red: the failing `System Test (Docker) (Talos, Docker, true)` leg was a **runner abort, not a test failure** — the job reported zero step conclusions and its log blob returned `BlobNotFound`. 26 of 27 legs were green, including every Flux and ArgoCD leg (the ones this diff actually touches), and the *same* leg passed on #6489 in an overlapping window. Re-running cleared it; the branch was then updated onto current `main`.

**Tried and evaluated as a user** — exercised at the exact PR head (`cc6f55a671ef45d8a31f3108fc6ce7683b0b5eb2`) as a RED/GREEN ablation:

| Arm | Code | Result |
|---|---|---|
| GREEN | head unmodified | `ok …/pkg/cli/cmd/cluster 2.404s` |
| RED | the added `checkFluxVerifyDrift(...)` call removed | `--- FAIL: TestSpecOnlyDiffChecksFluxVerifyDrift` |
| GREEN | restored | `ok … 11.739s` (full package), tree clean |

Controls on the ablation itself, so the RED arm cannot be passing for the wrong reason:

- `go build ./pkg/cli/cmd/cluster/` succeeds under the ablation — the failure is behavioural, not a compile error — and the run used `-vet=off`.
- The mutation was inspected rather than assumed: the diff is exactly the call line plus an adjacent blank. The explanatory comment above it survives, so the semantic change is precisely "the call no longer happens". (I had predicted a 1-line delta and it was 2 — checking rather than trusting the count is what confirmed the second line was only whitespace.)

The RED failure message states the user-visible consequence directly: `computeSpecOnlyDiff must run checkFluxVerifyDrift, so cluster diff previews the same verification repair cluster update applies`. Without this, a preview disagrees with the apply it previews, and verification drift is undetectable on every provisioner that has no Updater.

**Reviewed** — CodeRabbit requested at this head; promotion waits on its result.

PR #6503, opened 2026-08-05
fix(ci): pin the system test's cluster update to the version create used

### LINKED ISSUE #6502: System test's no-op cluster update fails whenever upstream k3s publishes a newer patch

## Evidence

`🧪 System Test (Docker) (K3s, Docker, true)` fails at the `system-test.cluster-update` step:

```
❌ ERROR: Unexpected changes detected during update after cluster creation
This indicates a regression — the cluster was just created with the same configuration.
```

The step's own output shows why, and it is not a regression:

```
► discovering available Kubernetes versions from rancher/k3s
ℹ Kubernetes upgrade path: v1.36.2-k3s1 → v1.36.3-k3s1 (1 step(s))
ℹ Kubernetes upgrade from v1.36.2-k3s1 to v1.36.3-k3s1 requires cluster recreation
```

**Controlled measurement, same branch, same job name, same loaded image:**

| run | time | image loaded | `cluster update` result |
|---|---|---|---|
| [31034652808](https://github.com/devantler-tech/ksail/actions/runs/31034652808) | 18:25Z | `rancher/k3s:v1.36.2-k3s1` | `No changes detected` → **pass** |
| [31043790395](https://github.com/devantler-tech/ksail/actions/runs/31043790395) | 20:44Z | `rancher/k3s:v1.36.2-k3s1` (identical) | upgrade path to `v1.36.3-k3s1` → **fail** |

The diff between the two runs is **not** in this repository — the only variable is when ksail queried
upstream. `v1.36.3+k3s1` was published 2026-08-04T19:42Z and became visible to version discovery
between those two runs.

## Root cause

The test creates the cluster from a **workflow-cached image pinned at `v1.36.2-k3s1`**, but
`ksail cluster update` **resolves the current upstream version live**. Those two sources are allowed
to drift, so the moment upstream publishes/promotes a newer patch, `update` correctly reports
"recreation required" against a cluster built from the stale cached image — and the step's
hard-coded `grep -q "No changes detected"` turns that correct behaviour into a red build.

This is a **time bomb, not a flake**: it fires for every PR and for `main` once upstream moves, and
keeps firing until the cached image is refreshed. It is unrelated to whatever change is being tested.

Note the sibling step `system-test.cluster-update-dry-run` does **not** fail, because its assertion
also accepts `Would apply 0 in-place, 0 reboot-required, 0 recreate-required`. Only the strict
`cluster update` assertion is exposed.

## Expected behaviour

The invariant the test wants — "creating and then updating with the same config is a no-op" — should
be tested against a **single source of truth for the Kubernetes version**, so an upstream release
cannot make it fail. Any of:

- pin the version explicitly for the system test and use that same pin for both create and update; or
- have the test derive its expectation from the version actually resolved at create time; or
- refresh the cached image in the same step that resolves the version, so the two cannot drift.

Whichever is chosen, an upstream patch release must not red the build.

## Acceptance criteria

- Publishing a new upstream k3s patch does not fail this step.
- The no-op-update invariant is still genuinely asserted (not weakened to accept real drift).
- A regression check pins create/update to the same resolved version so the two sources cannot
  silently diverge again.

Rough size: small–medium, confined to the system-test workflow and its version resolution.

devantler:

## Why

Every KSail PR — and `main` itself — goes red whenever upstream publishes a new k3s patch
release, for a reason that has nothing to do with the change being tested. It is a time bomb,
not a flake: it fires for everyone at once and keeps firing until someone refreshes a pinned
image. It is armed right now (upstream published `v1.36.3-k3s1` on 2026-08-04; we pin
`v1.36.2-k3s1`), and it is currently blocking #6501.

The cause is that one value has two sources that are free to drift apart: the system test
creates the cluster from a version KSail pins in the repo, but then asks `cluster update`
whether anything changed — and an unpinned update goes and looks up the newest version
published upstream. So the answer is "yes, recreate the cluster", which the test correctly
reads as a regression.

## What

Give the two steps one source of truth: the test now tells `cluster update` to target the same
pinned version the cluster was built from, read out of the very file KSail itself reads. When
Renovate or Dependabot bumps that pin, both sides move together, so they cannot silently
diverge again.

The no-op check itself is unchanged and just as strict — this removes a false alarm, it does
not weaken what the test asserts. Real configuration drift still fails the build.

Distributions that do not take their Kubernetes version from a pinned node image (Talos,
VCluster, KWOK, EKS) are untouched and behave exactly as before.

Fixes #6502

### DIFF
--- a/.github/actions/ksail-system-test/action.yaml
+++ b/.github/actions/ksail-system-test/action.yaml
@@ -391,6 +391,58 @@ runs:
       with:
         workload-timeout: ${{ inputs.workload-timeout }}
 
+    - name: 🔖 Resolve Kubernetes version pin for update
+      id: k8s-version-pin
+      shell: bash
+      env:
+        DISTRIBUTION: ${{ inputs.distribution }}
+      run: |
+        set -euo pipefail
+        # `cluster create` provisions from the node image KSail pins in
+        # pkg/fsutil/configmanager/<dist>/Dockerfile, but an *unpinned*
+        # `cluster update` resolves the newest version live from the registry.
+        # Those are two different sources for one version, so the moment upstream
+        # publishes a newer patch they disagree and a no-op update correctly
+        # reports "recreation required" — reddening every build for a reason that
+        # has nothing to do with the change under test.
+        #
+        # Pin the update to the tag create used, read from the very Dockerfile
+        # KSail parses, so a Renovate/Dependabot bump moves both sides together
+        # and the two sources cannot silently diverge again.
+        #
+        # Only distributions whose Kubernetes version comes from a pinned node
+        # image are covered. Talos pins an OS version on a separate axis,
+        # VCluster bakes its Kubernetes version into the SDK image, and KWOK/EKS
+        # have no pinned node image — those keep today's unpinned behaviour.
+        case "$DISTRIBUTION" in
+          K3s)     dockerfile="pkg/fsutil/configmanager/k3d/Dockerfile";  repo="rancher/k3s" ;;
+          Vanilla) dockerfile="pkg/fsutil/configmanager/kind/Dockerfile"; repo="kindest/node" ;;
+          *)
+            echo "ℹ️ $DISTRIBUTION has no pinned Kubernetes node image — leaving the update unpinned"
+            echo "flag=" >> "$GITHUB_OUTPUT"
+            exit 0
+            ;;
+        esac
+
+        if [[ ! -f "$dockerfile" ]]; then
+          echo "❌ ERROR: expected $dockerfile to exist for $DISTRIBUTION"
+          exit 1
+        fi
+
+        # Mirror KSail's own parse (image/parser.ParseImageFromDockerfile):
+        # FROM <repo>:<tag>[@sha256:<digest>] — take the tag, drop the digest.
+        # The tag must be passed WHOLE: KSail compares pinned vs running version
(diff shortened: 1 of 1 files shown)


devantler:


CI is settled and green (43 success, 0 failing) and the change is complete at this head. Self-review done. Requesting the priority-1 lane.

@coderabbitai review

devantler:

## Readiness — self-promoting at `8f2ccb17f0`

1. **Programmatically tested** — 46 checks green, 0 failing, 0 pending at this head.
2. **Reviewed** — CodeRabbit review completed at `8f2ccb17f0` (status `success :: Review completed`, 23:25:28Z) with its summary comment naming this head and reporting *"No actionable comments were generated in the recent review"*. Zero unresolved threads, no findings.
3. **Tried and evaluated as a user** — RED/GREEN control against the sibling PR that shares this base: the five `System Test (Docker) (K3s, …)` jobs are **FAILURE on #6501** (without this pin) and **SUCCESS here** (with it). That is the defect in #6502 reproduced and then fixed, observed behaviourally rather than reasoned. The Vanilla control passes in both, so the pin is not masking anything broader.

Note on the parse: the tag is passed **whole** (`v1.36.2-k3s1`, not `v1.36.2`) because KSail compares pinned-vs-running by semver *including* the suffix — a stripped tag would read as a downgrade instead of a no-op. Both parsed Dockerfiles carry exactly one `FROM`.

Issue #6373, opened 2026-07-27
CI reports main green while the Go test suite is skipped, hiding real breakage


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`main` was red on `internal/ciharness` and CI reported it green.

`a5ed9f85` ("fix(ci): judge EKS smoke teardown by whether a cluster remains", #6363) moved the EKS
teardown guard out of the workflow step's inline `Run` and into
`.github/scripts/delete-eks-smoke-cluster.sh`, but left
`TestEKSSmokePreparesCloudGitOpsAndBoundsCleanup` asserting the guard string in `cleanupStep.Run`.
That test fails on a clean worktree at unmodified `main`.

The reason it went unnoticed: on that push to `main`, **`✅ Validate Go Project / 🧪 Test` was
`skipped`**, so the unit tests never ran for the commit that broke them. Every subsequent `main` run
also reported success without executing them.

It only surfaced on an unrelated PR that happened to touch Go files (#6267), where the same failure
appeared and cancelled 27 downstream System Test jobs — presenting as *that* PR's problem rather than
as pre-existing breakage on `main`.

## Impact

A change can break the Go test suite on `main` and leave `main` reporting green. The breakage is then
inherited by the next unrelated PR, where it costs that PR's author a diagnosis and blocks unrelated
work behind a failure they did not cause. It also means the branch-protection signal on `main` is not
what it appears to be for Go changes.

## Expected

A push to `main` that changes Go sources, workflows, or CI scripts runs the Go test suite. The narrow
question is why the path filter did not consider `#6363` a Go-affecting change: it touched
`.github/workflows/**`, `.github/scripts/**`, and `internal/ciharness/**` — the last of which is Go.

## Acceptance criteria

- [x] Reproduce the skip: identify which filter/condition made `🧪 Test` skip on `a5ed9f85`.
- [x] A change to `internal/**` or `pkg/**` can never skip the Go test job on a push to `main`.
- [x] A regression test or a deliberate probe proves the job now runs for that path set.
- [x] Confirm no *other* required check is skippable on `main` in the same way (the skip pattern is
      the defect; this instance is one symptom).

## Size

Small-to-medium, and mostly diagnosis. The immediate red is already fixed separately in #6372 — this
issue is about the gap that let it reach `main` unseen, which is the more valuable half.

devantler:
> 🤖 Generated by the Agentic Engineer

## The same skip also hides the security scan — measured today

[...]

PR #6506, opened 2026-08-06
ci: stop cancelling in-progress runs on main

### LINKED ISSUE #6384: Runs on main are cancelled by the next merge, so a commit's checks never complete

## Evidence

`concurrency.cancel-in-progress: true` in `.github/workflows/ci.yaml` is keyed on
`ci-ksail-${{ github.workflow }}-${{ github.ref }}`. On `main` every merge shares one `github.ref`, so
a merge landing while the previous merge's checks are still running **evicts them**.

That is how the breakage in #6373 reached `main` unseen. `59cb206dc8` (the `#6363` merge) changed
`internal/ciharness/system_test_harness_test.go`, which matches the Go path filter — the tests were
not skipped, they were **cancelled** mid-flight by the next merge. `#6373`'s own evidence table records
`59cb206dc8` as `cancelled` for both `🧪 Test` and `🛡️ Vulnerability Scan`.

A cancelled check is not a failed check, so branch protection and the commit status both settle on a
green-looking `main` while the verification that would have caught the regression never completed.

Everyone who trusts `main`'s status, and every subsequent PR author. The eviction is silent and
timing-dependent, so it is most likely exactly when it hurts most — during a burst of merges — and the
inherited breakage then presents as the *next* unrelated PR's problem.

## Expected behaviour

On a pull request, cancelling superseded runs is correct and should stay: only the newest commit's
result matters.

On the default branch it is wrong. Each merge commit's run **is** the verification record for that
commit, and it is not superseded by a later commit — a later commit is a different subject. Runs on
`main` should be allowed to finish.

## Acceptance criteria

- [ ] Runs on the default branch are not cancelled by a subsequent push to the default branch.
- [ ] Cancellation behaviour on pull requests is unchanged (superseded PR runs still cancel).
- [ ] A deliberate probe demonstrates the change: two merges to `main` in quick succession both retain
      completed `🧪 Test` results, where today the first is `cancelled`.
- [ ] Check whether any other workflow in this repo shares the same ref-keyed cancellation on `main`
      (`cd.yaml`, the system-test workflows) and fix or reason about each.

The usual shape is `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`, but confirm that
against how the merge queue interacts with it before adopting it — this repo gates `main` behind a
merge queue, and `merge_group` refs must not be starved either.

## Size

Small — a one-line expression plus the probe. The care is in the merge-queue interaction, not the edit.

Part of #6373

devantler:

## Why

`main` can report green on tests that never ran. Every merge to `main` shares one concurrency key, so a merge landing while the previous merge's checks are still running **cancels them** — and a cancelled check is not a failed check, so the commit status settles green anyway.

That is how a broken Go test suite reached `main` unnoticed and then surfaced as an unrelated PR author's problem days later.

## What

Runs on `main` are now allowed to finish. Superseded pull-request and merge-queue runs still cancel exactly as before, so the fast-feedback behaviour developers rely on is unchanged — only the default branch, where each commit's run is its own verification record, stops being evicted.

Also adds a guard so no future workflow can reintroduce the same eviction on `main`.

Fixes #6384
Part of #6373

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,11 @@ env:
 
 concurrency:
   group: "ci-ksail-${{ github.workflow }}-${{ github.ref }}"
-  cancel-in-progress: true
+  # Superseded pull-request and merge-queue runs still cancel: each carries its
+  # own ref, so the group key is unique per subject. Every push to main shares
+  # one ref, so cancelling there would evict the previous merge's checks
+  # mid-flight and settle main on verification that never completed.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 permissions: {}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,11 +25,15 @@ env:
   GOMEMLIMIT: "6GiB"
 
 concurrency:
-  group: "ci-ksail-${{ github.workflow }}-${{ github.ref }}"
-  # Superseded pull-request and merge-queue runs still cancel: each carries its
-  # own ref, so the group key is unique per subject. Every push to main shares
-  # one ref, so cancelling there would evict the previous merge's checks
-  # mid-flight and settle main on verification that never completed.
+  # On main each run is its own verification record, so it gets a run-unique
+  # group. Sharing a group there is not made safe by cancel-in-progress alone: a
+  # group holds one running plus one pending run, and a third merge cancels the
+  # pending one — so the second merge's checks would still be lost. A unique key
+  # means nothing on main ever queues behind, or evicts, anything else.
+  #
+  # Pull requests and merge-queue entries keep the ref-keyed group, so superseded
+  # runs there still cancel exactly as before.
+  group: "ci-ksail-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}"
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 permissions: {}
--- a/internal/ciharness/ci_workflow_test.go
+++ b/internal/ciharness/ci_workflow_test.go
@@ -100,10 +100,11 @@ func TestNoDefaultBranchWorkflowCancelsRunsInProgress(t *testing.T) {
 
 		assert.Falsef(
 			t,
-			cancelsUnconditionally(workflow.Concurrency.CancelInProgress),
-			"%s runs on main and cancels in progress unconditionally, so one merge evicts "+
-				"the previous merge's checks before they complete",
+			mayCancelDefaultBranchRuns(workflow.Concurrency.CancelInProgress),
+			"%s runs on main and may cancel in progress there, so one merge evicts the "+
+				"previous merge's checks before they complete (allowed: absent, false, or %s)",
 			filepath.Base(path),
+			approvedCancelExpression,
 		)
 	}
 
@@ -120,23 +121,70 @@ func TestNoDefaultBranchWorkflowCancelsRunsInProgress(t *testing.T) {
 	)
 }
 
-// cancelsUnconditionally reports whether a cancel-in-progress value cancels runs
-// on every branch. A literal `true` is the obvious form, but GitHub also accepts
-// an expression, and YAML hands those back as strings — so `${{ true }}` renders
-// as a non-"true" string while still cancelling everything. An expression that
-// never mentions github.ref cannot be branch-conditional either, so it is
-// treated as unconditional rather than assumed safe.
-func cancelsUnconditionally(value any) bool {
-	rendered := strings.TrimSpace(fmt.Sprintf("%v", value))
-	if rendered == "true" {
-		return true
-	}
+// approvedCancelExpression is the only expression form allowed to gate
+// cancellation on a workflow that runs on the default branch.
+const approvedCancelExpression = "${{ github.ref != 'refs/heads/main' }}"
 
-	if !strings.HasPrefix(rendered, "${{") || !strings.HasSuffix(rendered, "}}") {
+// mayCancelDefaultBranchRuns reports whether a cancel-in-progress value could
+// cancel a run on main. It FAILS CLOSED: only an absent value, an explicit
+// false, or the exact approved expression are accepted.
+//
+// Inspecting the tokens of an arbitrary expression does not work, because what
+// matters is the value it evaluates to when github.ref is refs/heads/main —
+// which a substring check cannot tell. Both `${{ github.ref && true }}` and
+// `${{ github.ref != 'refs/heads/main' || true }}` mention github.ref and look
+// branch-conditional, yet both evaluate true on main and cancel the previous
+// run. Evaluating GitHub expressions here would mean shipping an interpreter
+// that itself needs testing, so an allowlist is the honest guard: a new form is
+// added deliberately, with its behaviour on main reasoned about once.
+func mayCancelDefaultBranchRuns(value any) bool {
+	if value == nil {
 		return false
 	}
 
-	inner := strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(rendered, "${{"), "}}"))
+	switch strings.TrimSpace(fmt.Sprintf("%v", value)) {
+	case "false", approvedCancelExpression:
+		return false
+	default:
+		return true
+	}
+}
 
-	return inner == "true" || !strings.Contains(inner, "github.ref")
+func TestMayCancelDefaultBranchRuns(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		value any
+		want  bool
+	}{
+		"absent":              {value: nil, want: false},
+		"explicit false":      {value: false, want: false},
+		"approved expression": {value: approvedCancelExpression, want: false},
+
+		"literal true":    {value: true, want: true},
+		"quoted true":     {value: "true", want: true},
+		"expression true": {value: "${{ true }}", want: true},
+
+		// Both mention github.ref and read as branch-conditional, yet each
+		// evaluates true when github.ref is refs/heads/main.
+		"truthy ref conjunction": {value: "${{ github.ref && true }}", want: true},
+		"disjunction with true": {
+			value: "${{ github.ref != 'refs/heads/main' || true }}",
+			want:  true,
+		},
+
+		// Fail closed on anything not explicitly approved, including forms that
+		// may well be correct — adding one is a deliberate act.
+		"unreviewed variant": {
+			value: "${{ github.ref != format('refs/heads/{0}', 'main') }}",
+			want:  true,
+		},
+	}
+
+	for name, testCase := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, testCase.want, mayCancelDefaultBranchRuns(testCase.value))
+		})
+	}
 }
--- a/internal/ciharness/ci_workflow_test.go
+++ b/internal/ciharness/ci_workflow_test.go
@@ -59,20 +59,16 @@ func TestCIWorkflowKeepsDefaultBranchRunsAlive(t *testing.T) {
 		"main must be bound to the run-unique key; an inverted condition gives main the shared ref key",
 	)
 
-	// A ref-keyed group gives every push to main the same key, so cancelling
-	// unconditionally evicts the previous merge's checks mid-flight. main then
-	// reports green on verification that never finished.
-	cancel, isExpression := workflow.Concurrency.CancelInProgress.(string)
-	require.Truef(
+	// Hold cancel-in-progress to the same allowlist the repo-wide test uses.
+	// Substring checks would be misleading here for exactly the reason they were
+	// wrong there: `${{ github.ref != 'refs/heads/main' || true }}` contains
+	// "github.ref", "refs/heads/main" and "!=" while still cancelling on main.
+	assert.Equalf(
 		t,
-		isExpression,
-		"cancel-in-progress must be an expression excluding the default branch, got %#v",
+		approvedCancelExpression,
 		workflow.Concurrency.CancelInProgress,
+		"cancel-in-progress must be exactly the approved expression",
 	)
-
-	assert.Contains(t, cancel, "github.ref")
-	assert.Contains(t, cancel, "refs/heads/main")
-	assert.Contains(t, cancel, "!=")
 }
 
 func TestNoDefaultBranchWorkflowCancelsRunsInProgress(t *testing.T) {
--- a/internal/ciharness/ci_workflow_test.go
+++ b/internal/ciharness/ci_workflow_test.go
@@ -1,0 +1,103 @@
+package ciharness_test
+
+import (
+	"os"
+	"path/filepath"
+	"slices"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+// concurrencyWorkflow captures the trigger and concurrency surface that decides
+// whether a run on the default branch can be evicted by a later one.
+//
+//nolint:tagliatelle // GitHub Actions defines these external keys in kebab-case.
+type concurrencyWorkflow struct {
+	On struct {
+		Push struct {
+			Branches []string `yaml:"branches"`
+		} `yaml:"push"`
+	} `yaml:"on"`
+	Concurrency struct {
+		Group            string `yaml:"group"`
+		CancelInProgress any    `yaml:"cancel-in-progress"`
+	} `yaml:"concurrency"`
+}
+
+func TestCIWorkflowKeepsDefaultBranchRunsAlive(t *testing.T) {
+	t.Parallel()
+
+	contents := readRepoFile(t, ".github/workflows/ci.yaml")
+
+	var workflow concurrencyWorkflow
+	require.NoError(t, yaml.Unmarshal(contents, &workflow))
+
+	// The group stays ref-keyed so superseded pull-request runs still cancel:
+	// each pull request carries its own refs/pull/<n>/merge ref, and each merge
+	// queue entry its own gh-readonly-queue ref.
+	assert.Contains(t, workflow.Concurrency.Group, "github.ref")
+
+	// A ref-keyed group gives every push to main the same key, so cancelling
+	// unconditionally evicts the previous merge's checks mid-flight. main then
+	// reports green on verification that never finished.
+	cancel, isExpression := workflow.Concurrency.CancelInProgress.(string)
+	require.Truef(
+		t,
+		isExpression,
+		"cancel-in-progress must be an expression excluding the default branch, got %#v",
+		workflow.Concurrency.CancelInProgress,
+	)
+
+	assert.Contains(t, cancel, "github.ref")
+	assert.Contains(t, cancel, "refs/heads/main")
+	assert.Contains(t, cancel, "!=")
+}
+
+func TestNoDefaultBranchWorkflowCancelsRunsInProgress(t *testing.T) {
+	t.Parallel()
+
+	paths, err := filepath.Glob(filepath.Join("..", "..", ".github", "workflows", "*.y*ml"))
+	require.NoError(t, err)
+	require.NotEmpty(t, paths)
+
+	checked := 0
+
+	for _, path := range paths {
+		// The glob supplies repository-owned workflow paths, never user input.
+		contents, readErr := os.ReadFile(path) //nolint:gosec
+		require.NoError(t, readErr)
+
+		var workflow concurrencyWorkflow
+		require.NoErrorf(t, yaml.Unmarshal(contents, &workflow), "parsing %s", path)
+
+		if !slices.Contains(workflow.On.Push.Branches, "main") {
+			continue
+		}
+
+		checked++
+
+		assert.NotEqualf(
+			t,
+			true,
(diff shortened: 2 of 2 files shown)


devantler:

## What I exercised, and what I could not

**Pull-request cancellation still works — observed on GitHub's real evaluator, not reasoned.** This PR got a free natural experiment: I pushed twice, 73 seconds apart, with the new expression already in place.

| head | `CI - KSail` run |
|---|---|
| `33869dbf` (first push) | **cancelled** |
| `c5233109` (second push) | proceeded |

For a `pull_request` event `github.ref` is `refs/pull/6506/merge`, so `${{ github.ref != 'refs/heads/main' }}` evaluates **true**, cancellation stays on, and the superseded run was evicted exactly as before. That satisfies acceptance criterion 2 and also proves GitHub parses and evaluates the expression at workflow level — a malformed one would have failed the run outright.

**The default-branch half cannot be proven pre-merge.** It is the complementary branch of that same evaluated expression, but demonstrating it needs two merges to `main` in quick succession (acceptance criterion 3). I will verify it post-merge against the first pair of closely-spaced merges rather than claim it now.

## A correction to the issue's premise

#6384 says *"this repo gates `main` behind a merge queue"* and asks that the fix be checked against it. **It does not, today.** Verified live: `repos/devantler-tech/ksail/rulesets` contains `pull_request`, `non_fast_forward`, `code_quality` and `code_scanning` rules and **no `merge_queue` rule**, and classic branch protection returns `404 Branch not protected`. The commented-out `cancel-stale-merge-queue` job in `ci.yaml` ("kept for future use when jobs are moved to `merge_group`") is consistent with that.

So the merge-queue interaction is currently inert. The expression is still correct if a queue is enabled later: each queue entry gets its own `refs/heads/gh-readonly-queue/...` ref, which is `!= refs/heads/main`, so those runs keep cancelling per-entry and are never starved.

## Scope check

`ci.yaml` is the only workflow in this repo with a truthy `cancel-in-progress` that can run on `main` — acceptance criterion 4. The other four workflows triggering on push to `main` (`desktop`, `release`, `todos`, `web-ui`) carry no `concurrency` block at all, so there is nothing to gate; the added repo-wide test pins that and will fail if any of them grows one.

The same defect exists in five other repositories — tracked separately as devantler-tech/monorepo#2690, not fixed here.

devantler:



Requesting review at `c523310967b3`. The code-relevant checks are green — `🧪 Test`, `🧹 Lint - golangci-lint`, `🏗️ Build`, `🧹 Lint - mega-linter` (which runs `actionlint` over the changed workflow) — with 33 passing and 0 failing. The remaining pending jobs are the 27-job System Test cluster matrix, which provisions clusters and cannot be affected by a workflow-level concurrency expression or a workflow-parsing unit test.

@coderabbitai review

devantler:

Valid, and it defeats acceptance criterion 1 of #6384 as I had implemented it — thank you. A concurrency group holds one *running* plus one *pending* run, and a third arrival cancels the pending one. So `cancel-in-progress: false` protected the running job while the second merge's checks could still be evicted while queued. The fix was half a fix.

Adopted the run-unique key for `main` in `0675cabb`:

```yaml
group: "ci-ksail-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}"
cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

Nothing on `main` now queues behind or evicts anything else. Pull requests and merge-queue entries keep the ref-keyed group, so superseded runs there still cancel — re-verified live: my previous two pushes produced `cancelled` for `33869dbf` and a live run for `c523310967b3`.

`cancel-in-progress` stays ref-conditional rather than being dropped: with a run-unique key on `main` it is redundant there, but it keeps the intent explicit and fails safe if the group expression is ever simplified back.

Test updated to pin the new mechanism (`github.run_id` and `refs/heads/main` in the group), and ablated — reverting the group to the shared ref-only form fails on both assertions.

devantler:



Re-requesting at `0675cabb06f4`. Your previous review was at `c523310967b3` and is superseded by the fix for the pending-run finding — thank you for it, it was a real gap.

Current state at this head: 32 passing, 0 failing; the code-relevant checks (`🧪 Test`, `🧹 Lint - golangci-lint`, `🏗️ Build`, `🧹 Lint - mega-linter`) are all green. The 4 outstanding jobs are the System Test cluster matrix, which cannot be affected by a workflow-level concurrency expression.

The same correction has been applied to the two sibling PRs — devantler-tech/platform#2992 (reviewed clean, now in the merge queue) and devantler-tech/kyverno-policies#35.

@coderabbitai review

devantler:

Both valid, and both were real holes in the regression protection rather than style points. Fixed in `f9d67435`.

**1. Inverted expression.** Confirmed — asserting `github.ref`, `github.run_id` and `refs/heads/main` as independent substrings passes for `github.ref != 'refs/heads/main' && github.run_id || github.ref`, which gives `main` the shared ref key and reinstates exactly the pending-run eviction this PR fixes. Now asserting the bound conditional:

```go
assert.Containsf(t, workflow.Concurrency.Group,
    "github.ref == 'refs/heads/main' && github.run_id", ...)
```

Ablated: flipping `==` to `!=` in the workflow fails with *"does not contain `github.ref == 'refs/heads/main' && github.run_id`"*.

**2. `${{ true }}`.** Confirmed — YAML hands that back as a string, so the literal `"true"` comparison passed over it while GitHub still cancels unconditionally. Replaced with a helper that rejects a literal `true`, an expression whose body is `true`, and any expression that never references `github.ref` (which therefore cannot be branch-conditional at all).

Ablated: setting `cancel-in-progress: ${{ true }}` now fails the repo-wide test.

This is the second time review has caught that my *fix* was right while my *proof* was weaker than it looked — the first being the pending-run gap itself. Both are worth having.

`go test ./internal/ciharness/` green, `golangci-lint` 0 issues. Re-requesting once CI settles at the new head.

devantler:



Re-requesting at `f9d67435068d`, which carries the fixes for both regression-test gaps you raised at `0675cabb`:

1. The group assertion now pins the **bound conditional** `github.ref == 'refs/heads/main' && github.run_id`, so the inverted form no longer passes. Ablated.
2. `cancel-in-progress` is now checked by a helper that rejects a literal `true`, an expression whose body is `true` (`${{ true }}`), and any expression that never references `github.ref`. Ablated.

State at this head: 33 passing, 0 failing, 0 unresolved threads. The code-relevant checks — `🧪 Test`, `🧹 Lint - golangci-lint`, `🏗️ Build`, `🧹 Lint - mega-linter` — are all green; the outstanding jobs are the System Test cluster matrix, which a workflow-level concurrency expression cannot affect.

The same change is now merged in three sibling repos: devantler-tech/kyverno-policies#35, devantler-tech/platform#2992 and devantler-tech/agent-plugins#113.

@coderabbitai review

devantler:

Valid — and this is the third time the *fix* held while my *proof* did not, so I've stopped trying to be clever about it. Fixed in `e4a448ad`.

Both counter-examples are real: `${{ github.ref && true }}` and `${{ github.ref != 'refs/heads/main' || true }}` each mention `github.ref`, read as branch-conditional, and evaluate **true** on `refs/heads/main`. The token check could never have caught them, because what matters is the value the expression takes on `main` — not which identifiers appear in it.

Of your two suggested remedies I took the **allowlist**, not the evaluator:

```go
const approvedCancelExpression = "${{ github.ref != 'refs/heads/main' }}"

func mayCancelDefaultBranchRuns(value any) bool {
    if value == nil { return false }
    switch strings.TrimSpace(fmt.Sprintf("%v", value)) {
    case "false", approvedCancelExpression: return false
    default: return true
    }
}
```

Evaluating GitHub expressions in a test would mean shipping an interpreter that itself needs testing and can drift from GitHub's semantics — a second thing to get wrong. Failing closed makes any new form a deliberate addition whose behaviour on `main` is reasoned about once, which is the property actually wanted here. The cost is that a legitimate variant (say `format('refs/heads/{0}', 'main')`) is rejected until someone adds it; that is the correct direction to fail.

Regression cases added for both of your forms, plus `${{ true }}`, quoted `"true"`, literal `true`, absent, explicit `false`, and an unreviewed-but-plausible variant. Ablated against the real workflow too: setting `ci.yaml` to `${{ github.ref && true }}` now fails the repo-wide test.

`go test ./internal/ciharness/` green, `golangci-lint` 0 issues.

devantler:



Re-requesting at `e4a448ad25c5`, which carries the fail-closed rewrite for your third finding.

`mayCancelDefaultBranchRuns` no longer inspects tokens. It accepts only an absent value, an explicit `false`, or the exact `approvedCancelExpression`, and treats everything else as cancelling — including forms that are probably fine, so adding one is a deliberate act. Both of your counter-examples (`${{ github.ref && true }}`, `${{ github.ref != 'refs/heads/main' || true }}`) are now regression cases, alongside `${{ true }}`, quoted `"true"`, literal `true`, absent, explicit `false`, and an unreviewed-but-plausible `format()` variant.

I took the allowlist rather than the evaluator deliberately: evaluating GitHub expressions in a test means shipping an interpreter that itself needs testing and can drift from GitHub's semantics.

State at this head: 33 passing, 0 failing, 0 unresolved threads. `🧪 Test`, `🧹 Lint - golangci-lint`, `🏗️ Build` and `🧹 Lint - mega-linter` are all green; the 3 outstanding jobs are System Test matrix entries, which a workflow-level concurrency expression cannot affect.

@coderabbitai review

devantler:



Advancing to the next lane. CodeRabbit answered this head at 03:46:56Z with a substantive verdict — *"No additional findings. The fail-closed allowlist closes the expression bypass. The workflow keeps a run-unique group for `main`. The repository-wide test covers the five workflows that explicitly push to `main`."* — but that arrived as a **command reply**, and six minutes later there is still no review object at this head, no post-request `summarize` comment, and the head's CodeRabbit status is unchanged since 03:30:57Z (before the request), still reading `Review skipped: automatic reviews are disabled`.

On the four reviews it completed normally this hour, the qualifying artifact followed the reply within about two minutes. A reply on its own is explicitly not a green, so this is recorded as `cr:no-gate@e4a448ad25c5` rather than treated as one — the verdict is encouraging, but it is not the gate.

@codex review

devantler:



Codex declined at 03:52:58Z: *"You have reached your Codex usage limits for code reviews."* Quota state with no retry window, so the lane cannot deliver at this head. Recorded as `codex:no-gate@e4a448ad25c5`.

This is direct evidence for this pull request, not inherited from the same message seen on devantler-tech/agent-skills#85 half an hour ago.

Lane state at `e4a448ad25c5`: CodeRabbit `no-gate` (verdict reply, no qualifying artifact), Codex `no-gate` (usage limit). Requesting Cursor Bugbot next — its trigger must be a comment containing nothing else, so it follows immediately below.

devantler:
@cursor review

devantler:



New head `d1e5a3bbdc7b`, so the ladder restarts here and the prior lane results are stale.

This commit came from my own review of the diff rather than from a reported finding: `TestCIWorkflowKeepsDefaultBranchRunsAlive` was still matching `cancel-in-progress` by substring, which is misleading for exactly the reason you identified in the repo-wide test — `${{ github.ref != 'refs/heads/main' || true }}` contains `github.ref`, `refs/heads/main` and `!=` while still cancelling on `main`. Both tests now compare against `approvedCancelExpression`. Ablated: applying that disjunction to `ci.yaml` fails the per-workflow test.

For context on the lane state at the previous head `e4a448ad25c5`: you returned "No additional findings" as a command reply with no qualifying artifact (recorded `cr:no-gate`), Codex returned its usage limit at 03:52:58Z, and Bugbot returned `neutral`/`Error` — "usage limit reached" — at 03:53:46Z. Both metered lanes are exhausted account-wide, so you are currently the only serving reviewer.

@coderabbitai review

devantler:

Both valid, both fixed in `7cbf28f8` — and the second one was not theoretical, which is worth recording.

**1. Group matched by substring.** Confirmed: `${{ github.ref == 'refs/heads/main' && github.run_id || github.run_id }}` satisfies a `Contains` check on the true-branch while its *fallback* hands pull-request and merge-queue runs unique groups — so superseded runs there stop cancelling entirely, quietly breaking the one behaviour this PR promised to preserve. The group is now compared against `approvedGroupExpression`. Ablated with exactly your counter-example: it fails with *"concurrency group must be exactly the approved expression"*.

**2. Selection missed triggers that reach `main`.** Also confirmed, and it was hiding a real workflow in this repository. `runsOnDefaultBranch` now handles an unfiltered `push:`, `branches-ignore`, and glob patterns, with a table test covering literal, unfiltered, tag-only, other-branch, `*`, `**`, `ma*n`, `branches-ignore: [develop]`, `branches-ignore: [main]` and `branches-ignore: ['mai?']`.

**The concrete consequence:** the checked set went from 5 workflows to 6. `copilot-setup-steps.yml` declares `push:` with no branch filter, so it runs on every branch including `main` — and the literal-name check skipped it silently, exactly as you described. It happens to carry no `concurrency` block, so it passes today; the point is that it was never being looked at.

I found this while re-running the guard-the-guard ablation: after neutering all five literal-`main` workflows the counter read `matched 1` rather than `0`, which is what surfaced the sixth.

Also worth noting what I did *not* do: I had spotted the unfiltered-push gap myself earlier in this PR and chose to document it as a limitation rather than close it. That was the wrong call and your finding is the correction.

`go test ./internal/ciharness/` green, `golangci-lint` 0 issues. Both threads resolved; re-requesting review at the new head once its checks land.

devantler:



New head `ea76b9ae38d9`. Both commits since your last review came from my own review of the diff rather than from a reported finding:

1. `matchesDefaultBranch` silently skipped any pattern `filepath.Match` rejected, so a pattern GitHub accepts and Go does not would drop the workflow from the checked set — the same silent hole the matcher exists to close, and directly contrary to the "fails open toward inclusion" contract in its own doc comment.
2. Fixing that exposed an inverted-sense bug: `branches-ignore` negates the result, so failing open there would mark the workflow *excluded* and skip it. The safe answer differs by caller — an inclusion list assumes an unreadable pattern matches, an exclusion list assumes it does not — so either way the workflow stays in the checked set. Both directions now have regression cases.

`TestRunsOnDefaultBranch` covers 13 cases; `go test ./internal/ciharness/` green, `golangci-lint` 0 issues.

@coderabbitai review

devantler:

**Differential finding on `runsOnDefaultBranch` — the matcher fails *closed* on two of GitHub's filter metacharacters, against its own stated contract.**

I built the same two fixes independently on this branch and lost the push race, so I ran my ablations against the merged-in version instead. Both of CodeRabbit's findings are genuinely closed by it — verified, not assumed:

- moved the group's fallback operand to `github.run_id` → `TestCIWorkflowKeepsDefaultBranchRunsAlive` fails naming the value (and both former substring checks pass on it, which is the gap that was closed);
- planted `cancel-in-progress: true` in `copilot-setup-steps.yml` → `TestNoDefaultBranchWorkflowCancelsRunsInProgress` fails naming that file, where the former literal-`main` selector never looked at it.

The one place my version and this one disagree is worth a line. `matchesDefaultBranch` delegates to `filepath.Match`, but GitHub's filter-pattern syntax is not glob — per its filter pattern cheat sheet, `+` means "one or more of the preceding character" and `?` means "zero or one of the preceding character", where Go's `?` is "exactly one character" and `+` is a literal. Neither produces `ErrBadPattern`, so `whenUnparseable` never fires and the pattern is silently judged a non-match.

Probed both matchers over patterns GitHub *does* match against `main`:

| pattern | GitHub | `runsOnDefaultBranch` |
|---|---|---|
| `main`, `ma*`, `*`, `**`, `m*n`, `mai[n]` | matches | ✅ included |
| `ma?in`, `maai?n` | matches | ❌ skipped |
| `mai+n`, `m+ain` | matches | ❌ skipped |

So a workflow filtered on `mai+n` runs on `main` and is silently dropped from the checked set — the same silent hole as the paths-only case, reached through a different door. The doc comment says the function "FAILS OPEN toward inclusion"; for these two characters it does not.

Contrived branch names, so this is low severity and not a merge blocker. But the fix is one line — treat a pattern containing `+` or `?` as undecidable and let `whenUnparseable` decide, exactly as a malformed pattern already does:

```go
if pattern == "main" || strings.ContainsAny(pattern, "*?+") {
    return true // for the inclusion arm; the ignore arm keeps its inverted sense
}
```

That also restores the symmetry the two call sites rely on: `whenUnparseable` is what makes the exclusion arm safe, and right now `+`/`?` bypass it on both arms.

Reasoned from a probe run against this branch's code, not from the diff.

devantler (commented):

## Self-review (fallback — CodeRabbit, Codex and Cursor Bugbot unavailable)

**Reviewed commit:** `ea76b9ae38d9d364cd307982215ab1cedc11d6d4`

Per-lane state at this head, each tried in priority order with evidence from this pull request:

- **CodeRabbit** — requested 04:09Z. Answered 04:09:52Z with a substantive verdict ("No additional findings", plus `[approve]`), but as a **command reply**. Thirteen minutes later: 0 review objects at this head, no post-request `summarize` comment, and the head's status unchanged since 04:09:08Z, still the not-run default `Review skipped: automatic reviews are disabled`. Its four completed reviews this hour produced the qualifying artifact within ~2 minutes. Recorded `cr:no-gate`.
- **Codex** — requested 03:5xZ on this PR. Declined 03:52:58Z: *"You have reached your Codex usage limits for code reviews."* Recorded `codex:no-gate`.
- **Cursor Bugbot** — requested 03:5xZ on this PR with a bare trigger. Check-run `Cursor Bugbot` returned `conclusion: neutral` with `output.title: Error` and summary *"Bugbot run failed"*; `cursor[bot]` commented *"Bugbot couldn't run - usage limit reached"* at 03:53:46Z. Run-never-started shape, not findings.

### Review

**Workflow change.** On `main` the group resolves to `ci-ksail-<workflow>-<run_id>`, unique per run, so nothing there queues behind or evicts anything else — which is the property `cancel-in-progress: false` alone cannot provide, since a group holds one running plus one pending run and a third arrival cancels the pending one. Off `main` the group is ref-keyed and cancellation stays enabled, so superseded pull-request runs still cancel; that half was observed live on this PR (push `33869dbf` cancelled, `c523310967b3` proceeded). `github.run_id` is always a nonzero integer, so the `||` fallback is unreachable when the condition holds. `merge_group` entries carry their own ref and are unaffected — and this repository has no merge queue configured (rulesets carry `pull_request`, `non_fast_forward`, `code_quality`, `code_scanning`; classic protection 404s).

**Test harness.** Both the group and the cancellation are compared against exact approved expressions rather than matched by substring. That is deliberate and is the whole lesson of this PR's review history: `github.ref != 'refs/heads/main' && github.run_id || github.ref` inverts the condition, `... && github.run_id || github.run_id` breaks pull-request cancellation via the fallback, and `${{ github.ref && true }}` / `${{ github.ref != 'refs/heads/main' || true }}` both evaluate true on `main` — every one of them satisfies a token check. Evaluating GitHub expressions in a test would mean shipping an interpreter that itself needs testing, so an allowlist is the honest guard.

`runsOnDefaultBranch` selects workflows by what their triggers admit rather than by a literal branch name, covering an unfiltered `push:`, `branches-ignore`, and glob patterns. It fails open toward inclusion, and the unknown-pattern answer is parameterised per caller because `branches-ignore` negates the result — failing open on both sides would skip the workflow. Twenty-two table cases across the two helpers pin these.

**Verdict: no P0/P1 findings.**

Two nits, recorded rather than fixed, as appropriate for a repository-local guard: the default branch name `main` and the `ci-ksail-` group prefix are hardcoded, so a rename would need this file updated — which the exact-match assertions make loud rather than silent.

### Limitation carried forward

The default-branch behaviour — two closely-spaced merges to `main` both retaining a completed `🧪 Test` — is observable only after this lands and is not claimed here. CI is still settling at this head (32 passing, 0 failing, 4 System Test entries outstanding); promotion waits on that.

devantler:

## Readiness

**Programmatically tested.** CI fully settled at `ea76b9ae38d9`: **70 checks, 0 failing** — including `CI - Required Checks`, the full 27-entry System Test matrix, `🧪 Test`, `🧹 Lint - golangci-lint`, `🏗️ Build` and `🧹 Lint - mega-linter` (actionlint over the changed workflow). `mergeStateStatus: CLEAN`, 0 unresolved threads, 0 unverified commits. The fix carries a RED/GREEN proof and eight ablations, each one moving a mechanism and naming the assertion expected to fire.

**Reviewed.** All three lanes were tried at this head with per-PR evidence and none delivered: CodeRabbit answered with a clean verdict as a command reply but produced no review object, no `summarize` comment and no status change in 13 minutes (its four completed reviews this hour produced the artifact in ~2); Codex returned its usage limit at 03:52:58Z; Bugbot returned `neutral`/`Error` — "usage limit reached" — at 03:53:46Z. The qualifying **local review round** is posted above as a real GitHub Review at exactly this SHA, clean: **no P0/P1 findings**. Recorded as `green_review=self@ea76b9ae38d9`.

**Tried and evaluated as a user.** The pull-request half was observed live on this PR rather than reasoned: two pushes 73 seconds apart produced `cancelled` for the superseded run `33869dbf` and a live run for `c523310967b3`, confirming GitHub evaluates the ref-scoped concurrency expression and that superseded PR runs still cancel. The identical pattern is merged and behaviourally confirmed in three sibling repos — on devantler-tech/platform#2992 the gated workflow ran on the merge commit and completed successfully under the new key.

**What is not claimed.** The default-branch half — two closely-spaced merges to `main` both retaining a completed `🧪 Test` — is observable only after this lands. #6384 goes to 📊 Verifying, not Done, until that is seen.

Promoting on that basis.

Issue #6563, opened 2026-08-14
Go validation is skipped on every pull request that changes Go code


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`ci-go` ("✅ Validate Go Project") gates its pull-request arm on the **wrong** filter output:

```yaml
  ci-go:
    name: ✅ Validate Go Project
    if: >-
      (github.event_name == 'push' && github.ref == 'refs/heads/main')
      || (github.event_name == 'pull_request'
      && github.event.pull_request.head.repo.full_name == github.repository
      && needs.changes.outputs.govuln-allowlist == 'true')
```

`govuln-allowlist` matches exactly one file:

```yaml
            govuln-allowlist:
              - '.govulncheck-allow.txt'
```

So on a pull request the Go suite runs **only when that single allowlist file changes**. The filter
that actually describes "this diff touched Go" is `code` (`**/*.go`, `go.mod`, `go.sum`,
`.govulncheck-allow.txt`, `.github/actions/**`) — it is computed and exported by the `changes` job,
and then never referenced here.

Measured on merged pull requests, neither of which touched the allowlist:

| PR | Go files changed | `✅ Validate Go Project` |
|---|---|---|
| #6550 | 3 | **SKIPPED** |
| #6528 | 2 | **SKIPPED** |

## Impact

Go tests, `golangci-lint`, dead-code analysis and the vulnerability scan do not run on pull requests
that change Go code. The suite still runs on push to `main` — verified on the 2026-08-13T18:15Z run,
where `🧪 Test`, `🛡️ Vulnerability Scan` and `🧹 Lint - mega-linter` all completed `success`.

So the gate is not missing, it is **misplaced**: a Go regression is caught *after* it merges, as a red
`main`, instead of blocking the pull request that introduced it. A PR can be fully green while nothing
has compiled or executed its tests.

This also silently weakens review: #6258 adds a Go regression test and its author recorded that their
local run timed out without producing results. That test has therefore never executed anywhere, on a
pull request that otherwise reads as green and reviewed.

## Expected behaviour

A pull request that changes Go code runs the Go validation suite before it can merge.

## Smallest useful change

Gate the pull-request arm on `needs.changes.outputs.code`, which already covers the allowlist file:

```yaml
      && needs.changes.outputs.code == 'true')
```

## Acceptance criteria

- [ ] A pull request changing a `.go` file runs `✅ Validate Go Project` (positive control).
[...]

PR #6569, opened 2026-08-14
chore(ci): say why ci-go's pull-request gate is deliberately narrow

### LINKED ISSUE #6568: chore(ci): say why ci-go's pull-request gate is deliberately narrow

### Evidence

Three consecutive engineering runs read `ci-go`'s pull-request gate as a bug, and one of them opened
#6563 and #6564 to "fix" it. Both are now closed: the gate is deliberate, and widening it would have
added nine duplicate jobs to every Go pull request.

The reason the misreading keeps happening is that nothing at the gate says what it is for. A reader
sees a full Go suite gated on a single allowlist filename and reasonably concludes the filter is
wrong. Recovering the truth requires knowing that
`devantler-tech/actions/.github/workflows/validate-go-project.yaml` carries its own `pull_request:`
trigger and runs org-required on every pull request — which is not visible from this repository at
all.

`TestGoValidationIncludesVulnerabilityAllowlistChanges` already pins the gate correctly. It was read
as pinning a stale contract and was edited to match the wrong change, because it too states only what
it asserts and not why.

### Problem and audience

The audience is the next person or agent to touch `ci.yaml` — the cost so far is three runs of
forensics, two closed artifacts, and one nearly-merged regression.

### Expected behaviour

The gate explains itself where it is read, and the contract test says why it exists, so the next
reader is told the rationale instead of rediscovering it.

### Acceptance criteria

- The `ci-go` `if:` block carries a comment stating that the pull-request arm is narrow **on purpose**:
  the org-required direct run of the same reusable workflow already covers every Go pull request, and
  this arm exists only to give an allowlist-only pull request the `scan-default-branch: true` scan
  that the required run (which takes the input's `false` default) does not perform.
- The comment states that the `push`-to-`main` arm is the only Go validation `main` gets, because the
  required workflow is pull-request scoped.
- `TestGoValidationIncludesVulnerabilityAllowlistChanges` carries a doc comment explaining that
  `govuln-allowlist` is the intended gate and that widening it to `code` duplicates the required
  workflow — so a future reader does not "correct" the assertion.
- Behaviour-preserving: no workflow logic changes.

Rough size: small.

devantler:

### Why

`ci-go`'s pull-request gate has been read as a bug by three engineering runs in a row. Two of them
opened work to "fix" it — #6563 and #6564, both now closed — and one came close to merging a change
that would have doubled the Go suite on every Go pull request while adding no coverage at all.

The gate is correct. It just does not say so anywhere, and the fact that makes it correct lives in
another repository, so each reader re-derives it from scratch and gets it wrong.

### What

Records the rationale at the two places it is actually read: the workflow gate itself, and the test
that pins it. No workflow logic changes — comments only.

Fixes #6568

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -356,6 +356,26 @@ jobs:
   ci-go:
     name: ✅ Validate Go Project
     needs: [changes]
+    # The two arms below cover what the ORG-REQUIRED workflow does not, and nothing
+    # more. validate-go-project.yaml carries its own `pull_request:` / `merge_group:`
+    # triggers alongside its `workflow_call` interface, so GitHub already runs the
+    # complete Go suite directly on every pull request here, with no help from this
+    # file. That direct run publishes its checks WITHOUT the `✅ Validate Go Project /`
+    # prefix, which is why a prefix-filtered read of a pull request shows "skipped"
+    # over a suite that has in fact passed (ksail#6563, ksail#6564 — both closed as
+    # measurement errors of exactly that shape).
+    #
+    #   * push to main — the required workflow is pull-request scoped and never runs
+    #     on the default branch, so this arm is the ONLY Go validation main gets.
+    #
+    #   * pull_request, gated on govuln-allowlist — DELIBERATELY NARROW. The required
+    #     run takes `scan-default-branch`'s default of false, so it will not scan a
+    #     pull request that edits only a .govulncheck-allow.txt; this arm exists so a
+    #     risk acceptance is validated by the scanner it applies to. Do NOT widen this
+    #     to `code`: .govulncheck-allow.txt is already a member of the `code` filter,
+    #     so `code` is a strict superset whose extra region is precisely where the
+    #     required workflow already runs the identical nine jobs. Widening it buys no
+    #     coverage and doubles the Go suite on every Go pull request.
     if: >-
       (github.event_name == 'push' && github.ref == 'refs/heads/main')
       || (github.event_name == 'pull_request'
--- a/internal/ciharness/system_test_harness_test.go
+++ b/internal/ciharness/system_test_harness_test.go
@@ -88,6 +88,21 @@ type ciWorkflow struct {
 	} `yaml:"jobs"`
 }
 
+// TestGoValidationIncludesVulnerabilityAllowlistChanges pins ci-go's pull-request gate to
+// govuln-allowlist ON PURPOSE. The assertion below looks like it pins a stale contract, and
+// twice it has been "corrected" to needs.changes.outputs.code (ksail#6563, ksail#6564 — both
+// closed once measured). It is not stale:
+//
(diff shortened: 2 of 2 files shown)


devantler:

## Verification record

### The guard actually fires — RED/GREEN on the contract test

The doc comment alone would be inert if the assertion beneath it did not catch the mutation it warns
about. Ablated by making exactly the edit #6564 made, and nothing else:

| Arm | `TestGoValidationIncludesVulnerabilityAllowlistChanges` |
|---|---|
| Gate widened to `needs.changes.outputs.code == 'true'` | **FAIL** — `"…outputs.code == 'true')" does not contain "needs.changes.outputs.govuln-allowlist == 'true'"` |
| Gate as shipped | **ok** |

So the test does refuse the precise change that was twice attempted. `gofmt` clean; diff is
comment-only, `+35 / -0`.

### Evaluated as a user — this PR demonstrates its own claim

The comment asserts that the org-required workflow runs the Go suite on every pull request while
`ci-go` correctly stays out of the way on a non-allowlist diff. This PR is exactly that shape — it
changes a `.go` file and no `.govulncheck-allow.txt` — so its own check list is the observation:

- **Bare-named Go suite present and running**: `🧪 Test`, `🛡️ Vulnerability Scan`, `🏗️ Build`,
  `📊 Code Coverage`, `🔍 Dead Code Analysis`, `🧹 Lint - golangci-lint` — the org-required direct run.
- **Prefixed `✅ Validate Go Project / …` checks: 0** — `ci-go` did not fire, exactly as the narrow
  gate intends.

That is the whole claim, observable on this pull request: the Go suite runs, and it does not run
twice. Had the gate been widened, both sets would be present — which is the regression the comment
and the test now guard against.

CI is still settling; review will be requested once it is green.

devantler:



@coderabbitai review

devantler:

Readiness at `85aa3ca97a6d5354f60f3171a277e5fa3571f8a3`:

- **Programmatically tested** — 65 checks success, 5 skipped, none failing; `mergeStateStatus=CLEAN`.
  That includes the Go suite and `TestGoValidationIncludesVulnerabilityAllowlistChanges`, the test
  this annotates.
- **Reviewed** — CodeRabbit green at this head: its summary names the full head and reports no
  actionable comments; zero unresolved threads and no finding sections.
- **Tried and evaluated as a user** — this change has **no exercisable runtime surface**: the diff is
  +35 lines, all of them comments, in `ci.yaml` and one Go test file, with nothing removed and no
  expression, gate or assertion altered. There is no behaviour to observe, so it was evaluated
  statically: the `if:` condition and the pinned assertion are byte-identical before and after, and
  the text was checked against the measured behaviour it records.

Its value is precisely that measurement error: the same gate has been "corrected" twice
(ksail#6563, #6564, both closed once measured) because the org-required run publishes its checks
without the `✅ Validate Go Project /` prefix, so a prefix-filtered read shows "skipped" over a
passing suite. This records that where the next reader will look.

Issue #6574, opened 2026-08-14
refactor(webui): converge the two loopback-origin helpers into one


devantler:
> 🤖 Generated by the Agentic Engineer

## Why

Two pull requests fixed the same DNS-rebinding class in `pkg/webui/api` — #6255 (pod exec WebSocket upgrade) and #6251 (plugin install). Because they are on separate branches, each carries its own loopback-origin helper, deliberately named differently (`isLoopbackHost` and `isLoopbackOriginHost`) so the two branches cannot collide on merge with a duplicate definition.

Once both have landed the package will hold two near-identical helpers. That is a correctness risk rather than untidiness: a future change to what counts as a trusted origin could be applied to one and missed in the other, silently leaving one of the two endpoints on the weaker rule.

## What

After both PRs merge, collapse them into a single shared origin helper and have the exec upgrader and the plugin install guard call it. Behaviour-preserving; existing regression tests on both paths must pass unmodified.

Blocked until #6255 and #6251 are both merged.

devantler:
> 🤖 Generated by the Agentic Engineer

Another reason to converge, found the hard way: the two copies of this origin logic have now **diverged in a security-relevant way**. CodeRabbit caught the `wails://` carve-out trusting the scheme alone (so `wails://attacker` passes) on #6251, where it is fixed in `c7aa7e8d`; the identical pattern sits in `exec_handler.go`, which merged to `main` in #6255 before that review landed → filed as #6576.

That is the whole argument for this issue: one review found one copy, and the other copy shipped. Whichever converged helper this produces should be the single place the wails-origin and loopback rules are expressed.

PR #6577, opened 2026-08-14
fix(webui): trust only the desktop origin on the exec upgrade

### LINKED ISSUE #6576: fix(webui): exec origin check trusts any wails:// authority, not just the desktop origin

## Evidence

`pkg/webui/api/exec_handler.go` trusts the desktop shell by **scheme alone**:

```go
// The desktop shell serves the SPA from the fixed wails:// origin through the Wails AssetServer, with
// no TCP listener at all. A remote page can never carry that scheme, so it is trusted like loopback.
if strings.EqualFold(parsed.Scheme, "wails") {
    return true
}
```

That admits **any** authority under the scheme — `wails://attacker` passes as readily as
`wails://wails` — and on this path the scheme check is the only thing between the caller and a shell
in the cluster, since `execCheckOrigin` returns `true` outright.

The desktop shell serves exactly one origin. `desktop/main.go` documents it: the SPA, REST endpoints
and SSE stream are all served *"same-origin (at the `wails://wails` production origin)"*. So the
carve-out only ever needed that one value.

## How this was found, and why it is filed separately

CodeRabbit raised it on #6251 against the **plugin install** guard, which carries the same scheme-only
check. It is fixed there in `c7aa7e8d`. The identical pattern is in the **exec** handler, which merged
to `main` in #6255 shortly before that review landed — I merged it, and the review that would have
caught it arrived on the sibling PR afterwards. Filing rather than widening #6251, since this is a
different file already on `main`.

## Smallest useful change

Mirror the #6251 fix: match scheme **and** host exactly, and reject user info, path, query and
fragment. Compare `URL.Host` rather than `Hostname()`, so a port fails the comparison instead of being
stripped away before it.

## Acceptance criteria

- `Origin: wails://wails` still upgrades (the desktop path is pinned by a test).
- `wails://attacker`, `wails://wails:8080`, `wails://user@wails` and `wails://wails/path` are all
  rejected, each with a test asserting the upgrade does not happen.
- The ablation is checked: restoring the scheme-only form must fail every reject case.
- Both handlers end up sharing one helper rather than two copies — this is the second instance of the
  same reasoning diverging, so #6574 (converge the two loopback helpers) should absorb it.

## Severity

Reachable only from inside a Wails webview, so this is hardening rather than a live remote hole — but
the exec endpoint is unauthenticated and grants a shell, so the exact-origin form is what the comment
already claims the code does.

devantler:

### Motivation

The web UI's terminal endpoint hands the caller a shell inside the cluster, so it only accepts
connections from a page the user actually opened locally. It also has to allow the native desktop app,
which serves its pages from a fixed internal address rather than from the local server.

That desktop exception was too wide: it accepted anything using the desktop's address *scheme*, rather
than the one address the desktop app actually uses. A page that could present any other address under
that scheme would have been let through.

### What this changes

The exception now matches the desktop app's single real address exactly, and rejects anything with
extra detail attached to it. The desktop app keeps working unchanged, and the everyday local-browser
path is untouched.

Reachable only from inside the desktop app's own web view, so this is hardening rather than a live
remote hole — but it is the endpoint least worth leaving loose.

Fixes #6576

### DIFF
--- a/pkg/webui/api/exec_handler.go
+++ b/pkg/webui/api/exec_handler.go
@@ -136,10 +136,19 @@ func execCheckOrigin(request *http.Request) bool {
 		return false
 	}
 
-	// The desktop shell serves the SPA from the fixed wails:// origin through the Wails AssetServer, with
-	// no TCP listener at all. A remote page can never carry that scheme, so it is trusted like loopback.
+	// The desktop shell serves the SPA from the fixed wails://wails origin (see desktop/main.go) through
+	// the Wails AssetServer, with no TCP listener at all, so that one origin is trusted like loopback.
+	// The scheme alone is not enough: it would also admit wails://attacker, and on this path the check is
+	// the only thing between the caller and a shell in the cluster. A browser-sent Origin carries scheme
+	// and authority and nothing else, so a non-empty user, path, query or fragment marks a crafted header.
 	if strings.EqualFold(parsed.Scheme, "wails") {
-		return true
+		// Host, not Hostname, so a port (wails://wails:8080) fails the comparison rather than being
+		// stripped away before it.
+		return strings.EqualFold(parsed.Host, "wails") &&
+			parsed.User == nil &&
+			parsed.Path == "" &&
+			parsed.RawQuery == "" &&
+			parsed.Fragment == ""
 	}
 
 	return isLoopbackHost(parsed.Hostname())
--- a/pkg/webui/api/exec_handler.go
+++ b/pkg/webui/api/exec_handler.go
@@ -139,16 +139,14 @@ func execCheckOrigin(request *http.Request) bool {
 	// The desktop shell serves the SPA from the fixed wails://wails origin (see desktop/main.go) through
 	// the Wails AssetServer, with no TCP listener at all, so that one origin is trusted like loopback.
 	// The scheme alone is not enough: it would also admit wails://attacker, and on this path the check is
-	// the only thing between the caller and a shell in the cluster. A browser-sent Origin carries scheme
-	// and authority and nothing else, so a non-empty user, path, query or fragment marks a crafted header.
+	// the only thing between the caller and a shell in the cluster.
 	if strings.EqualFold(parsed.Scheme, "wails") {
-		// Host, not Hostname, so a port (wails://wails:8080) fails the comparison rather than being
-		// stripped away before it.
-		return strings.EqualFold(parsed.Host, "wails") &&
-			parsed.User == nil &&
-			parsed.Path == "" &&
-			parsed.RawQuery == "" &&
-			parsed.Fragment == ""
+		// Compare the raw header, not the parsed fields. A browser-sent Origin is scheme and authority
+		// and nothing else, and an empty delimiter parses away to nothing: "wails://wails?" and
+		// "wails://wails#" both leave RawQuery and Fragment empty, so a field-by-field comparison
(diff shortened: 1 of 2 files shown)


devantler:


@coderabbitai review

devantler:



@coderabbitai review

devantler:

Resolution record for the non-thread finding on the CodeRabbit review at
`55be549a481478b4eaaf094a38cd341731dd20a9`.

**The finding** (pre-merge *Linked Issues check*, ❌ Error):

> The PR enforces the exact desktop origin and adds rejection tests, but it does not show shared
> validation logic with the plugin-install handler. Resolution: Share the origin-validation logic
> between the plugin-install and exec handlers as required by issue `#6576`.

**Refuted — the linked issue requires the opposite, and sharing now would break `main`.**

1. **#6576 prescribes mirroring, not sharing, and says so explicitly.** Its *Smallest useful change*
   is "Mirror the #6251 fix: match scheme **and** host exactly", and it records the reason it was
   filed at all: *"Filing rather than widening #6251, since this is a different file already on
   `main`."* The check reads "as required by issue #6576" for a requirement the issue deliberately
   declined to make.

2. **The shared symbol would collide with an open PR in another lane.** #6251
   (`codex/fix-plugin-install-trust-gate-vulnerability`, still an open draft) carries the sibling fix
   in the same `package api`. Extracting a shared origin helper here means both PRs declare it, and
   whichever merges second breaks `main` — a collision **neither PR's CI can see**, because each
   builds only against `main` plus its own diff.

3. **The convergence is already tracked separately, as #6574** —
   *"refactor(webui): converge the two loopback-origin helpers into one"*. That is the correct home
   for it: it can be done once against a settled `main`, after #6251 lands, as a behaviour-preserving
   refactor rather than as a rider on a security fix.

No files changed, so this restarts at the same head. The review itself carried **no actionable
comments** and the PR has **0 review threads**; this was CodeRabbit's ancillary pre-merge evaluator,
which is assessed on merit rather than treated as a separate gate.

devantler:

**Readiness record — user evaluation at `55be549a481478b4eaaf094a38cd341731dd20a9`**

Exercised the real behaviour rather than reasoning about it, on a worktree detached at exactly this
head (verified `HEAD == headRefOid`, clean tree before and after).

**Observed.** `go test ./pkg/webui/api/ -run TestExec` — all 7 tests pass, including
`TestExecAllowsDesktopWailsOrigin` (the desktop app's real origin is still accepted, so this does not
lock the desktop shell out) and all 9 rejection subtests.

**Ablation — the assertions discriminate.** Reverting only the origin comparison back to the
field-by-field form and re-running produced **exactly** the predicted movement:

| Subtest | Ablated result |
|---|---|
| `wails://wails?` | **FAIL** |
| `wails://wails#` | **FAIL** |
| `wails://wails?#` | **FAIL** |
| `wails://attacker`, `wails://wails:8080`, `wails://user@wails`, `wails://wails/path`, `wails://wails?x=1`, `wails://wails#frag` | pass |

Three failures, exactly the three new cases, no others — so the fix is load-bearing and the new tests
are not vacuous. Restored afterwards and re-ran: green, tree clean, `HEAD` unchanged.

**Readiness:** programmatically tested (CI 46/46 green + the above), reviewed (CodeRabbit at this
head, no actionable comments, 0 threads; its ancillary pre-merge check refuted in the preceding
comment), and evaluated as a user. Promoting.

PR #6605, opened 2026-08-19
fix(ci): address the container image user by its numeric id

devantler:

### Why

Every open pull request on this repository is currently failing its lint check, including the dependency bots' own. The cause is not in any of those changes — a code-quality rule now objects to how the container image names the user it runs as, and that check only runs on pull requests, so `main` stays green while nothing can get through.

### What

Refers to the same user by its number instead of its name. The base image already defines that user with this exact number, so the image runs as it did before.

It is also the form Kubernetes can actually act on: its "must not run as root" safety check can only verify a user given by number, and silently cannot when given a name.

**Security floor:** the one place that safety check could not be enforced now can be.
**Everyday path:** unchanged — no new step for anyone building or running the image.

Unblocks the pull-request queue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,9 @@ ARG TARGETPLATFORM
 # GoReleaser v2 places binaries in ${TARGETPLATFORM}/ subdirectories
 COPY ${TARGETPLATFORM}/ksail /ksail
 
-# Use nonroot user from distroless
-USER nonroot:nonroot
+# Use the distroless nonroot user by its numeric id. A name is not resolvable
+# without /etc/passwd, and Kubernetes runAsNonRoot can only verify a numeric id.
+USER 65532:65532
 
 # Add a simple healthcheck compatible with distroless (exec form only)
 # This verifies the binary is present and runnable.


devantler:



@coderabbitai review

devantler:

## Readiness at `9bffffc7b17b340b3c74dcbb1078f892aa758934`

**Programmatically tested.** RED/GREEN against the real linter, not a reasoned argument:
`docker run --rm -i hadolint/hadolint:latest < Dockerfile` reports `DL3066` at line 15 with exit 1 on
`origin/main`'s file, and exits 0 with no findings on this one.

**Equivalence checked rather than assumed.** The base image's own config sets `User: 65532`, read via
`docker buildx imagetools inspect --format '{{json .Image}}'` (config blob only, no pull), so the
numeric form addresses the same identity the name did.

**Tried and evaluated as a user.** The observable effect is on the real pipeline: `🧹 Lint - mega-linter`
was `FAILURE` on every open pull request in this repository — #6589, #6603, #6604 — and is `SUCCESS`
on this one. That is the behaviour this change exists to produce, observed rather than inferred.

**Pentad:** 16 checks SUCCESS / 0 FAILURE / 0 incomplete · 0 unresolved threads · no actionable review
findings (the sole collapsed section is the informational `🔇 Additional comments (1)`, `Dockerfile
14-16: LGTM!`) · `mergeStateStatus: CLEAN` · green review at this head.

Scope check: the diff is one line plus a comment. It does not change what the image runs as.

Issue #2261, opened 2026-02-14
Remove VCluster D-Bus workaround after upstream fix


devantler:
## Context

PR #2241 adds VCluster × Docker (Vind) support. During CI testing on GitHub Actions Ubuntu runners, the VCluster SDK's `install-standalone.sh` fails with:

```
failed to start vCluster standalone: exit status 1: Failed to connect to bus: No such file or directory
```

**Root cause:** The SDK runs `systemctl restart systemd-journald` inside the privileged Docker container before systemd has initialized D-Bus (`/run/dbus/system_bus_socket`). The container starts with `--tmpfs /run`, so the socket doesn't exist immediately. Combined with `set -e`, this aborts the install script.

This works on macOS (Docker Desktop provides a fully-initialized systemd VM) but fails on Linux CI runners due to a timing race.

## Current workaround

KSail detects the `"Failed to connect to bus"` error from `CreateDocker`, waits for D-Bus readiness inside the container, then re-runs the install script. See `recoverFromDBusError` in `pkg/svc/provisioner/cluster/vcluster/provisioner.go`.

## Action items

- [x] Verify whether the D-Bus recovery workaround is needed after CI runs confirm the fix works
- [x] If confirmed necessary, file an upstream issue on [loft-sh/vcluster](https://github.com/loft-sh/vcluster) suggesting one of:
  - Adding a D-Bus readiness wait before `systemctl restart systemd-journald` in `install-standalone.sh`
  - Making the journald restart non-fatal (remove `set -e` or use `|| true`)
  - Adding an SDK option to skip persistent logging setup
- [ ] Remove the workaround once an upstream fix is released

## References

- `install-standalone.sh` line 333: `systemctl restart systemd-journald`
- VCluster SDK: `github.com/loft-sh/vcluster v0.32.0-rc.1`
- Chart version: `0.32.0-alpha.2`

**Blocker:** loft-sh/vcluster#3805 | last-verified 2026-08-19: **not shipped** — the upstream issue
was filed 2026-04-07 and recorded the journald/D-Bus race as still present in v0.33.1 and
v0.34.0-alpha.1. The Go module proxy reports latest stable **v0.36.1** (2026-07-29), unchanged since
the 2026-08-17 check, so nothing has been released since that verification. Whether the race is fixed
in v0.36.1 is not determinable from non-repository sources. Cheapest next step remains dropping
`recoverFromDBusError` behind a test against current vcluster and seeing whether CI still reproduces
[...]

PR #6616, opened 2026-08-21
fix(vcluster): report D-Bus recovery lifecycle

### LINKED ISSUE #6295: test(vcluster): make the D-Bus recovery path self-reporting so its removal becomes evidence-based

## Evidence

`recoverFromDBusError` (`pkg/svc/provisioner/cluster/vcluster/provisioner.go`) exists to work around a
D-Bus startup race in the vCluster SDK's standalone installer. Whether it is still needed has been
decided three times by a human re-reading upstream source — the audit comments record v0.33.2-rc3
(April 2026) and v0.34.0 (May 2026) — and the trail is already stale against the current
`v0.35.2` pin.

That audit method cannot be kept current: `install-standalone.sh` is fetched at runtime and is **not**
vendored in the module, so the upstream behaviour is invisible from the dependency we pin.

## Problem

The parent issue's closing condition — *"remove the workaround once an upstream fix is released"* —
is unobservable. The recovery path logs one `Infof` line when it fires and is otherwise silent, so
nothing surfaces whether it triggered in a given run. We cannot tell a still-needed workaround from
dead code.

## Expected behaviour

When the D-Bus recovery path fires, that fact is visible in CI output as a distinct, greppable signal
(and, symmetrically, its absence across a VCluster × Docker Linux CI run is meaningful evidence).

## Smallest useful change

Emit an explicit marker when `recoverFromDBusError` is entered and when it completes, at a level that
survives normal CI log capture — rather than only the existing mid-path `"D-Bus ready, re-running
install script..."` line. No behaviour change to the recovery itself.

## Acceptance criteria

- [ ] Entering the recovery path emits a distinct, greppable marker in CI output.
- [ ] A unit test asserts the marker is emitted on the recovery path and **not** emitted on the
      happy path (both states covered).
- [ ] No change to recovery behaviour — existing `retry_test.go` cases pass unmodified.
- [ ] The parent issue is updated to key removal on observed CI evidence instead of a manual
      upstream re-read.

Rough size: small.

Part of #2261.

devantler:

## Why

The D-Bus recovery workaround can run in CI without leaving a stable signal that says when recovery started or completed. That makes the parent removal decision depend on repeated source audits instead of observed behavior.

## What

- emit a stable Info-level marker when in-place D-Bus recovery starts
- emit a matching completion marker only after recovery succeeds
- cover the recovery and ordinary create paths through the real stream logger

This changes observability only; retry and recovery behavior are unchanged.

## Validation

- strict TDD RED: the recovery-path test failed on both missing markers while the happy path passed
- focused GREEN: 3/3 marker subtests passed
- affected package race suite: 101/101 passed
- go vet ./... passed
- go build passed
- diff-only golangci-lint reported no issues
- host-enabled go test ./... passed 5,837 tests; six unrelated open/chat child-process fixtures failed under full-suite contention, then that package passed 127/127 alone

Fixes #6295
Part of #2261

### DIFF
--- a/pkg/svc/provisioner/cluster/vcluster/provisioner.go
+++ b/pkg/svc/provisioner/cluster/vcluster/provisioner.go
@@ -363,14 +363,14 @@ func tryDBusRecovery(
 	logger loftlog.Logger,
 	recoverDBus dbusRecoverFn,
 ) error {
-	logger.Infof("ksail.vcluster.dbus_recovery state=entered cluster=%q", clusterName)
+	logger.Info("ksail.vcluster.dbus_recovery state=entered")
 
 	recoverErr := recoverDBus(ctx, globalFlags, clusterName, logger)
 	if recoverErr != nil {
 		return fmt.Errorf("D-Bus recovery failed: %w", recoverErr)
 	}
 
-	logger.Infof("ksail.vcluster.dbus_recovery state=completed cluster=%q", clusterName)
+	logger.Info("ksail.vcluster.dbus_recovery state=completed")
 
 	return nil
 }
--- a/pkg/svc/provisioner/cluster/vcluster/provisioner.go
+++ b/pkg/svc/provisioner/cluster/vcluster/provisioner.go
@@ -363,12 +363,14 @@ func tryDBusRecovery(
 	logger loftlog.Logger,
 	recoverDBus dbusRecoverFn,
 ) error {
-	logger.Infof("D-Bus not ready in container — recovering in place...")
+	logger.Infof("ksail.vcluster.dbus_recovery state=entered cluster=%q", clusterName)
 
 	recoverErr := recoverDBus(ctx, globalFlags, clusterName, logger)
 	if recoverErr != nil {
 		return fmt.Errorf("D-Bus recovery failed: %w", recoverErr)
 	}
+
+	logger.Infof("ksail.vcluster.dbus_recovery state=completed cluster=%q", clusterName)
 
 	return nil
 }
--- a/pkg/svc/provisioner/cluster/vcluster/retry_test.go
+++ b/pkg/svc/provisioner/cluster/vcluster/retry_test.go
@@ -1,6 +1,7 @@
 package vclusterprovisioner_test
 
 import (
+	"bytes"
 	"context"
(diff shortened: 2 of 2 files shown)


devantler:

## CodeQL resolution — 9c7a8ca1baef966dfb6de65257199f9bcddacec2

The first hosted round reported two medium log-injection alerts because both lifecycle markers included the user-provided cluster name. The cluster field was unnecessary for the greppable recovery signal, so this head removes user input from both marker lines instead of suppressing the scanner.

Strict regression proof: a hostile cluster value containing a newline first made the real-logger test fail because its text appeared in the marker output; after the fix, the marker test passed 3/3, the affected-package race suite passed 101/101, diff-only golangci-lint found no issues, and repository-wide vet/build passed.

devantler:


@coderabbitai review

devantler:

## CodeRabbit ancillary finding resolved — 9c7a8ca1baef966dfb6de65257199f9bcddacec2

Refuting the Docstring Coverage warning in the current-head review summary: the touched production function tryDBusRecovery already has a full Go doc comment immediately above it. The other touched functions counted by the generic 80% metric are test-only helpers and Test... entrypoints; this repository does not add API docstrings to those symbols, and doing so would add non-behavioral noise solely to satisfy the provider heuristic.

The substantive review reported no actionable comments, and no code change is warranted for this ancillary warning.

devantler:

## Exact-head behavioral evaluation — 9c7a8ca1baef966dfb6de65257199f9bcddacec2

Tried the change through the real stream logger at Info level, not only by inspecting the code:

- the D-Bus recovery path emitted exactly the stable `state=entered` and `state=completed` lifecycle markers;
- a hostile cluster name containing a newline did not appear in either marker;
- the ordinary successful-create path emitted neither recovery marker.

The first hosted round's CodeQL finding was reproduced by the hostile-input test and fixed at the root by removing user input from the markers. On this exact head, the focused marker test passes 3/3, the affected-package race suite passes 101/101, repository-wide vet/build pass, and all 70 hosted checks—including CodeQL and the Docker system-test fanout—are terminal green. Current-head CodeRabbit review reports no actionable comments.

Issue #5948, opened 2026-07-07
fix(reconcile): ArgoCD cold-start ComparisonError misclassified as permanent → flaky GitOps reconcile & E2E System-Tests


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

The E2E System-Test matrix (`ksail up` → GitOps reconcile) **flakes intermittently**, preferentially on the slower `--cni Calico` leg, with ArgoCD `ComparisonError`s during control-plane cold-start. It recurs across PRs and blocks CI portfolio-wide (own PRs *and* the dependabot cluster stall on the required System-Test gate). Observed signature (from a failing Calico leg):

```
Condition: ComparisonError - Failed to load target state: ... rpc error ...
  failed to resolve revision "dev": ... lookup talos-default-local-registry ... no such host
redis: connection pool: failed to dial after 5 attempts: ... connect: connection refused
Ignoring temporary failed attempt to compare app state against repo ... will retry
```

Crucially the **cluster bootstraps fine** — argocd-redis and argocd-repo-server reach `1/1 Running`, "No pods in error state found", the workload image is pulled. ArgoCD's *own* internal loop logs "will retry" and self-heals seconds later. The test fails only because ksail's reconcile wait has already given up.

## Root cause

ksail's ArgoCD reconciler classifies **cold-start transients as permanent, and the poll loop treats any classifier error as terminal on the first miss** — before ArgoCD self-heals:

- `pkg/client/argocd/reconciler.go:190-208` — `isSourceRelatedError()` matches `"connection refused"`, `"unable to resolve"`, `"failed to fetch"`, `"not found"`. These are exactly what a *just-starting* repo-server/redis and a not-yet-resolvable local-registry DNS emit during cold-start.
- `pkg/client/argocd/reconciler.go:176-187` — `checkConditions()` returns `ErrSourceNotAvailable` the moment an `ApplicationCR` carries a `ComparisonError`/`SyncError` whose message matches those patterns.
- `pkg/client/reconciler/poll.go:74-95` — `PollUntilReady` returns on **any** non-`Halt` check error as `"permanent failure: %w"` (terminal). So the very first poll during the cold-start window aborts the wait; there is no distinction between "the source genuinely doesn't exist" (permanent) and "the control-plane hasn't finished starting" (transient, self-healing).
[...]

PR #6621, opened 2026-08-21
fix(reconcile): keep ArgoCD cold-start transients from failing the readiness poll

devantler:

### Why

Setting up a cluster sometimes failed for no real reason. While the cluster is still starting, ArgoCD's own internal pieces are not up yet, so the very first check of the deployment reports a connection problem. ArgoCD fixes that by itself within seconds — but KSail had already declared the whole thing broken and given up.

Because it depended on timing, it failed at random. It repeatedly blocked our own changes and the automated dependency updates from getting through, which made a green build something you had to retry rather than trust.

### What

KSail now waits out that brief start-up period instead of failing on the first stumble, and only reports a failure if the source is *still* unreachable once things have had time to come up.

A genuinely wrong setting — a bad repository address, for example — still fails with the same actionable message as before. That distinction is deliberate: the goal is to stop reporting a problem that isn't there, without ever hiding one that is.

Part of #5948

### DIFF
--- a/pkg/cli/cmd/workload/reconcile.go
+++ b/pkg/cli/cmd/workload/reconcile.go
@@ -812,6 +812,10 @@ func buildArgoCDApplicationTasks(
 // pollUntilApplicationReady polls a named ArgoCD Application until it is
 // synced and healthy, or the context's deadline expires. On permanent failure,
 // it returns an actionable error including the resource name and failure details.
+// A source-availability error inside the cold-start warm-up window is treated as
+// "not ready yet" rather than a permanent failure, because ArgoCD resolves those
+// itself within seconds of the control plane coming up.
+//
 // The caller is expected to provide a context with a deadline (shared across all
 // application tasks) so that the total reconcile time is bounded.
 func pollUntilApplicationReady(
@@ -819,12 +823,18 @@ func pollUntilApplicationReady(
 	argoReconciler *argocd.Reconciler,
 	name string,
 ) error {
+	start := time.Now()
+
 	return reconcilerclient.PollUntilReady( //nolint:wrapcheck // identity preserved
 		ctx,
 		argoCDApplicationPollInterval,
 		func(ctx context.Context) (reconcilerclient.CheckResult, error) {
 			ready, err := argoReconciler.CheckNamedApplicationReady(ctx, name)
 			if err != nil {
+				if argocd.IsColdStartTransient(err, time.Since(start), argoCDSourceWarmupGrace) {
+					return reconcilerclient.CheckResult{Ready: false}, nil
+				}
+
 				return reconcilerclient.CheckResult{}, err //nolint:wrapcheck // identity preserved
 			}
 
--- a/pkg/cli/cmd/workload/shared.go
+++ b/pkg/cli/cmd/workload/shared.go
@@ -245,8 +245,13 @@ const (
 	defaultReconcileTimeout       = 5 * time.Minute
 	fluxKustomizationPollInterval = 500 * time.Millisecond
 	argoCDApplicationPollInterval = 500 * time.Millisecond
-	reconcileConcurrency          = 5
-	reconcileCmdLong              = "Trigger reconciliation/sync and wait for completion. " +
+	// argoCDSourceWarmupGrace bounds how long a source-availability error is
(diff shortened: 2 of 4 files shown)


devantler:

## Verification

**Branch base.** The `applications/ksail` gitlink was **7 commits behind** `origin/main`, so this branch is cut from `origin/main` (`56a39e15`) rather than the checked-out pin — otherwise the proofs below would describe a tree this PR does not merge into.

**RED → GREEN.** `argocd.IsColdStartTransient` did not exist; the new test failed to build (`undefined: argocd.IsColdStartTransient`), then passed once implemented. Full affected-package run, `-count=1` so nothing is served from the test cache: `pkg/client/argocd` **ok 0.647s**, `pkg/cli/cmd/workload` **ok 52.667s** (plus `cipher`, `gen`).

**Both states are asserted, which is what makes the test non-vacuous** — the point is not that transients are tolerated, but that tolerance is *bounded*:

| Case | Elapsed | Expected |
|---|---|---|
| source unavailable, cold start | inside window | retryable |
| source unavailable, real misconfiguration | outside window | **terminal** |
| failed operation | inside window | **terminal** — never masked |
| unrelated error / `nil` | inside window | **terminal** / not transient |
| exact window boundary | `== grace` | **terminal** |

The three terminal-inside-the-window cases are the no-regression guard: they fail if the gate is ever widened from `ErrSourceNotAvailable` to "any error", which is the obvious wrong fix.

## A design note worth recording

My first instinct was to reclassify the *message* — treat `connection refused` as transient in `isSourceRelatedError`. **That is wrong, and an existing test proves it:** `failed to fetch repository: connection refused` is asserted to be `ErrSourceNotAvailable`, and it should be. A cold-start connection failure and a genuinely unreachable repository are **textually identical**, so no pattern list can separate them, and making that string transient would have meant weakening a correct assertion to make my change pass.

Elapsed poll time is the only real discriminator, which is why the gate lives in the poll loop and the classifier is untouched.

## Scope — `Part of`, not `Fixes`

The issue defines slice 1 as the readiness gate, with slices 2–3 to follow, and its acceptance criteria require the `--cni Calico`/`--cni Cilium` System-Test legs to pass **reliably across repeated runs**. That is E2E behaviour on the real reconcile path, so this PR's CI matrix is the acceptance signal — and one green run is not yet "reliably". Leaving #5948 open until that evidence exists.

**Honest limitation:** the unit tests cover the decision function in both states; the *wiring* into the poll loop has no unit test, because `pollUntilApplicationReady` takes a concrete `*argocd.Reconciler` with no seam for a fake. The System-Test matrix is what exercises it.

devantler:


@coderabbitai review

devantler:

## Readiness — all three conditions hold at head `023f9114`

**1. Programmatically tested.** RED→GREEN recorded above: `argocd.IsColdStartTransient` did not exist (test failed to build), then passed once implemented; affected packages run with `-count=1` so nothing came from the test cache. Both states asserted, with three terminal-inside-the-window controls that fail if the gate is ever widened to "any error", plus the `== grace` boundary.

**2. Reviewed.** CodeRabbit at this exact head: `Review completed`, summary comment updated 17:05:37Z naming `023f9114d24e93ad31fe611f0716a31bc988fdd5` — *"No actionable comments were generated in the recent review."* Zero review-body finding sections (only the informational `🔇 Additional comments (4)`, which is non-actionable), zero unresolved threads, no rate-limit or service marker.

**3. Tried and evaluated as a user — I read the run, not just the badge.** The honest limitation recorded above is that the *wiring* into `pollUntilApplicationReady` has no unit test, so the System-Test matrix is what exercises it. That matrix is now **56/56 green, 0 failures**, and specifically **all five `--gitops-engine ArgoCD` legs passed** — KWOK, Vanilla, K3s, VCluster and Talos — which is the real reconcile path this change sits on, driven end to end.

Judged as the change's user: a cold-start ArgoCD source failure inside the grace window no longer fails the readiness poll, while a genuine misconfiguration outside it still fails fast rather than hanging for the full timeout.

**Scope is unchanged: `Part of #5948`, not `Fixes`.** That issue's acceptance criteria require the CNI legs to pass *reliably across repeated runs*, and one green matrix is not yet "reliably", so #5948 stays open for that evidence.

PR #6645, opened 2026-08-24
fix(clusterapi): bind EKS capture to create account

### LINKED ISSUE #6617: fix(clusterapi): capture compares selector names, so a repointed profile's contents still bind an unrelated EKS cluster

## Evidence

#6615 closes #6451 by re-resolving the AWS selector at capture and refusing when it no longer names
the identity the create was pinned to. The comparison is `AWSResolution.SelectorEquals`, which
compares selector **names**: profile, access-key id, region, and the env-var sources.

Those values come from `ResolveAWS`, which reads the **environment and option aliases only** — it
does not read the AWS credentials file. So for a profile-based selection `AccessKeyID` is empty and
the only discriminating field is the profile **name**.

Read together: if the *contents* of the selected profile are edited mid-create to name a different
account, the profile name is unchanged, `SelectorEquals` returns true, and capture proceeds to freeze
and record a same-named cluster in the other account — the outcome #6451 exists to prevent.

Checked, not assumed, at `54dbd8ea`:

- `NewAWSOptionsResolver` carries env-var **names**, and `AWSOptionsResolver.Value` reads the
  environment at call time, so the capture-time re-resolution is genuinely live rather than a replay
  of stored values. That half of the guard is sound.
- `ResolveAWS` populates `AccessKeyID` from `resolver.Value(AWSAccessKeyID)`, i.e. the env var — so it
  is empty whenever the selection is by profile.

## Why it matters

#6451's own preconditions describe exactly this shape: "a mutable selector (profile or default
chain), repointed mid-create". Editing what a profile points at is a repoint of a mutable selector;
it is simply not a rename. The consequence is unchanged from #6451 — the recorded account, ARN and
`CreatedAt` all come from the wrongly-observed cluster, so the record is internally consistent and
every later guarded delete, start or stop verifies successfully against an unrelated live cluster and
then mutates it. An EKS delete cannot be undone.

This is narrower than #6451 as filed: it needs the profile's credentials to be rewritten during the
15–20 minute create, not merely a different profile selected. It is the residual that the shipped
guard does not cover, recorded here so closing #6451 does not close it silently.

## Expected behaviour

Capture refuses when the selector resolves to a **different AWS account** than the create started
under, not only when the selector's name changed.

## The shape that closes it

#6451's option 2, which #6615 deliberately did not take in order to avoid the extra call: resolve
the account once at create start (one STS `GetCallerIdentity`), carry it on the create identity, and
compare accounts at capture. `SelectorEquals` stays as the cheap first check; the account comparison
is what makes the refusal complete.

## Acceptance criteria

- [ ] Capture refuses when the resolved account at capture differs from the account observed at
      create start, demonstrated by a test that fails without the change.
- [ ] A same-account key rotation still captures successfully — the property `SelectorEquals` was
      written to preserve, and the reason a naive "compare everything" check is wrong.
- [ ] The added call is made once at create start, not per capture.
- [ ] The doc comment on `defaultEKSCapture` states the guarantee accurately; as delivered it reads
      "a mid-create profile or access-key change", which a profile **content** change is not.

Rough size: small — one STS call, one field on the create identity, one comparison, two tests.

Part of #6451.

devantler:

## Summary

- resolve the selected AWS account once before EKS provisioning starts
- carry that account through create and refuse capture-time account drift before any state write
- preserve same-account credential rotation while closing same-named profile repoints

## Validation

- focused RED: missing account pin and capture-account refusal failed before implementation
- focused GREEN: account ordering, fail-closed resolution, mismatch refusal, and rotation acceptance
- `go test ./pkg/cli/clusterapi`
- `go test -p 1 ./...`
- `golangci-lint run --default=none --enable err113 --enable funlen --enable paralleltest --enable wrapcheck ./pkg/cli/clusterapi`
- `go build -o /private/tmp/ksail-binary-01a035d3 .`

Fixes #6617

### DIFF
--- a/pkg/cli/clusterapi/eks_create_identity_test.go
+++ b/pkg/cli/clusterapi/eks_create_identity_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/devantler-tech/ksail/v7/pkg/apis/cluster/v1alpha1"
 	"github.com/devantler-tech/ksail/v7/pkg/cli/clusterapi"
+	"github.com/devantler-tech/ksail/v7/pkg/svc/credentials"
 	clusterprovisioner "github.com/devantler-tech/ksail/v7/pkg/svc/provisioner/cluster"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -54,6 +55,7 @@ type captureRecord struct {
 	pinned          bool
 	awsOptions      v1alpha1.OptionsAWS
 	selectionRegion string
+	accountID       string
 }
 
 // recordCaptureForCreate runs an EKS create whose spec resolves AWS through awsOptions, and returns
@@ -83,12 +85,13 @@ func recordCaptureForCreate(
 	records := make(chan captureRecord, 1)
 
 	service.SetEKSOwnershipCaptureRecorderForTest(
-		func(name string, pinned bool, options v1alpha1.OptionsAWS, region string) {
+		func(name string, pinned bool, options v1alpha1.OptionsAWS, region, accountID string) {
 			records <- captureRecord{
 				name:            name,
 				pinned:          pinned,
 				awsOptions:      options,
 				selectionRegion: region,
+				accountID:       accountID,
 			}
 		},
 	)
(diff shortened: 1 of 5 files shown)


devantler:

@coderabbitai review

devantler:

@coderabbitai review

Please review exact head b817a8fc8a3ba17dce46e960398c381e842a3c13; this follow-up only clarifies the documented account-drift refusal that the prior review requested.

devantler:

Exact-head ancillary-check disposition for `b817a8fc8a3ba17dce46e960398c381e842a3c13`:

The CodeRabbit docstring-coverage warning is non-blocking and refuted for this change. KSail's authoritative maintenance contract and required CI do not define CodeRabbit's 80% touched-function threshold; this exact head passes the repository-required build, lint, test, generated-artifact, security, and required-check gates. The heuristic counted 13 touched functions across five files, including test functions and unexported helpers, while the production ownership helpers that benefit from behavioral comments already document account pinning and account-drift refusal. No code change is warranted solely to satisfy this vendor heuristic.

devantler:

**Diagnosed the `BLOCKED` state — it is a real gate, not a stale `mergeStateStatus`, and it is
invisible to a check-run sweep.** Recording it because the obvious reading of this PR is the wrong
one.

Every check-run at head `b817a8fc` is green: 16 `success`, 18 `skipped`, **zero failures**. Unresolved
threads are 0, `required_approving_review_count` is 0, and there is a current-head CodeRabbit green.
By the ordinary pentad this PR looks mergeable — and it is not.

The gate is a **required workflow that failed before it created any job**:

```
run     32791158313   .github/workflows/validate-go-project.yaml
event   pull_request        head_sha  b817a8fc8a3ba17dce46e960398c381e842a3c13
status  completed/failure   run_attempt 1
jobs    total_count = 0
created 2026-08-24T23:51:07Z   updated 2026-08-24T23:51:07Z
```

A run with zero jobs **publishes no check-run**, so it cannot appear in
`commits/<sha>/check-runs` at all. That is why the rollup reads clean while the merge stays blocked —
the failure is only visible by enumerating required workflows directly.

**I am not re-running it.** `created == updated` to the second, one attempt, zero jobs: that is a
startup/config rejection, not a transient, and the rule is to rerun only a demonstrated transient.
Re-running would re-hit the same rejection and tell us nothing.

**Named blocker: `devantler-tech/actions#1041`** — *"fix(validate-go): keep signed fixes valid as a
required workflow"* — which addresses exactly this required-workflow startup failure. It is open and
actively being worked (head pushed ~26 min before this comment, with a review in flight), so I am
deliberately not contending on it.

**Blocker:** devantler-tech/actions#1041 | last-verified 2026-08-25: open, in flight, not yet merged

Parking this PR on that blocker rather than leaving it looking merely stale. Once #1041 lands, this
needs a fresh run of the required workflow — not a re-run of `32791158313`, which is pinned to the
old configuration.

Issue #6660, opened 2026-08-25
Bump docker/docker past the moby/go-archive v0.1.0 pin so GO-2026-6253 can be fixed rather than allowlisted


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`GO-2026-6253` (CVE-2026-17106, `github.com/moby/go-archive`) was published `2026-08-25T19:43:10Z` — five minutes after the last green run of `CI - KSail` on `main` (`19:38:04Z`). The next run of `main`, at `22:22:55Z`, failed on it, and `main` has been red since:

```
Reachable vulnerabilities:
  - GO-2026-6253 (fixed: v0.3.0) - BLOCKING
::error::Reachable vulnerabilities not accepted in ./.govulncheck-allow.txt: GO-2026-6253
```

The advisory is a crafted tar archive writing outside its extraction directory. It reaches us through k3d, not through our own code:

```
github.com/devantler-tech/ksail/v7/pkg/fsutil/validator/k3d
  -> github.com/k3d-io/k3d/v5/pkg/runtimes
  -> github.com/k3d-io/k3d/v5/pkg/runtimes/docker
  -> github.com/docker/docker/pkg/archive
  -> github.com/moby/go-archive
```

## Why the obvious fix does not work

`go.mod` already pins `moby/go-archive` back to `v0.1.0`, because `v0.2.0` removed the package-level `Compression` type alias that `docker/docker` still uses. The advisory is fixed in `v0.3.0`, so the pin now sits below the fix.

Raising it was tried and **fails to compile** — measured on `f4fd458f` with `go-archive v0.3.3`:

```
# github.com/docker/docker/pkg/archive
docker@v28.5.2+incompatible/pkg/archive/archive_deprecated.go:103:47: undefined: archive.Compression
docker@v28.5.2+incompatible/pkg/archive/archive_deprecated.go:159:43: undefined: archive.Compression
```

`v28.5.2` is the newest `v28` on the module proxy, so there is no patch-level bump that clears it. Moving to `docker/docker v29` is the real fix, and it is not a one-liner: `go.mod` separately pins `docker/cli` to v28 because k3d v5.9.0-rc.0 uses the docker/docker monolith types that v29 migrated away from. So this is a coordinated docker/docker + docker/cli + k3d bump.

## What was done in the meantime

[...]

PR #6662, opened 2026-08-26
fix(security): patch reachable dependency vulnerabilities

devantler:

## Summary

- replace the vulnerable `moby/go-archive` v0.1.0 override with the exact
  v0.3.0 source that fixes GO-2026-6253
- retain only the deprecated compression aliases still required by Docker
  28.5.2, with a compatibility regression test and removal condition
- update `golang.org/x/image` to v0.45.0 for GO-2026-6222 in both Go modules

The vulnerability allowlist is unchanged.

## Provenance

The embedded module is upstream tag `v0.3.0` at
`1c23372e409716c3691a540871806083644f348a` with module checksum
`h1:nos4BtzzUIqB406BgQnWGMI4qib9BZ8XUHU+ucv/n1c=`. Its shipped source,
tests, internal packages, module files, and license match upstream byte for
byte; KSail adds only `compat_legacy.go`, its test, and `KSail-PATCH.md`.

## Validation

- `go -C third_party/go-archive test ./...` — pass
- `go build -o /private/tmp/ksail-6253-maint .` — pass
- built CLI `--help` smoke test — pass
- `go -C desktop test ./...` — pass
- `go mod tidy -diff` in root and desktop — pass
- `govulncheck -mode=binary` on the rebuilt binary — GO-2026-6253 and
  GO-2026-6222 absent; the remaining nine findings exactly match the existing
  no-fix allowlist
- `go test ./...` — all packages passed except the chat helper-process package
  under whole-suite resource contention; that package passed alone on both
  exact `main` and this head
- source-mode `govulncheck ./...` — UNKNOWN locally after the bounded
  ten-minute analysis window; hosted vulnerability CI remains required
- `golangci-lint run --timeout 5m` — UNKNOWN locally because the analyzer did
  not exit or report its configured timeout; hosted lint remains required

## Delivery state

Draft until hosted source vulnerability scanning, lint, tests, and review are
green at this exact head.


devantler:

Heads-up on an interaction between this PR and **#6661**, which is in flight on the same advisory from another lane. There is **no textual conflict** — this PR does not touch `.govulncheck-allow.txt` — which is exactly why it is worth stating, because the two can both merge cleanly and leave the repository in a state its own policy forbids.

**What #6661 does.** `main` has been red since 22:22Z on `GO-2026-6253` and the gate is required, so every ksail PR is blocked behind it — including this one. #6661 is a one-line, explicitly **temporary** allowlist entry to unblock the pipeline, recorded as *not* one of the file's permanent "no upstream fix exists" entries: it names the constraint blocking the bump and cites **#6660** as the issue whose closure removes it.

**Why that matters here.** This PR is the real fix, and it is the better one — vendoring `v0.3.0` and retaining only the deprecated compression aliases sidesteps the `docker/docker` v28.5.2 `archive.Compression` break that made a straight pin bump uncompilable. But once it lands, `GO-2026-6253` is genuinely fixed, and the allowlist header states the policy plainly: *"The moment an upstream fix ships, delete the entry and bump the affected module instead."* An entry left behind for an advisory that no longer fires is precisely the permanent risk-acceptance the temporary category was written to prevent.

**So, whichever order they merge in:**

- **#6661 first, then this** — this PR should also **delete the `GO-2026-6253` entry** (the block at the foot of `.govulncheck-allow.txt`, including its `# --- github.com/moby/go-archive` comment header) and **close #6660**, since it delivers exactly what #6660 tracks.
- **This first** — #6661 becomes unnecessary; close it and #6660 rather than merging a risk-acceptance for something already fixed.

I am driving #6661 toward merge because it unblocks the queue in minutes and this PR's ~80-file vendored module deserves a real review rather than a rushed one. I will not touch this branch. If it lands first, I will close #6661 and #6660 myself.

For reference, `main`'s red run failed on exactly two jobs of 34 — `✅ Validate Go Project / 🛡️ Vulnerability Scan` and the `CI - Required Checks` aggregator downstream of it — so this advisory is the whole of the breakage, and either PR resolves it.

devantler:

Retiring this draft because older active repair lane #6661 became visible while
the hotfix was being validated. The signed commit and full evidence remain
available as a boundary-preserving implementation candidate for #6660; closing
the duplicate does not endorse expanding the vulnerability allowlist.

PR #6705, opened 2026-08-27
fix(deps): pin runc so Dependabot can resolve the root module graph

### LINKED ISSUE #6664: The grouped root Dependabot job cannot resolve the module graph: runc removed libcontainer/user, which ory/dockertest still imports

## Evidence

The GitHub-managed Dependabot job for the **grouped root** unit — `go_modules in /. - Update` — is the newest run of that unit on `main` and it **failed**, at `bc4a60ce7b` on 2026-08-25T17:12:49Z ([run 32876550266](https://github.com/devantler-tech/ksail/actions/runs/32876550266)). A repeated managed failure is ours to repair rather than a first-of-streak to ignore.

The updater reports **two** failures, both the same error type:

| Dependency | Error type |
|---|---|
| `github.com/siderolabs/image-factory` | `dependency_file_not_resolvable` |
| `github.com/siderolabs/talos` | `dependency_file_not_resolvable` |

## The mechanism

Both fail on the same unresolvable import, and the updater prints the whole chain:

```
ksail/v7/pkg/cli/clusterapi imports
  github.com/getsops/sops/v3/keyservice imports
  github.com/getsops/sops/v3/hcvault  tested by
  github.com/getsops/sops/v3/hcvault.test imports
  github.com/ory/dockertest/v3 imports
  github.com/ory/dockertest/v3/docker imports
  github.com/ory/dockertest/v3/docker/pkg/homedir imports
  github.com/opencontainers/runc/libcontainer/user:
    module github.com/opencontainers/runc@latest found (v1.5.1),
    but does not contain package github.com/opencontainers/runc/libcontainer/user
```

`runc` **removed** `libcontainer/user` by v1.5.1, and `ory/dockertest/v3` still imports it through `docker/pkg/homedir`. When Dependabot resolves the module graph to compute an update it takes `runc@latest`, finds the package gone, and the whole resolution fails — so it never gets as far as proposing the `image-factory` or `talos` bumps.

**This is why the ordinary build stays green.** The path reaches us only through `hcvault.test` — a *test* package of a transitive dependency — so it is inside the module graph Dependabot must resolve but outside anything we compile. Repository CI is green on this head (44 checks, all success/skipped), which is exactly why the failure has gone unnoticed.

## Not a duplicate of #6607

#6607 covers a **different unit with a different cause**: the ebpf-specific job failing `pull_request_exists_for_latest_version` because #6589 is open and red. That issue explicitly scoped itself to the ebpf job. This one is the *grouped root* job failing `dependency_file_not_resolvable`. Both must clear before the root-module ecosystem updates again; neither fixes the other.

## Expected behaviour

The grouped root Dependabot job resolves the module graph and proposes the `image-factory` and `talos` updates.

## Candidate fixes, cheapest first

1. **Pin `github.com/opencontainers/runc` to the last version containing `libcontainer/user`** with an explicit `require` (or `replace`) in the root `go.mod`, so graph resolution stops taking `@latest`. Smallest change; needs a comment recording why, or a later tidy will drop it.
2. **Raise `github.com/ory/dockertest/v3`** to a release that no longer imports `runc/libcontainer/user`. Cleaner if such a release exists — it removes the pin rather than adding one.
3. **Drop the `hcvault` test path from the graph** if `sops`'s Vault key service is not actually used by `pkg/cli/clusterapi`. Largest change and the most likely to be wrong; listed for completeness only.

## Acceptance criteria

- [ ] The root `go.mod` resolves without reaching a removed `runc` package.
- [ ] `go mod tidy` and the full build/test suite pass unchanged.
- [ ] The next `go_modules in /. - Update` run on `main` completes without `dependency_file_not_resolvable`.
- [ ] If option 1 is taken, the pin carries a comment naming this issue so a future tidy does not silently undo it.

## Size

Small-to-medium — the diagnosis is settled; the work is choosing between options 1 and 2 and verifying the graph.

### LINKED ISSUE #6666: fix(deps): Dependabot go_modules updates blocked by removed runc/libcontainer/user package

## Evidence

Dependabot's `go_modules in /.` update job has failed on every run since **2026-08-24** and has
**never produced a green run** for that unit (`2026-08-10`, `2026-08-24`, `2026-08-25` all
`failure`). Every *other* ecosystem unit in the same workflow succeeds — `docker in /...` (×12+),
`npm_and_yarn in /docs`, `go_modules in /desktop` — so this is isolated to the main module, not a
broken workflow.

Failing run: [`32876550266`](https://github.com/devantler-tech/ksail/actions/runs/32876550266) —
`Dependabot encountered '2' error(s)`, both `dependency_file_not_resolvable`, for
`github.com/siderolabs/talos` and `github.com/siderolabs/image-factory`:

```
go: finding module for package github.com/opencontainers/runc/libcontainer/user
go: github.com/devantler-tech/ksail/v7/pkg/cli/clusterapi imports
	github.com/getsops/sops/v3/keyservice imports
	github.com/getsops/sops/v3/hcvault tested by
	github.com/getsops/sops/v3/hcvault.test imports
	github.com/ory/dockertest/v3 imports
	github.com/ory/dockertest/v3/docker imports
	github.com/ory/dockertest/v3/docker/pkg/homedir imports
	github.com/opencontainers/runc/libcontainer/user: module github.com/opencontainers/runc@latest
	found (v1.5.1), but does not contain package github.com/opencontainers/runc/libcontainer/user
```

## Why it happens

`runc` removed `libcontainer/user` in **v1.2** (it moved to `github.com/moby/sys/user`).
`ory/dockertest/v3` still imports the old path.

`go.mod` requires `github.com/opencontainers/runc` **not at all**, so when Dependabot re-resolves the
graph it has to *find a module providing that package*, looks at `runc@latest` (v1.5.1), and the
package is not there. `go.sum` carries `runc v1.3.6`, which also post-dates the removal.

## Why CI is green while this is red

`go build` / `go test ./...` only resolve packages in KSail's own build-and-test graph. The broken
edge is a **test dependency of a dependency** (`sops/v3/hcvault`'s *test* imports), which only a
full `go mod tidy`-style resolution reaches — which is exactly what Dependabot does. So the repo's
own checks cannot see this, and it will stay silent until someone reads the Dependabot logs.

## Impact

Go dependency updates for the **main module** are not landing. That includes Go **security** bumps —
directly relevant while #6660/#6665 is patching reachable Go CVEs. The job still opens PRs for
dependencies it can resolve (azcore, cel-go, sigstore were created in the same run), so the failure
is partial and easy to miss.

## Candidate remedies (a decision is needed — none is free)

1. **Explicit `require github.com/opencontainers/runc v1.1.x // indirect`** — the last line carrying
   `libcontainer/user`. Resolves the graph deterministically. It is needed only to *resolve* the
   graph and is never compiled into any KSail binary (test-dep of a dependency's test), so
   reachability is nil — but it pins an EOL branch and will show up in dependency inventories.
2. **`replace` directive** — this repo treats replaces as debt to remove (see #3772).
3. **Upstream fix** — `ory/dockertest` migrating to `moby/sys/user` (KSail already has
   `moby/sys/user v0.4.1` indirect), or `sops` dropping `dockertest` from its `hcvault` tests.
   `dockertest v3.12.0` (2025-03-12) is the **latest** release, so there is no newer version to bump
   to; this path is slow and not in our control.
4. **Scope the Dependabot config** so the main-module Go updater does not re-resolve the offending
   path — treats the symptom, not the cause.

## Acceptance criteria

- The `go_modules in /.` Dependabot unit completes without `dependency_file_not_resolvable`.
- The chosen remedy's security trade-off is recorded in the PR body.
- A check exists that would catch this class again without waiting for a human to read Dependabot
  logs (the current gap: repo CI structurally cannot see it).

Rough size: S–M once the remedy is chosen; the choice itself is the work.

devantler:

### Why

Dependabot's root Go update job has failed every run since 2026-08-24 and has **never**
had a green run. Root Go dependency updates — including security bumps — are not landing,
while every other ecosystem in the same workflow succeeds. The failure is partial and
silent: the job still opens PRs for what it can resolve, so nothing goes red and nobody
notices without reading Dependabot's own logs.

### What

Pins one indirect dependency to the version the build **already** uses, which makes
Dependabot's resolution deterministic again, and adds a CI job so the next occurrence of
this class fails loudly with the fix instead of going unnoticed.

⚠️ **New dependencies — flagging per convention.** Making that pin explicit also promotes
the pinned module's own requirements into this module's graph: **9 new indirect entries**,
and one existing dependency (`cyphar/filepath-securejoin`) moves up a minor version. No
dependency is downgraded and nothing lands on an end-of-life branch.

Two corrections to the issue's premises, both measured rather than reasoned — they change
the decision it was asking for:

- The issue records the package as removed in **v1.2**, which would have forced a pin to an
  end-of-life v1.1.x. It was actually removed in **v1.4**, so the pin lands on a current
  branch and that trade-off disappears.
- The issue's leading remedy — an explicit requirement — is **not durable on its own**:
  `go mod tidy` strips it, which would have merged green and silently regressed. That is
  precisely why the CI guard is in the same change rather than deferred.

Fixes #6666
Closes #6664


devantler:

Closing without merging. Both halves of this change are unsound, and I would rather withdraw it
than leave a draft that looks like a fix.

- **The pin cannot merge.** Root `go mod tidy` strips it, and CI's 📦 Tidy check enforces root
  tidiness — it went red here for exactly that reason. I had checked this repo's own workflow
  files and concluded root tidy was not gated; the check comes from the reusable *Validate Go
  Project* workflow, so that conclusion was wrong.
- **The guard is vacuous.** It returns 0 on the pristine tree while Dependabot is failing, so it
  would have shipped as coverage for a defect it cannot see.

The full measurements — including the corrected v1.4 removal boundary, the true cost of the pin,
and the elimination of remedy 1 — are recorded on #6666, which stays open. #6664 remains closed as
a duplicate of it.

Nothing from this branch reached `main`.

PR #6765, opened 2026-08-29
fix(ci): stop the path-filter comment self-triggering the to-do scanner

### LINKED ISSUE #6763: scanner
caller\. Its subject is a WORKFLOW file, not a Go file, so the org\-required

validate\-go\-project run cannot reach it: that workflow gates its own test job

on a \`go\` filter of \*\*/\*\.go\|go\.mod\|go\.sum\|\.golangci\.\* and skips everything

else\. A scanner\-only bump therefore merges with the contract unexecuted and

reddens main instead \(ksail\#6750 merged 2026\-08\-28T17:17Z with its test job

\`skipped\`; main went red 94 minutes later\)\. This filter feeds a job of its own

so that class of change is gated on the PR that makes it\. The harness sources

and THIS workflow are included for the same reason dependency\-test\-graph

includes ci\.yaml below: the job and its assertions both live in files this

filter would otherwise not match, so a PR editing only one of them could

weaken or delete the gate while the aggregator recorded it as an ordinary

skip\.

https://github.com/devantler-tech/ksail/blob/43da3331373fea3ea64e3d6735ccd7a2b1fde347/.github/workflows/ci.yaml#L166-L178

```yaml

              - 'go.sum'
              - '.govulncheck-allow.txt'
              - '.github/actions/**'
            # The Go harness in internal/ciharness asserts the shape of the TODO scanner
            # caller. Its subject is a WORKFLOW file, not a Go file, so the org-required
            # validate-go-project run cannot reach it: that workflow gates its own test job
            # on a `go` filter of **/*.go|go.mod|go.sum|.golangci.* and skips everything
            # else. A scanner-only bump therefore merges with the contract unexecuted and
            # reddens main instead (ksail#6750 merged 2026-08-28T17:17Z with its test job
            # `skipped`; main went red 94 minutes later). This filter feeds a job of its own
            # so that class of change is gated on the PR that makes it. The harness sources
            # and THIS workflow are included for the same reason dependency-test-graph
            # includes ci.yaml below: the job and its assertions both live in files this
            # filter would otherwise not match, so a PR editing only one of them could
            # weaken or delete the gate while the aggregator recorded it as an ordinary
            # skip.
            todos-contract:
              - '.github/workflows/todos.yaml'
              - '.github/workflows/ci.yaml'
              - 'internal/ciharness/**'
            # Deliberately NARROWER than `code`: the job resolves a module graph, so only the
            # module files matter — not every .go file. It also includes THIS workflow, because
            # the check lives here and a workflow-only PR would otherwise be able to weaken or

```

devantler:

### Why

Our to-do scanner files a GitHub issue whenever it sees the uppercase marker in a comment. That is exactly what it is for — but it means a comment that merely *talks about* the scanner files an issue about itself. One landed yesterday and produced #6763: an issue whose entire body was a comment block, with no work behind it. It was also the only issue in the whole portfolio missing an Issue Type, so it showed up as triage debt on top of being noise.

### What

Refers to the marker as "to-do" in that one explanatory comment — the same convention the shared workflow in `devantler-tech/actions` already applies to its own source, for this exact reason — and adds a test so the trap cannot come back unnoticed.

The guard is deliberately narrow: filing issues from real markers is the scanner's job, so it checks only that this one comment block stays clean rather than banning the marker repository-wide.

Fixes #6763

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -163,7 +163,7 @@ jobs:
               - 'go.sum'
               - '.govulncheck-allow.txt'
               - '.github/actions/**'
-            # The Go harness in internal/ciharness asserts the shape of the TODO scanner
+            # The Go harness in internal/ciharness asserts the shape of the to-do scanner
             # caller. Its subject is a WORKFLOW file, not a Go file, so the org-required
             # validate-go-project run cannot reach it: that workflow gates its own test job
             # on a `go` filter of **/*.go|go.mod|go.sum|.golangci.* and skips everything
@@ -176,6 +176,9 @@ jobs:
             # filter would otherwise not match, so a PR editing only one of them could
             # weaken or delete the gate while the aggregator recorded it as an ordinary
             # skip.
+            # ("to-do" is deliberately lowercase and hyphenated here: the scanner keys on the
+            # uppercase marker, so spelling it out in prose files an issue against this very
+            # comment. ksail#6763 was exactly that — a false positive with no work behind it.)
             todos-contract:
               - '.github/workflows/todos.yaml'
               - '.github/workflows/ci.yaml'
--- a/internal/ciharness/todos_workflow_test.go
+++ b/internal/ciharness/todos_workflow_test.go
@@ -23,7 +23,12 @@ type todosWorkflow struct {
 	} `yaml:"jobs"`
 }
 
-// TestTODOScannerExcludesOnlyVendoredSources pins the scanner's delegation and its blast radius.
+// Pins the shared scanner's delegation and the blast radius of its ignore pattern.
+//
+// This comment deliberately does not open with the function's own name, the way a doc comment
+// normally would: that name embeds the uppercase marker, and the scanner keys on the marker
+// wherever it appears in a comment. See TestCIFilterCommentDoesNotSpellTheScannerMarker below,
+// which pins the same trap after it fired once for real.
 //
 // KSail calls the shared scanner in devantler-tech/actions instead of carrying its own copy, so the
 // job must stay a bare reusable-workflow call: a runs-on or steps key here would mean the
--- a/internal/ciharness/todos_workflow_test.go
+++ b/internal/ciharness/todos_workflow_test.go
@@ -2,6 +2,7 @@ package ciharness_test
 
 import (
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


devantler:

CI has settled: 65 checks green. The single red is `Analyze (go)`, GitHub's **managed** CodeQL
(`event: dynamic`) failing on cgo extraction — intermittent rather than caused by this change
(#6766 and #6742 pass it at the same time, and `main` passes it on its 5 newest runs). Tracked in
#6767, which also records that it gates the merge through the `code_scanning` ruleset rule.



@coderabbitai full review

devantler:



@coderabbitai full review

devantler:

## Readiness record @ `88aea14269d618926e5bec141b2ddd660c8c43df`

**Tried and evaluated as a user — RED/GREEN ablation, run locally at this exact head.**

| State | `ci.yaml` filter comment | `TestCIFilterCommentDoesNotSpellTheScannerMarker` |
|---|---|---|
| GREEN (as shipped) | says `to-do` | **PASS** |
| RED (ablated) | marker restored | **FAIL** at `todos_workflow_test.go:119` |

The ablation fixture was asserted to have actually built (`git diff --numstat` → `1+/1-`, the reworded
line flipping back), and the failure was checked to be **the intended assertion** rather than a sibling
conjunct — it is the `NotContains` on the marker, reporting
`comment must not spell the scanner marker: doing so files a spurious issue (ksail#6763)`. Reverted
afterwards and re-run green, so the tree is unmodified.

**Census, with a positive control.** Marker occurrences on **comment** lines: `ci.yaml` → 0,
`todos_workflow_test.go` → 0. The control (a fixture comment containing the marker) matched 1, so the
zero is a real absence rather than a broken pattern.

Worth recording explicitly, because a naive census reads as a failure here: the marker still appears
3× in `ci.yaml` and 2× in the test file — as job *names*, an `echo` string, and the Go test-function
identifier. None is a comment, and the scanner keys on the marker **in comments**, so those are
correctly untouched. The test scopes itself the same way: it walks only the contiguous comment block
immediately above the `todos-contract:` filter key, which is exactly the block that produced #6763.

The test also carries a proper guard against passing vacuously —
`require.NotEmpty(t, block, "the todos-contract filter must keep its explanatory comment")` — so
deleting the comment fails the test rather than silently satisfying it.

**Pentad:** checks green · 0 unresolved threads · 0 non-thread findings · no conflict ·
`green_review=cr@88aea1426` (CodeRabbit completed 05:04:07Z, "No actionable comments were generated";
the only body section is the excluded informational `🔇 Additional comments`).

Promoting and merging on that basis. Closes the untyped-residual issue #6763.

PR #6766, opened 2026-08-29
fix(security): pin the k3k privileged-pod exemption to the controller identity

### LINKED ISSUE #6586: Pin the k3k privileged-pod guard exemption to the exact controller identity

### Evidence

The k3k privileged-pod guard admits an unsafe pod only when it looks like the KSail-managed k3k
server pod. Until #6261 that test was pod name + labels, both of which anyone able to create a Pod
in the namespace can reproduce, so the namespace's blanket `privileged` PSA exemption was reachable
by a crafted pod.

#6261 closed the user-spoofing path by additionally requiring
`request.userInfo.username.startsWith('system:serviceaccount:')`. `userInfo` is populated by the
API server from the authenticated request and cannot be forged by the requester.

### Problem

That check is at the *class* level: it proves a service account made the request, not *which* one.
A compromised or hostile service account holding pod-create in the k3k namespace still satisfies
the exemption and can obtain a privileged pod.

The stronger form pins the exact creator, e.g.
`request.userInfo.username == 'system:serviceaccount:kube-system:statefulset-controller'`.

### Why it was not done in #6261

Shipping the exact identity asserts *which* controller creates the k3k server pod. k3k is an
external controller and KSail only creates the `Cluster` CR, so that fact is not verifiable from
this repository. The guard runs with `FailurePolicy: Fail`, so if the assumption is wrong the
server pod is rejected and clusters become permanently unprovisionable. That is a worse outcome
than the residual, so the safer partial fix shipped instead.

### Acceptance criteria

- [ ] Determine the actual creator identity of the k3k server pod against a live k3k cluster
      (`kubectl get pod <server-pod> -o yaml` ownerReferences, plus an audit-log or
      `SelfSubjectReview`-style confirmation of the creating principal).
- [ ] Narrow `serverPodExemptionTemplate` to that exact principal.
- [ ] Add a negative case to `TestGuardCEL_ExemptionAdmitsOnlyTheManagedServerPod` covering a
      non-controller service account, so the narrowing is pinned.
- [ ] Verify a real k3k cluster still provisions with the narrowed guard.

### Size

Small once a live k3k cluster is available; the blocker is the live confirmation, not the change.

Follow-up to #6261.

devantler:

### Why

The k3k namespace carries a blanket `privileged` Pod Security exemption, and an admission guard is what keeps that exemption reachable only by the cluster's own server pod. Until now the guard only checked that *a* service account made the request — not *which* one. Any service account able to create a pod in that namespace could copy the server pod's name and labels and obtain a privileged pod.

While pinning it, a second problem surfaced: the old check would have **rejected the real server pod outright** on any host cluster that runs its controllers under a single shared identity. Because the guard fails closed, that would have made k3k clusters impossible to create on those hosts.

### What

Narrows the exemption from "any service account" to the specific controller that actually creates the server pod, and accepts both spellings that controller can present depending on how the host cluster is configured — closing the escalation path without introducing the availability risk above.

The blocker recorded on the issue was that the creator could not be confirmed without a live cluster. It could: k3k is a direct dependency of this repository, and its source settles the question.

Fixes #6586

### DIFF
--- a/pkg/svc/provisioner/cluster/k3d/kubernetes_provisioner.go
+++ b/pkg/svc/provisioner/cluster/k3d/kubernetes_provisioner.go
@@ -726,20 +726,33 @@ const unsafePodExpression = "object.spec.hostPID == true || " +
 	"(has(object.spec.ephemeralContainers) && object.spec.ephemeralContainers.exists(c, " +
 	unsafeContainerPredicate + "))"
 
+// statefulSetControllerPrincipals are the identities that can create the k3k server pod.
+//
+// k3k runs the server as a StatefulSet (rancher/k3k pkg/controller/cluster/server: StatefulServer
+// builds an appsv1.StatefulSet named <cluster>-server), and a StatefulSet controller creates its
+// pods directly rather than through an intermediate ReplicaSet. The creator is therefore always
+// kube-controller-manager's StatefulSet controller, under one of two spellings: the per-controller
+// service account when the host runs with --use-service-account-credentials, and the shared
+// component identity when it does not. Both are populated by the API server from the authenticated
+// request, so neither can be forged, and neither is obtainable by a namespace user.
+const statefulSetControllerPrincipals = "['system:serviceaccount:kube-system:statefulset-controller', " +
+	"'system:kube-controller-manager']"
+
 // serverPodExemptionTemplate is the CEL expression admitting an unsafe pod only when it is a
 // KSail-managed k3k server pod. Formatted with the server pod name prefix and the cluster name.
 //
 // The pod name and labels are attacker-controlled: anyone able to create a Pod in this namespace
 // can reproduce them. They are therefore paired with `request.userInfo`, which the API server
 // populates from the authenticated request and a requester cannot forge, so a pod hand-crafted by
-// a namespace user no longer satisfies the exemption. The k3k server pod is created by a
-// controller acting as a service account, which is what this admits.
+// a namespace user no longer satisfies the exemption. The username is additionally pinned to the
+// exact principals in statefulSetControllerPrincipals, so a compromised or hostile service account
+// that merely holds pod-create in this namespace no longer reaches the exemption either.
 //
 // Every map lookup is key-guarded. CEL raises "no such key" on a missing map key, and because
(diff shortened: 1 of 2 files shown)


devantler:

CI has settled green at this head (57 checks, 0 failures). Requesting review.



@coderabbitai review

devantler:

## Readiness record @ `c774a690d65cbbba8c4745cd64bfc4bf8afbc817`

**Pre-merge finding — Docstring Coverage (66.67% vs 80%): refuted, not fixed.**
The gap is on `exemptionCases`, a **pre-existing** test helper this diff only *touched* (it gained two
table cases). Both symbols this change actually **adds** carry full doc comments — the
`statefulSetControllerPrincipals` constant and
`TestGuardCEL_ExemptionPinsTheStatefulSetControllerIdentity`. The threshold is missed only because the
check scopes to functions touched by the diff, so it reports a documentation gap this change did not
introduce and would not close by any edit to its own new code. Pushing a docstring onto an untouched
helper would stale a green review on a security fix to satisfy a warning about someone else's line.
(`guardEnv` and `evalGuardExpr` are undocumented for the same pre-existing reason — a separate cleanup
if it is ever worth one.)

**Tried and evaluated as a user.** I ran the guard's real behaviour rather than reading it: the CEL
expressions were compiled and evaluated against concrete pod objects (`go test -count=1 -run TestGuardCEL`,
all pass), including both new decisions —
- `server pod created by a shared kube-controller-manager identity is admitted` → admitted, so a host
  running **without** `--use-service-account-credentials` can still provision;
- `non-controller service account spoofing the server pod is rejected` → rejected, which is the residual
  this closes.

The regression test carries its own **negative control**: it reconstructs the previous
`startsWith('system:serviceaccount:')` form and asserts that form *did* admit the hostile service
account. So the test fails if the exemption is ever widened back, and it proves the narrowing changed a
real decision rather than restating current behaviour.

**Pentad @ this head:** 57 checks green / 0 failures · 0 unresolved threads · no actionable review
findings (the only category section is the excluded `🔇 Additional comments (1)`) · `mergeStateStatus:
CLEAN` · green review `cr@c774a690` (auto-summary naming this head, head status `CodeRabbit success —
Review completed`, no rate-limit marker).

Promoting and merging.

PR #6779, opened 2026-08-30
fix(deps): migrate cel-go to its cel.dev module path

### LINKED ISSUE #6732: fix(deps): resolve the remaining root go_modules blockers (vcluster checksum, cel-go module path)

## Evidence

The root `go_modules` update job has failed on every run since 2026-08-24 (`08-24`, `08-25`, `08-26`, `08-27`); last green was `2026-08-20`. The newest run
([33097225356](https://github.com/devantler-tech/ksail/actions/runs/33097225356), 2026-08-27T17:42Z) reports **four** errors:

| module | error |
|---|---|
| `github.com/siderolabs/talos` | `dependency_file_not_resolvable` — `runc/libcontainer/user` |
| `github.com/siderolabs/image-factory` | `dependency_file_not_resolvable` — `runc/libcontainer/user` |
| `github.com/loft-sh/vcluster` | `dependency_file_not_resolvable` — checksum mismatch |
| `github.com/google/cel-go` | `go_module_path_mismatch` |

devantler-tech/ksail#6731 fixes the first two (the `runc` cause). **This issue covers the remaining two**, which are independent and will keep the job red on their own.

### 1. `loft-sh/vcluster` — checksum mismatch

```
go: github.com/loft-sh/vcluster@v0.36.1 requires
	github.com/loft-sh/apiserver@v0.0.0-20260707184419-aef558a5ae8d: verifying go.mod: checksum mismatch
```

A checksum mismatch is either a legitimately changed upstream pseudo-version or a `GONOSUMDB`/`GOFLAGS` gap. Determine which before changing anything — a mismatch is also what a tampered module looks like, so do not add a blanket `GONOSUMCHECK`/`GOFLAGS=-mod=mod` workaround to make it pass.

### 2. `google/cel-go` — module path mismatch

`go_module_path_mismatch` means the module declares a path different from the one requested, usually a major-version retag (`/v2`) or a repository move.

All root Go dependency updates, including security patches. CI stays green throughout, so this is invisible without reading the Dependabot logs — the same blind spot #6666 documents.

## Acceptance criteria

- [ ] The `vcluster` checksum mismatch is diagnosed as legitimate-or-not and resolved without weakening module verification.
- [ ] The `cel-go` module path mismatch is resolved.
- [ ] A root `go_modules` Dependabot run completes with **zero** errors.

Size: small–medium (two independent dependency fixes).

Related: #6666, #6731. Possibly-stale prior report of the same job failing for a different reason: #6607.

devantler:

### Why

Automatic dependency updates for KSail's main Go module have been **completely dead since 2026-08-24** — every daily run has failed, so no Go dependency update has landed in six days, security patches included. CI stays green throughout, so this is invisible without reading the Dependabot logs.

The cause is that one library, `cel-go`, moved to a new home and renamed itself. The old name simply cannot be resolved any more, and that single unresolvable name aborts the whole update job for every other dependency with it.

### What

Point KSail at the library's new name. The update job stops hitting the dead name, so the other ~100 Go dependencies can flow again.

The old name stays in the dependency graph indirectly, because around twenty upstream projects (Kubernetes, Cilium, Talos, vCluster) have not moved yet — that is expected and harmless, and it does not re-block the job.

Also covers the desktop module, which shares the main module's code and would otherwise stop building.

Fixes #6732

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -222,9 +222,10 @@ linters:
             - google.golang.org/grpc
             - google.golang.org/protobuf
             # cel-go powers the native CEL rule validation of rendered GitOps
-            # manifests in pkg/svc/gitops/celrules (#5693, epic #5344). Already
+            # manifests in pkg/svc/gitops/celrules (#5693, epic #5344). The module
+            # renamed its path from github.com/google/cel-go in v0.32.0. Already
             # linked transitively via the Kubernetes apiserver stack.
-            - github.com/google/cel-go
+            - cel.dev/cel-go
             - gopkg.in/yaml.v3
             - helm.sh/helm
             - k8s.io
--- a/desktop/go.mod
+++ b/desktop/go.mod
@@ -26,6 +26,7 @@ require (
 
 require (
 	al.essio.dev/pkg/shellescape v1.6.0 // indirect
+	cel.dev/cel-go v0.32.0 // indirect
 	cel.dev/expr v0.25.2 // indirect
 	cloud.google.com/go v0.123.0 // indirect
 	cloud.google.com/go/auth v0.22.0 // indirect
--- a/desktop/go.sum
+++ b/desktop/go.sum
@@ -2,6 +2,8 @@ al.essio.dev/pkg/shellescape v1.6.0 h1:NxFcEqzFSEVCGN2yq7Huv/9hyCEGVa/TncnOOBBeX
 al.essio.dev/pkg/shellescape v1.6.0/go.mod h1:6sIqp7X2P6mThCQ7twERpZTuigpr6KbZWtls1U8I890=
 c2sp.org/CCTV/age v0.0.0-20251208015420-e9274a7bdbfd h1:ZLsPO6WdZ5zatV4UfVpr7oAwLGRZ+sebTUruuM4Ra3M=
 c2sp.org/CCTV/age v0.0.0-20251208015420-e9274a7bdbfd/go.mod h1:SrHC2C7r5GkDk8R+NFVzYy/sdj0Ypg9htaPXQq5Cqeo=
+cel.dev/cel-go v0.32.0 h1:irvpFKr5EuGPyxeME03ERh0rii1TX+BDAnB9eL3IvNk=
+cel.dev/cel-go v0.32.0/go.mod h1:DnVip7tpJSsgZymwfT+m1tnEVy3ivAjSMXPx12YrMkU=
 cel.dev/expr v0.25.2 h1:K6j46C81hXtZQfuX60cVWQFBJahKSE2gfRbNuvr5bFs=
 cel.dev/expr v0.25.2/go.mod h1:hrXvqGP6G6gyx8UAHSHJ5RGk//1Oj5nXQ2NI02Nrsg4=
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 )
 
 require (
+	cel.dev/cel-go v0.32.0
 	cloud.google.com/go/container v1.53.0
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.22.0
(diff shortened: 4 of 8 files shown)


devantler:

## Root cause, proof and user evaluation

### Root cause (established from the module proxy, with a control)

| version | declared module path |
|---|---|
| `github.com/google/cel-go@v0.31.0` | `module github.com/google/cel-go` — resolvable (current pin) |
| `github.com/google/cel-go@v0.32.0` | `module cel.dev/cel-go` — **mismatch** |

Both paths list an identical version set at the same commit `f2039bc647bca407d882d90436fc8b91bab1ae62`; the repo moved to `cel-expr/cel-go` and the canonical path is now `cel.dev/cel-go`.

### RED — reproduced the exact CI failure locally, before the fix

```
go: github.com/google/cel-go@v0.32.0 requires github.com/google/cel-go@v0.32.0: parsing go.mod:
	module declares its path as: cel.dev/cel-go
	        but was required as: github.com/google/cel-go
```

Byte-identical to the `go_module_path_mismatch` in run [33193614934](https://github.com/devantler-tech/ksail/actions/runs/33193614934).

### GREEN

- `go get cel.dev/cel-go@v0.32.0` → rc 0
- `go build ./...` → rc 0
- `go test` on all three affected packages → rc 0 (`gitops/celrules`, `client/celrules`, `provisioner/cluster/k3d`)
- `gofmt -l` clean on all three changed Go files

### Ablation — fired for the right reason

Requiring the **old** path at v0.32.0 still fails with the identical mismatch message (rc 1), and left `go.mod` byte-unchanged. This proves the mismatch is a property of the module rather than an artifact of the local environment.

My first ablation attempt used `go list -m -json`, which returned rc 0 — it resolves version metadata without ever checking the declared path, so it exercised nothing. Recording that because the wrong probe looked like a passing control.

### User evaluation — exercised the real engine, both states

The `celrules` packages are not CLI-wired yet, so I drove the client API directly against real manifests and observed the effect:

```
COMPLIANT: violations=0 hasErrors=false
VIOLATING: violations=1 hasErrors=true
    -> error: rule/require-resource-limits Deployment/demo/bad: evaluation error: no such key: resources
```

CEL compiles and evaluates correctly under the new module path, with violations still attributed to the right resource. The v0.31.0 → v0.32.0 bump carries no API breakage.

### Scope note on the acceptance criteria

AC2 (`cel-go` path mismatch resolved) is proven above. AC1 (vcluster checksum) was already satisfied before this change — the 08-28 run reports **one** error, down from the four in the issue body.

AC3 ("a root `go_modules` run completes with zero errors") is **only provable by the next scheduled Dependabot run**, which is post-merge. Reasoning for why it should now pass: the failure was on a *direct* dependency, `.github/dependabot.yaml` declares no `allow:` block, and Dependabot's Go default is direct-only — so the now-indirect old path should no longer be attempted. That is an argument, not a measurement, and I will verify it against the real run rather than close the issue on it.

devantler:



@coderabbitai review

devantler:

## Readiness at the current head `beba93ac`

The evaluation above was recorded at the previous head; the only change since is the depguard
config fix. Re-confirming all three conditions hold at `beba93acad23f425d949e0a28a8389e973d142e1`:

1. **Programmatically tested** — 61 SUCCESS / 17 SKIPPED, 0 failing, 0 pending. The jobs that
   actually *consume* these changes ran rather than skipping: `🧹 Lint - golangci-lint` (the exact
   depguard consumer, and the check the config fix targets), `🏗️ Build`, `🧪 Test`,
   `📊 Code Coverage`, and the full `🧪 System Test (Docker)` matrix across KWOK / Vanilla / K3s.
2. **Reviewed** — `cr@beba93ac`. CodeRabbit's auto-generated summary, refreshed after the
   authenticated request, names the full head and reports *"No actionable comments were generated
   in the recent review."* All 20 collapsed sections enumerated; the only finding-shaped one is
   `🔇 Additional comments (7)`, the explicitly non-actionable section. No refusal marker.
3. **Tried and evaluated as a user** — the cel engine was driven directly against real manifests in
   both states (recorded above); the depguard change is lint configuration whose observable effect
   is golangci-lint's verdict, which is green in a job that ran at this head.

Threads: 0. `mergeStateStatus`: CLEAN. Promoting and merging.

Issue #6816, opened 2026-09-01
fix(deps): grouped gomod updates skip the desktop module, so the talos group PR fails before reaching the real bump


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Dependabot's **grouped** gomod updates in this repository update only the root module, while
**ungrouped** ones update both modules. Measured across six PRs on 2026-09-01:

| PR | grouped? | files changed |
|---|---|---|
| [#6797](https://github.com/devantler-tech/ksail/pull/6797) `golang.org/x/mod` (merged) | no | `go.mod` `go.sum` **`desktop/go.mod` `desktop/go.sum`** |
| [#6798](https://github.com/devantler-tech/ksail/pull/6798) `azcore` (merged) | no | `go.mod` `go.sum` **`desktop/go.mod` `desktop/go.sum`** |
| [#6799](https://github.com/devantler-tech/ksail/pull/6799) `sigstore-go` (merged) | no | `go.mod` `go.sum` **`desktop/go.mod` `desktop/go.sum`** |
| [#6724](https://github.com/devantler-tech/ksail/pull/6724) `omni/client` | no | `go.mod` `go.sum` **`desktop/go.mod` `desktop/go.sum`** |
| [#6796](https://github.com/devantler-tech/ksail/pull/6796) `machinery` | no | `go.mod` `go.sum` **`desktop/go.mod` `desktop/go.sum`** |
| [#6813](https://github.com/devantler-tech/ksail/pull/6813) **talos group** | **yes** | `go.mod` `go.sum` — **`desktop/` NOT updated** |

The consequence is a build failure that never reaches the real dependency question. `#6813` @
`da13b012c7`, job `99668418073` (`🏗️ Build (Linux)`), failing step `go build -o ksail-desktop .`:

```
go: updates to go.mod needed; to update it:
	go mod tidy
```

`desktop/go.mod` carries the root module's transitive requirements as `// indirect` (it declares
`replace github.com/devantler-tech/ksail/v7 => ../`, and at `da13b012c7` still pins
`github.com/siderolabs/omni/client v1.9.1 // indirect`). When the root module moves and `desktop/`
does not, the desktop build is stale by construction.

Config context: `.github/dependabot.yaml` has exactly **one** `gomod` entry (`directory: /`), there is
**no** `/desktop` entry, and there is **no `go.work`**. So the second module is not independently
configured — yet ungrouped updates reach it and grouped ones do not.

## Why this matters

[...]

PR #6822, opened 2026-09-01
fix(desktop): mirror the parent module's runc replace

devantler:

### Why

The desktop app is a separate Go module, so it has to keep its own copy of the compatibility
pins the main module uses — Go does not share them across modules. One of those pins was
missing from the copy, which makes dependency updates fail while the main build stays green.
That is why it went unnoticed: it only bites when an update shifts which dependency versions
get picked, and then it blocks the update rather than the product.

### What

Adds the missing pin to the desktop module so its dependency housekeeping resolves again, with
a note explaining why it has to be duplicated. No dependency versions actually change — the
lockfile is untouched — so this carries no product risk and only removes a recurring blocker
from dependency updates.

Verified by reproducing the failure on the currently-stuck talos update, confirming the fix
clears it, and confirming the desktop app still builds.

Part of the same blocker chain as #6816 — that issue makes grouped updates refresh the
desktop module; this makes that refresh actually succeed. Neither alone is sufficient.

### DIFF
--- a/desktop/go.mod
+++ b/desktop/go.mod
@@ -17,6 +17,13 @@ replace (
 	github.com/loft-sh/log => github.com/devantler/log v0.0.0-20260407144227-16cc61ebdb79
 	// Mirrors the parent module's local v0.3.0 compatibility module.
 	github.com/moby/go-archive => ../third_party/go-archive
+
+	// dockertest imports runc's libcontainer/user, removed in runc v1.4.0, through the tests
+	// of a dependency (sops/v3/hcvault.test). `go mod tidy` resolves that graph, so without
+	// this pin it falls back to runc@latest, which no longer provides the package, and tidy
+	// fails. Mirrors the parent module's replace, where the full rationale and its security
+	// trade-off are recorded.
+	github.com/opencontainers/runc => github.com/opencontainers/runc v1.3.6
 )
 
 require (


devantler:



@coderabbitai review

devantler:

**User evaluation at `993fdfb99d24b39a961d6ef2627d4d7dd390e84e`** — exercised against the real stuck
update rather than a fixture, since this change exists to unblock one.

**RED, on `#6813`'s head `1a74f372`** (the currently-failing talos group PR) — ran the exact command
CI's `🧩 Verify Desktop Module Tidy` runs:

```
go: finding module for package github.com/opencontainers/runc/libcontainer/user
... github.com/opencontainers/runc/libcontainer/user: module github.com/opencontainers/runc@latest
found (v1.5.1), but does not contain package .../libcontainer/user
```

**GREEN, same head, with this change applied:** `go mod tidy` exits **0**.

**No-op on `main`, which is the property that makes it safe to land now:** applying it to `main`
produces a **7-line `desktop/go.mod` change with zero `go.sum` churn** — so no dependency version
actually moves — `go mod tidy` stays exit 0 and is idempotent (a second run produces no further diff,
which is exactly what the CI check asserts), and `go build ./...` in `desktop/` exits **0**.

**Judged as the check's user:** CI's `🧩 Verify Desktop Module Tidy` **passed on this PR** (1m5s), so
the consuming job actually ran rather than being filtered out — the green is not vacuous. 19 checks
pass, 0 failing.

**Why the mirror is the right fix rather than a pin here:** the parent module already carries this
replace with its rationale and security trade-off recorded, and `desktop/go.mod` states that it
duplicates the parent's replaces because Go does not inherit them. Comparing the two replace sets
programmatically, `runc` was the **only** difference. This restores the stated invariant rather than
introducing a new constraint.

devantler:

**Promoting — all three readiness conditions hold at `993fdfb99d24b39a961d6ef2627d4d7dd390e84e`.**

1. **Programmatically tested** — 52 checks, 0 failing; `CLEAN`; 0 review threads, 0 unresolved;
   `🧩 Verify Desktop Module Tidy` (the check this fixes) **passed**, so the consuming job ran.
2. **Reviewed** — CodeRabbit green at this exact head. Authenticated deliberately, because there is
   **no substantive review object** here: the satisfier is the auto-generated summary, updated
   09:35:48Z (after the 09:30:59Z request), naming the **full head SHA**, "No actionable comments were
   generated"; head status `Review completed`; and **zero** rate-limit/service markers in any
   CodeRabbit comment. Keying on review objects alone would have misread this real green as none.
3. **Tried and evaluated as a user** — recorded at this head (RED on `#6813`'s head → GREEN with the
   fix; no-op on `main` with zero `go.sum` churn; desktop builds exit 0).

Issue #6767, opened 2026-08-29
Intermittent CodeQL `Analyze (go)` Autobuild runner termination blocks merges, and GitHub refuses to re-run it


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Measured 2026-08-29 03:36Z across every open PR, with a `main` control:

| PR | `Analyze (go)` |
|---|---|
| #6742 | success |
| #6764 | **failure** |
| #6765 | **failure** |
| #6766 | success |
| `main`, 5 newest settled runs | success ×5 |

So it is **intermittent, not systemic** — same code, both outcomes. The failing job's annotations are
cgo extraction errors, not analysis findings:

```
could not import C (no metadata for C)
cannot convert w.window (variable of Pointer type pointer) to type windowPointer
```

## Why this blocks merges

`main` carries a **`code_scanning` ruleset rule** (`CodeQL: alerts=all, security=all`) in addition to
`required_status_checks` (`CI - Required Checks`). A run that fails to produce Go results therefore
gates the merge even though `Analyze (go)` is **not** in the required-status-checks list. That
distinction is easy to miss: reading only `required_status_checks` suggests CodeQL is advisory, and it
is not.

Both affected PRs sit `BLOCKED` with every other check green — including #6764, a **release** PR with
auto-merge already armed, so this stalls the release train rather than one feature.

## Why it does not clear itself

The analysis is GitHub-**managed** (`event: dynamic`, `path: dynamic/github-code-scanning/codeql`), so
there is no workflow file in the repository and `gh run rerun` is refused outright:

```
This workflow run cannot be retried
```

The only way to get a fresh analysis is a **new head commit**. For a bot-authored release PR nothing
will produce one, so it can sit blocked indefinitely.

## Expected

An intermittent extractor failure should not permanently block a merge, and recovering from one should
not require inventing a commit.

## Acceptance criteria

- [ ] Root-cause the cgo extraction failure (which package pulls cgo into the Go analysis, and whether
      it can be excluded from CodeQL's build without losing coverage of first-party code).
- [ ] Either stabilise the analysis, or move off default setup to an advanced-setup workflow so the run
      lives in this repository and **can** be re-run.
- [ ] A blocked PR has a documented recovery path that does not depend on an empty commit.
- [ ] Confirm the intended `code_scanning` rule behaviour is what we want for an extractor *error*
      (as opposed to a real alert).

## Size

[...]

PR #6850, opened 2026-09-03
ci(codeql): replace managed default setup with re-runnable advanced setup

devantler:

### Why

Roughly one PR in five gets permanently blocked by a CodeQL run that nobody can retry.

CodeQL runs here through GitHub's *managed* default setup, which means the run lives outside this
repository. When its runner is preempted mid-analysis, GitHub refuses to re-run it — and the
`code_scanning` rule on `main` then blocks the pull request with no way out except pushing a new
commit. A bot-authored release PR never gets one, so it can sit blocked indefinitely. #6849 is
blocked by exactly this right now, and it only changes four version strings.

Measured on this repository: **11 of the 60 most recent CodeQL runs failed (18.3%)**, including two
back-to-back failures on `main` on 2026-09-02 that cleared only by chance.

### What

Moves CodeQL into a workflow this repository owns, so a preempted run can simply be re-run.

The analysis itself is unchanged — same query suite, same languages, no exclusions — so security
coverage is identical. The structure and pinned versions follow the equivalent workflow already
running green in `devantler-tech/actions`.

**This cannot merge until CodeQL default setup is switched off in this repository's settings**
(*Settings → Code security → Code scanning → Default setup → Disable*). The two setups are mutually
exclusive, so while default setup is on, this workflow cannot upload its results. That flip is a
repository security setting, so it is yours to make rather than mine — and it is the one thing this
PR needs from you.

Refs #6767

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,11 @@
+name: "CodeQL config — devantler-tech/ksail"
+
+# Mirrors what CodeQL default setup ran for this repository, with NO query
+# exclusions: default setup was configured with query_suite "extended", which is
+# the `security-extended` suite below.
+#
+# Default setup also declared threat_model "remote". That is CodeQL's own
+# default threat model, so it needs no explicit key here — stating it would
+# change nothing about which sources are treated as tainted.
+queries:
+  - uses: security-extended
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -26,7 +26,11 @@ on:
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
-  cancel-in-progress: true
+  # Only ever cancel superseded PR runs. Cancelling a run on main, in the merge
+  # queue, or on the weekly schedule would leave that ref with no uploaded
+  # results, which is the exact "code scanning has no results" block this
+  # workflow exists to remove.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions: {}
 
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,90 @@
+name: 🔍 CodeQL
+
+# Advanced setup (replaces CodeQL default setup), running the same extended
+# suite over the same languages — see .github/codeql/codeql-config.yml.
+#
+# Why this exists rather than default setup: a managed default-setup run lives
+# outside this repository, so when its runner is preempted mid-analysis GitHub
+# refuses to re-run it ("This workflow run cannot be retried", HTTP 403). The
+# `code_scanning` ruleset on `main` then blocks the PR with no recovery short of
+# inventing a new head commit — which a bot-authored release PR will never
+# produce. Owning the workflow here makes a preempted run re-runnable. See
+# #6767.
+#
+# CodeQL default setup must stay DISABLED for this to run: the two are mutually
+# exclusive, and advanced setup cannot upload results while default setup is
+# configured.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  merge_group:
+  schedule:
+    - cron: "30 1 * * 1" # weekly, mirroring the former default-setup schedule
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      contents: read
+      actions: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Default setup listed actions, go, javascript, javascript-typescript
+          # and typescript. `javascript`, `typescript` and `javascript-typescript`
+          # are the same CodeQL analysis, so advanced setup expresses them once.
+          - language: actions
+            build-mode: none
+          - language: go
+            build-mode: autobuild
+          - language: javascript-typescript
+            build-mode: none
+    steps:
+      - name: 🛡️ Harden runner
+        uses: step-security/harden-runner@05e31511f85b41b11d1cf0ef85d0992719546e2c # v2.21.0
+        with:
+          egress-policy: audit
+
+      - name: 📑 Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: 🐹 Set up Go
+        if: matrix.language == 'go'
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: 🔍 Initialize CodeQL
+        uses: github/codeql-action/init@c35d1b164463ee62a100735382aaaa525c5d3496 # codeql-bundle-v2.25.6
+        with:
+          languages: ${{ matrix.language }}
+          build-mode: ${{ matrix.build-mode }}
+          # Emit BOTH code-scanning (security) AND code-quality results — matching
+          # what default setup produced — so this repository's "Require code
+          # scanning results" and "Require code quality results" rulesets are both
+          # satisfied. analysis-kinds is flagged internal/subject-to-change in
+          # codeql-action, but it is the only way to produce code-quality from
+          # advanced setup; safe here because the action is SHA-pinned above and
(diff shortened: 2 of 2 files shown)


devantler:

## Evaluation record (readiness — NOT yet met)

**This PR is deliberately staying a draft.** It cannot reach a green, meaningful CI state until CodeQL
default setup is disabled in repository settings, because the two setups are mutually exclusive and
advanced setup cannot upload results while default setup is configured. That flip is a repository
security setting, so it is the maintainer's to make.

Visible on this PR right now: each `Analyze (…)` check appears **twice** — once from default setup's
managed run, once from this workflow — which is the conflict itself, rendered.

### What I did verify

- `actionlint` clean (exit 0) on the workflow.
- **Negative control**: ablating `${{ matrix.build-mode }}` to a nonexistent matrix key makes actionlint
  exit 1 with `property "nonexistent-key" is not defined in object type {build-mode: string; language: string}`.
  The inferred type confirms `build-mode` resolves as a **property**, not as a subtraction expression —
  the thing most likely to be subtly wrong here.
- `zizmor` (workflow security scan) — success on this head.
- `CI - Required Checks` — success.
- Config and workflow both parse (`yq`).
- Structure, SHA pins and the `analysis-kinds: code-scanning,code-quality` input are taken from
  `devantler-tech/actions`' `active-codeql.yaml`, which is green in production; that repository's default
  setup reads `not-configured`, which is the control for the mutual-exclusivity claim above.

### What I could not verify, and why

That the analysis actually uploads and satisfies the `code_scanning` / code-quality rulesets. That is
unobservable until the settings flip, so it is the first thing to check afterwards.

### Self-review finding, already fixed in `6b1e1da`

The first commit set `cancel-in-progress: true`. On `main`, in the merge queue, or on the weekly
schedule that would cancel an in-flight analysis and leave the ref with **no uploaded results** —
reintroducing exactly the block this PR removes. Now scoped to `pull_request` runs only.

devantler:

## The blocker is real, but the design behind this PR does not survive it — do not flip default setup on the strength of this branch

Re-verified live at head `6b1e1da1`. All three `Analyze (*)` jobs fail, and the job log gives one
root cause for every one of them:

```
Code Scanning could not process the submitted SARIF file:
CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
```

So this PR's CI **cannot** go green while default setup is on. That part is expected and is the
chicken-and-egg this PR always had. What is new is the second thing the same log says.

### `analysis-kinds: code-scanning,code-quality` is already a no-op, and is on its way to being fatal

```
The `analysis-kinds` input is experimental and for GitHub-internal use only. An analysis kind other
than `code-scanning` was specified in a custom workflow. This is not supported and will become a
fatal error in a future version of the CodeQL Action.

Specifying multiple values as input is no longer supported.
Continuing with only `analysis-kinds: code-scanning`.
```

The workflow comment claims `analysis-kinds` "is the only way to produce code-quality from advanced
setup". The action now says the opposite twice over: a custom workflow may only do `code-scanning`,
and the multi-value form is **already being dropped**. The code-quality half of this workflow does
nothing today and is documented to become a hard failure.

### Why that matters more than this PR

`devantler-tech/ksail` has an **active `code_quality` ruleset** on `main` (ruleset `16885082`,
`severity: all`) alongside the `code_scanning` one (`6455804`). Every CodeQL analysis in this
repository today comes from the managed default setup — `gh api repos/devantler-tech/ksail/code-scanning/analyses`
returns `analysis_key: dynamic/github-code-scanning/codeql:analyze` for all of them. Default setup is
therefore the only producer of code-quality results in the repo.

Disabling it, as this PR requires, removes that producer while advanced setup provably cannot replace
it. That lands in one of two states, and I have deliberately **not** established which:

1. the `code_quality` ruleset finds no results and **blocks every PR in ksail**, or
2. it passes vacuously and the code-quality gate is **silently switched off**.

Both are materially worse than the intermittent problem this PR set out to fix, and neither is in its
stated design. Determining which one it is needs the flip itself, which is precisely the action that
should not be taken speculatively.

### What is confirmed about the original problem (#6767)

The underlying complaint is real and I re-measured it this run:

- `ksail#6849` (`chore(copilot-plugin): release v7.181.8`, auto-merge armed by `ksail-bot` at
  03:02:55Z) has been stuck ~8h. Its only required status check, `CI - Required Checks`, is
  **SUCCESS**; it has **0** unresolved threads; `mergeable: MERGEABLE`. What blocks it is the
  `code_scanning` ruleset, because its managed CodeQL run (`33709911261`, `event: dynamic`) failed
  once and managed runs return 403 on re-run.
- The failures are **intermittent, not a streak** — managed CodeQL on `main` failed at 18:39Z and
  19:06Z on 09-02 and recovered by 19:24Z. A PR only gets *permanently* stuck when it will never push
  again, which is exactly the bot-authored release case.

So the problem statement in #6767 holds. The remedy in this PR does not.

### Suggested next step

Reframe #6767 around remedies that keep default setup: making a stuck bot PR recoverable without a new
head commit, or narrowing the `code_scanning` ruleset so a *failed analysis* is distinguishable from a
*failing analysis*. I have not picked one — that is a real design decision and it should be made on
this evidence rather than inherited from this branch.

Leaving this PR as a draft, parked on that decision rather than on the settings flip.

devantler:

**Correction to the comment above — one sentence overstated its evidence.**

I wrote that "default setup is therefore the only producer of code-quality results in the repo". I
cannot establish that, and I should not have asserted it. GitHub exposes no analysis-kind anywhere I
can read: `repos/devantler-tech/ksail/code-scanning/analyses` has no `analysis_kind` field, and
`repos/devantler-tech/ksail/code-scanning/default-setup` returns only
`languages, query_suite, runner_label, runner_type, schedule, state, threat_model, updated_at`.

What is actually established:

- the `code_quality` ruleset (`16885082`, `severity: all`) is **active** on `main`;
- every CodeQL analysis the API does expose comes from managed default setup
  (`analysis_key: dynamic/github-code-scanning/codeql:analyze`);
- a custom/advanced workflow is refused any analysis kind other than `code-scanning`, per the action's
  own error, and already drops the multi-value form.

So whether default setup is today satisfying that `code_quality` rule, or the rule is passing
vacuously because nothing produces code-quality results at all, is **not determinable from outside**
— which is exactly why the two-state disjunction in the comment above was left open, and why the flip
should not be made speculatively.

That does not change the conclusion, and it removes one thing from it: the risk is that disabling
default setup either blocks every PR or silently drops a quality gate, and **which** of those it is
still has to be discovered by doing it. The `analysis-kinds` finding — that the code-quality half of
this workflow is already a no-op and is documented to become fatal — stands unchanged and rests only
on the action's own log output.

devantler:

## Closing: this approach cannot reach green, and the step that would unblock it is harmful

Re-verified live against `ksail` today (2026-09-03) at head `6b1e1da1`, rather than inherited from the
earlier run that first spotted it.

### 1. The CI is structurally ungreenable, not flaky

All three `Analyze` jobs (`actions`, `go`, `javascript-typescript`) fail with one root cause:

```
CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
```

`code-scanning/default-setup` reports `state: configured`. So this PR cannot go green until default
setup is disabled — and disabling it is a repository-settings change only the maintainer can make.
That is a chicken-and-egg: the PR exists to justify the flip, and cannot demonstrate anything until
after it.

### 2. The flip itself is the harmful part

`ksail` has **two** active rulesets here, not one:

| Ruleset | ID | State |
|---|---|---|
| Require code scanning results | `6455804` | active |
| Require code quality results | `16885082` | active |

Advanced setup would have to produce **both** kinds. It cannot: this PR's own run log shows the
`analysis-kinds` input already degrading —

```
analysis-kinds: code-scanning,code-quality
  → multiple values no longer supported. Continuing with only `analysis-kinds: code-scanning`
  → `analysis-kinds` input is experimental and for GitHub-internal use only.
```

So after the flip, `Require code quality results` has no producer. Either every PR blocks on a result
nothing emits, or the quality gate silently stops gating. **Both outcomes are worse than the
intermittent failure this PR set out to fix**, and the second is worse than it looks, because a gate
that silently stops gating reads as healthy.

I am deliberately **not** escalating the settings flip to the maintainer. The flip is not a blocked
step waiting on authority — it is a step that should not be taken.

### 3. What I am not claiming

I previously asserted that default setup is the only thing that can produce code-quality results. I
could not prove that and have withdrawn it: the API exposes no `analysis_kind` field anywhere
(`analyses` has none, and `default-setup` returns only languages, query suite, runner, schedule, state,
threat model, and updated-at). The defensible claim is the narrower one above — advanced setup as
configured here demonstrably cannot emit the second kind, so the ruleset loses its producer.

### Disposition

Closing as **not the right remedy**. No work is lost: the underlying problem stays open on #6767,
which I am reframing toward remedies that keep default setup enabled. Nothing here was a
partial fix worth salvaging — the diff's entire content is the advanced-setup workflow whose premise
this evidence refutes.

PR #6851, opened 2026-09-03
ci: leave Dependabot branches to Dependabot so bumps can rebase themselves

### LINKED ISSUE #6832: ksail-bot's sync commit permanently disables Dependabot rebase, stranding ~half of dependency PRs

## Evidence

`ksail-bot` pushes a `chore: sync modules and update generated files` commit onto Dependabot's own
branches. That single push permanently forfeits Dependabot's ability to rebase the PR:

> Looks like this PR has been edited by someone other than Dependabot. That means Dependabot can't
> rebase it - sorry!

Observed live on both open Dependabot PRs, 2026-09-01:

| PR | sync commit | state |
|---|---|---|
| #6830 | `ksail-bot` 17:49:13Z | `DIRTY`, checks green, rebase refused 21:03:18Z |
| #6826 | `ksail-bot` 17:31:36Z | `DIRTY`, 38 checks failing, rebase refused 20:03:42Z |

Sampling the 12 most recent Dependabot PRs, **7 carry the sync commit** (#6814, #6815, #6826, #6827,
#6828, #6829, #6830) and 5 do not (#6812, #6813, #6824, #6825, #6831).

## Impact

A Dependabot PR that has been synced cannot recover from base drift on its own. As soon as any other
PR merges to `main`, it goes `DIRTY` and stays there — auto-merge is armed but can never fire, and
the ordinary `@dependabot rebase` remedy is refused. It then needs manual agent intervention every
time.

This is what produced the current queue stall: #6827 merged at 20:26Z and immediately stranded #6830,
which was otherwise green on all 77 checks. The same mechanism previously stranded #6813, whose
closure recreated it as #6826 — so the failure mode also generates PR churn, because closing is the
intuitive-but-wrong remedy (`@dependabot recreate` rebuilds in place; closing opens a new number).

Because roughly half of Dependabot PRs are affected and the ecosystem is capped at an open-PR limit,
stranded PRs occupy that cap and throttle dependency intake generally.

## Expected behaviour

A routine dependency bump should reach merge without agent intervention. Either the generated-file
sync should not land on the Dependabot branch, or it should land in a way that preserves Dependabot's
ownership of the branch.

Options worth weighing:

- **Move the sync off the PR branch.** Let the bump merge as Dependabot authored it and run the
  generated-file sync on `main` afterwards, as its own commit or PR.
- **Make the sync idempotent and re-appliable**, and standardise on `@dependabot recreate` when the
  branch goes stale, accepting that the sync is rebuilt each time.
- **Gate the sync** so it only runs when the bump actually changes generated output, which would cut
  the affected share substantially.

## Acceptance criteria

- [ ] A Dependabot PR whose base has moved can recover without an agent comment — either it rebases
      itself, or the repository's automation reliably rebuilds it.
- [ ] Generated files stay correct on `main` after such a bump merges.
- [ ] The behaviour is verified on a real bump, not only reasoned about: a PR is observed going stale
      and then reaching merge with no agent intervention.

Rough size: M. Worth a short investigation first to decide which of the three options fits the
repository's release and codegen constraints.

devantler:

### Motivation

Dependency bumps keep getting stuck and needing a person to nudge them. GitHub takes
Dependabot's rebase ability away permanently as soon as anything else pushes to its
branch — and our CI was pushing a generated-file sync onto those branches. About half
of all Dependabot PRs were affected: as soon as any other PR merged, they fell behind,
auto-merge could never fire, and the usual "rebase this" remedy was refused. Because
the dependency queue has a cap on how many PRs can be open at once, the stuck ones sat
in that cap and throttled dependency intake generally.

Three PRs are stranded by this right now (#6839, #6845, #6826).

### Change

CI no longer pushes the generated-file sync onto Dependabot's own branches, so
Dependabot keeps ownership and can rebase itself. The generated changes are not lost —
they land through the path we already have, which opens a sync PR once the bump has
merged.

Picks option 1 of the three the issue put forward. It is the only one that removes the
cause rather than reducing how often it bites, and the machinery it depends on already
exists.

Fixes #6832

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -760,7 +760,9 @@ jobs:
       # modes are not expressible. Reject anything other than a regular 0644 file below
       # rather than dereferencing a symlink or silently losing an executable bit.
       - name: 📤 Commit and push generated changes (PR branch)
-        if: github.event_name == 'pull_request'
+        if: >-
+          github.event_name == 'pull_request'
+          && github.event.pull_request.user.login != 'dependabot[bot]'
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
           BRANCH: ${{ github.head_ref }}
--- a/internal/ciharness/ci_workflow_test.go
+++ b/internal/ciharness/ci_workflow_test.go
@@ -311,8 +311,6 @@ func TestRunsOnDefaultBranch(t *testing.T) {
 
 // autoCommitWorkflow captures the auto-commit job's step surface, which decides
 // whether CI pushes a generated-file commit onto a pull request's own branch.
-//
-//nolint:tagliatelle // GitHub Actions defines these external keys in kebab-case.
 type autoCommitWorkflow struct {
 	Jobs map[string]struct {
 		Steps []map[string]any `yaml:"steps"`
--- a/internal/ciharness/ci_workflow_test.go
+++ b/internal/ciharness/ci_workflow_test.go
@@ -308,3 +308,93 @@ func TestRunsOnDefaultBranch(t *testing.T) {
 		})
 	}
 }
+
+// autoCommitWorkflow captures the auto-commit job's step surface, which decides
+// whether CI pushes a generated-file commit onto a pull request's own branch.
+//
+//nolint:tagliatelle // GitHub Actions defines these external keys in kebab-case.
+type autoCommitWorkflow struct {
+	Jobs map[string]struct {
+		Steps []map[string]any `yaml:"steps"`
+	} `yaml:"jobs"`
+}
+
+// The exact conditions the two auto-commit delivery paths must carry.
+//
+// The PR-branch push is the one that has to exclude Dependabot; the protected-branch
+// path is asserted alongside it because it is where a skipped Dependabot sync is
+// expected to land instead, and silently losing it would turn this fix into a
+// generated-file correctness regression on the default branch.
+const (
+	approvedAutoCommitPRBranchCondition = "github.event_name == 'pull_request'" +
+		" && github.event.pull_request.user.login != 'dependabot[bot]'"
+	approvedAutoCommitProtectedBranchCondition = "github.event_name != 'pull_request'"
+)
+
+// stepCondition returns the `if:` of the uniquely-named step in the given job.
+//
+// It requires exactly one match rather than taking the first: two steps sharing a
+// name would make the assertion below silently describe whichever came first.
+func stepCondition(t *testing.T, job []map[string]any, stepName string) string {
+	t.Helper()
+
+	var found []map[string]any
+
+	for _, step := range job {
+		if name, ok := step["name"].(string); ok && name == stepName {
+			found = append(found, step)
+		}
+	}
+
+	require.Lenf(t, found, 1, "expected exactly one step named %q", stepName)
+
+	condition, ok := found[0]["if"].(string)
+	require.Truef(t, ok, "step %q must carry a string `if:` condition", stepName)
+
+	return condition
+}
+
+// Pins Dependabot's ownership of its own branches.
+//
+// GitHub permanently revokes Dependabot's ability to rebase a pull request once any
+// other actor pushes to its branch ("this PR has been edited by someone other than
+// Dependabot"). CI's generated-file sync used to push onto every same-repo PR branch,
+// Dependabot's included, so a synced bump could never recover from base drift: the
+// moment anything else merged it went behind base, auto-merge could never fire, and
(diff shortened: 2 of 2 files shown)


devantler:

## What I exercised (readiness record — NOT yet met)

**RED/GREEN.** The contract test fails on `main` for exactly the intended reason —
`github.event_name == 'pull_request'` vs the expected condition — and the
protected-branch assertion passed in the same RED run, so the failure isolates the one
thing this change alters. Green after the fix, with the whole `internal/ciharness`
suite (`-count=1`, which is load-bearing here: the subject is a YAML file the Go build
graph cannot see).

**The discriminator actually discriminates.** The condition is a no-op unless the login
spelling is exact, so I read it off the live payload rather than assuming:
`user.login` is `dependabot[bot]` on all three stranded PRs (#6839, #6845, #6826), and
the controls differ — `ksail-bot[bot]` (#6849) and `devantler` (#6850). So the step
skips for Dependabot and still runs for everyone else.

**`actionlint` introduces no new finding.** Control on the pristine `main` file and on
this one both exit 1 with the same single pre-existing finding (the unknown
`code-quality` permission scope at line 437, unrelated to this change); the finding
lists are byte-identical.

**The fallback path is real, not assumed.** The generated change is only "not lost" if
the protected-branch path actually runs: `ci.yaml` triggers on `push: branches: [main]`,
and that path has genuinely produced a PR before — #4490 on
`chore/sync-generated-files/main`. Both drift gates that could otherwise fail a
Dependabot PR in the gap (generated files, desktop tidy) are **fork-only**, so nothing
goes red while the sync is pending.

## What is NOT yet proven

The third acceptance criterion on #6832 — *"a PR is observed going stale and then
reaching merge with no agent intervention"* — **cannot be satisfied before this merges**.
It needs a real Dependabot bump to go stale and recover on its own. This should sit in
**📊 Verifying** after merge until that is observed, not be treated as done at merge.

One honest caveat on the fallback: #4490 is the only historical instance I found and it
closed unmerged, with no comment saying why. That is consistent with the drift being
resolved by a later commit (`create-pull-request` closes a PR whose diff goes empty),
but I did not confirm it, so the post-merge check should also confirm the sync PR
actually lands.

Staying a draft until CI is green and a review lane has returned a green at this head.

devantler:

CI is fully settled green (67 pass, 8 skipped, 0 fail) and the branch is CLEAN against base. Opening the review lane.



@coderabbitai review

devantler:

The earlier request hit the OSS review limit, which named a 7-minute reset; that window has elapsed, so re-requesting rather than spending a metered lane on a head the free lane can take.



@coderabbitai review

devantler:

The previous trigger drew only the rate-limit notice (which then reported the limit lifted) and no review followed in ~20 minutes, so that request was absorbed rather than queued. Re-requesting as a full review to also clear any incremental wedge.



@coderabbitai full review

devantler:

## Readiness — all three conditions met at `af3f8af3`

**1. Programmatically tested.** 67 successful checks, 0 failing, `mergeStateStatus: CLEAN`. The new `ci_workflow_test.go` case parses the **real** `ci.yaml` and asserts the exact `if:` on both auto-commit delivery paths — so this is pinned against the shipped workflow rather than a copy, and it also asserts the protected-branch path still exists, which is where a skipped Dependabot sync has to land instead.

**2. Reviewed.** CodeRabbit full review completed at this head (09:39:37Z): *"No actionable comments were generated in the recent review"*, status `Review completed`, summary naming `af3f8af37d11c2f847c467a8ce1ccce8411bd314`, no rate-limit marker, 0 review threads. The only collapsed section is `🔇 Additional comments (2)`, which is informational.

**3. Tried and evaluated as a user — one half observed, one half deferred, stated plainly.**

The change narrows when CI pushes a generated-file commit onto a PR's own branch. It has two observable halves and I can only exercise one from here:

- **Observed now:** the normal path must keep working, and it does. On this PR's own `CI - KSail` run (`33726522507`, at exactly `af3f8af3`), the job `📤 Auto-Commit Generated Changes` ran and its step `📤 Commit and push generated changes (PR branch)` completed `success`. This PR is not Dependabot-authored, so the added clause must not suppress it — and it did not. That is the regression this change could most easily have caused, and it is ruled out by observation rather than by reading the condition.
- **Deferred to post-merge:** the step must *skip* on a Dependabot-authored PR. I cannot author one, so that half is asserted by the test above against the real workflow and confirmed behaviourally on the next Dependabot PR.

The problem it fixes is already visible: #6845, #6826 and #6839 have all been disowned by Dependabot ("can't rebase it") after CI pushed to their branches, which is exactly what this suppresses.

Promoting and merging; the Dependabot-side confirmation is a post-merge verification, not a claim being made now.

Issue #6607, opened 2026-08-20
The root-module Dependabot update job fails every cycle because ksail is not yet compatible with cilium/ebpf 0.22.0


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

The GitHub-managed Dependabot update job for the **root** Go module fails on every cycle, and it has
now failed on two consecutive runs against `main` — which is what makes it ours to repair rather than
a transient to ignore:

| Run | Head | Result |
|---|---|---|
| 2026-08-19T20:29:00Z | `7376eb6535` | failure |
| 2026-08-19T20:34:19Z | `125114c3f2` | failure |

preceded by a success on 2026-08-17. The updater reports one precise cause:

```
| github.com/cilium/ebpf | pull_request_exists_for_latest_version | dependency-version: 0.22.0 |
```

## The mechanism

Dependabot is trying to raise `github.com/cilium/ebpf` to `0.22.0`, finds that a pull request for
that exact version already exists, and **errors** instead of skipping. The PR it finds is #6589
(`0.21.0` → `0.22.0`), open since 2026-08-17 and `BLOCKED` — it cannot auto-merge because its own CI
is red:

- `🧹 Lint - mega-linter`
- `📊 Code Coverage`
- `🧪 System Test (Docker) (K3s, Docker, true)`
- `🧪 System Test (Docker) (VCluster, Docker, true)`
- `CI - Required Checks`

So this is a self-sustaining loop: the bump cannot land because the code is not yet compatible with
`0.22.0`, and while it cannot land, the update job reports a failure on `main` every cycle.

## Scope — deliberately narrower than it first looks

Only the **ebpf-specific** update job fails. The grouped root-module job covering ebpf alongside the
other bumps succeeded in the same cycle (20:29:02Z), as did every `/desktop` job. This is not a
jammed ecosystem; it is one dependency whose upgrade is blocked on real incompatibility.

## Expected outcome

`ksail` builds and passes its system tests against `github.com/cilium/ebpf` 0.22.0, so the bump can
merge and the managed update job stops reporting red on `main`.

## Acceptance criteria

- [ ] Identify what in the `0.21.0` → `0.22.0` change breaks the build, the two Docker system tests,
      and coverage — the four failures should be traced to a cause, not fixed symptomatically.
- [ ] `main` is compatible with `0.22.0`, with the change covered by tests.
- [ ] The Dependabot update job for the root module reports success on the cycle after the bump lands.
[...]

PR #6883, opened 2026-09-04
ci(deps): ignore the unsatisfiable k3k 1.2.0 bump

devantler:

### Why

KSail's root dependency-update job has been failing on every cycle. It keeps trying to take a new version of one upstream component, and that version cannot be installed by anyone — the people who published it shipped it referring to a piece of itself that they never released. Nothing we can change on our side makes it installable.

While that failure repeats, real dependency updates are harder to see, because a red result is the normal state.

### What

Tells the update job to skip that one broken version, with the reasoning recorded next to it. Any later fixed version still comes through automatically, so this does not freeze the component — it skips exactly the release that is broken.

No product behaviour changes; this is dependency-update configuration only.

Part of #6607 — that issue's recorded cause was a different component and has since been resolved; the current cause is this one, and the evidence is on the issue.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -56,6 +56,14 @@ updates:
     ignore:
       - dependency-name: "k8s.io/*"
         update-types: ["version-update:semver-minor"]
+      # k3k v1.2.0 splits pkg/apis into its own module but publishes a go.mod that
+      # requires it at v0.0.0-00010101000000-000000000000, resolved only by k3k's OWN
+      # `replace ... => ./pkg/apis`. Go honours replace only in the main module, so every
+      # consumer gets `unknown revision 000000000000`, and pkg/apis publishes no version
+      # to require instead — the bump is unsatisfiable for anyone, not a compatibility
+      # gap here (ksail#6607). Scoped to 1.2.0 so a fixed 1.2.1 still comes through.
+      - dependency-name: "github.com/rancher/k3k"
+        versions: ["1.2.0"]
 
   - package-ecosystem: github-actions
     directory: /


devantler:



@coderabbitai full review

devantler:

**Readiness record for `26ba0ab2`.**

**Programmatically tested.** The config is asserted structurally, not read by eye: YAML parses; the root `gomod` ignore list goes 1 → 2 entries; the new entry is present exactly once with `versions == ["1.2.0"]` and **no** `update-types`, so it is version-scoped rather than a blanket freeze on the dependency; the existing `k8s.io/*` entry survives; all 25 update blocks are intact; the diff is 8 insertions and **0 deletions**, and a whole-file conservation sweep lost no original line. A **negative control fired** — the same assertion refuses `["1.3.0"]`, so it is discriminating rather than vacuously true.

One correction worth recording: my first assertion reported `Error: no matches found` on correct data — the array-literal comparison in my `yq` expression was wrong, not the entry. Re-asserted via `yq -o=json` piped to `jq -e`, which passed. Reading the parsed value is what settled it; re-editing the file would have "fixed" something that was never broken.

**Reviewed.** CodeRabbit green at this exact head — the auto-generated summary was updated at 21:23:38Z, after the authenticated request marker at 21:21:02Z, names the full head, carries "No actionable comments were generated", and shows no rate-limit or service marker. Its only collapsed finding-shaped section is `🔇 Additional comments (1)`, the non-actionable one. Pre-merge checks 5/5 passed, and the PR carries 0 review threads. The head's `CodeRabbit` status reads `Review completed`, corroborating that a review actually ran rather than being skipped.

**Tried and evaluated as a user — with an explicit carve-out, stated rather than glossed.** This change has **no exercisable runtime surface on our side**: it is configuration consumed by GitHub's Dependabot service on its own schedule, so there is no command I can run that observes it. What I did instead was verify the *facts the change rests on*, empirically and from a non-repository source (the Go module proxy):

- k3k `@latest` is v1.2.0, published 2026-08-27, and we require v1.1.0 — so this is the bump being retried.
- v1.2.0's own `go.mod` requires `github.com/rancher/k3k/pkg/apis` at `v0.0.0-00010101000000-000000000000` and resolves it only through its **own** `replace … => ./pkg/apis`.
- `pkg/apis` publishes **no versions at all**, so there is nothing a consumer could require instead.

Since Go honours `replace` only in the main module, that release is unconsumable by anyone, which is what makes ignoring exactly `1.2.0` correct rather than a workaround for something we could fix.

**The outcome is observable and not yet proven.** The next scheduled root-module update run either stops reporting `dependency_file_not_resolvable` for k3k or it does not. Until one runs, this is a *Verifying* item — I am not claiming the cycle is fixed, only that the cause is established and the bump is skipped.

PR #6897, opened 2026-09-05
ci: run the operator chart lint in the merge queue

### LINKED ISSUE #6123: ci: chart validation is skipped in the merge queue and counts as a pass

## Problem

The `operator-chart-lint` job (helm lint + the chart template permutation assertions) is **skipped in the merge queue**, and the required aggregate check **counts a skipped job as a pass**. So at the one moment the *merged* tree is actually tested, **zero chart validation runs**.

The chain, in `.github/workflows/ci.yaml`:

1. The job is gated on `if: github.event_name != 'merge_group' && ...`, while `merge_group:` is itself one of the workflow's triggers.
2. The required check `aggregate-job-checks` consumes `needs.operator-chart-lint.result`.
3. That action treats `skipped` as acceptable — `success|skipped) ;;`, documented as *"Skipped jobs are treated as acceptable (non-failure)"*.

Net effect: a chart change can pass every gate on the PR, then be merged through a queue run that never re-validates it against the merged base. The queue exists precisely to catch semantic conflicts between a PR and whatever landed before it — and for the chart, it catches nothing.

This is **pre-existing** and independent of any one PR; it was surfaced while hardening a chart assertion in #6121 (whose own fail-open guard is fixed there).

## Proposed direction

Decide deliberately which of these the chart job is:

- **Cheap enough to run in the queue** → drop `merge_group` from the job's `if:` so it actually runs there. `helm lint` + a handful of `helm template` renders is seconds, so this is the likely answer.
- **Genuinely too expensive** → then it must not silently count as a pass: the aggregate check should distinguish "skipped because not applicable" from "skipped because we opted out of a required gate".

Worth auditing the other jobs behind the same `!= 'merge_group'` gate at the same time — this is a class, not a one-off, and any of them that is load-bearing has the same vacuous-pass property.

## Rough size

S — a one-line `if:` change if the first option is taken, plus a check of the sibling jobs on the same gate.

devantler:

### Why

The operator Helm chart's lint and template checks were skipped in the merge queue, and the required aggregate check counts a skipped job as a pass. The queue is the one place the merged tree is actually tested, so a chart change could pass every check on its PR and then merge through a queue run that never re-validated it against whatever landed before it.

### What

The chart lint now runs on every merge-queue run (it takes seconds) while keeping its path filter on pull requests, so a chart change is validated at the point it merges. No other job changes.

Fixes #6123

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1185,7 +1185,19 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     needs: [changes]
-    if: github.event_name != 'merge_group' && needs.changes.outputs.operator-chart == 'true'
+    # The merge queue is the one run that validates the MERGED tree, and `changes` is skipped there
+    # (no PR diff for its path filter to inspect), so gate on the event rather than on the filter:
+    # helm lint plus a handful of template renders cost seconds, cheap enough to re-validate every
+    # queued merge against whatever landed before it. Without this the required aggregate check
+    # counted the skipped job as a pass, so a chart change was never validated at the point it
+    # merged. `!cancelled()` is what lets the job run when its `changes` dependency was skipped
+    # instead of being skipped along with it (mirrors operator-chart-e2e).
+    if: >-
+      !cancelled()
+      && (
+        github.event_name == 'merge_group'
+        || needs.changes.outputs.operator-chart == 'true'
+      )
     permissions:
       contents: read
     steps:


devantler:

Closing this draft as no longer applicable at `2e3ec083aea50467d84bb25fca04a3dba5e6be60`. The live `main` merge queue is absent, and #6123 was closed as not planned with the same verified operational premise. The complete diff changes only the chart-lint merge-group condition, so it does not repair a path currently used by KSail.

The linked issue already records the broader requirement if a merge queue returns: the changes-detection job and all path-gated jobs must be validated together. Merging this single-job adjustment now would not establish that coverage. Reopen #6123 when a merge queue is introduced and validate the complete merge-group pipeline then.

Fresh head, activity, control-comment, review-thread and claim checks found no competing work or maintainer instruction. The source branch is retained.

devantler:

@coderabbitai review

devantler:

Concurring with the close — no revival, and the review request above was posted in error against an already-closed PR.

I selected #6123 before it was closed and independently reached the same finding while preparing this change: `gh api repos/devantler-tech/ksail/rulesets` carries no `merge_queue` rule in any of the active rulesets, so `main` is not queue-gated and `merge_group` runs do not occur. The chart lint therefore already runs on the exact commit that merges, and this diff repairs a path KSail does not use.

Nothing here is lost: #6123 records the requirement, and this branch is retained if a merge queue is ever introduced.

PR #6899, opened 2026-09-05
ci(cd): fail the release cleanup when the orphaned tag survives

### LINKED ISSUE #6136: fix(release): a failed release leaves its tag and poisons the cask branch

## Problem

A release that fails to publish leaves its **tag** behind, and the evergreen cask branch keeps the
version it was written with — so the tap ends up holding a cask for a release that does not exist.

Live right now: tag **`v7.169.3`** exists, but there is **no `v7.169.3` release** (its CD run
[29336552552](https://github.com/devantler-tech/ksail/actions/runs/29336552552) failed at `🧩 Release
CLI`). The `ksail-desktop` evergreen cask branch was written with `version "7.169.3"`, so the tap's
`🔍 Audit Casks` gate fails on it — `brew audit --online` 404s on a download URL that has no asset
behind it (homebrew-tap#1184).

The `cleanup-failed-release` backstop is meant to remove the orphaned draft release **and** the tag
precisely so an incomplete release "never lingers or flows downstream to tag-tracking consumers". The
tag is still there, so either it did not run or it did not fully clean up. Two consequences:

1. The orphaned tag is visible to anything that tracks tags.
2. The cask branch is poisoned until the next successful desktop release rewrites it.

The audit gate caught it, which is the system working — but the cask PR should never have been offered
a phantom version in the first place.

## Proposed direction

- Make the failed-release cleanup actually remove the tag (and verify it did), so a failed publish
  leaves no trace.
- When a release fails after the cask branch has been written, reset that branch to the last published
  version rather than leaving it pointing at a version that never shipped.
- Delete the stale `v7.169.3` tag.

Found while fixing #6134 (cask PRs were never being promoted, so the tap had silently fallen six
releases behind and this never surfaced).

## Rough size

S

devantler:

### Why

When a release fails, the cleanup job is supposed to remove the orphaned tag so nothing downstream sees a version that has no release. It reported success every time while the tag stayed behind: the repository's "Protect release tags" ruleset refuses tag deletion for the workflow token, and the job read that refusal as "tag already absent". Four orphaned tags currently exist (v7.169.3, v7.176.3, v7.176.9, v7.178.26), and the last one's cleanup log says "already absent" for a tag that is still there.

### What

The cleanup now verifies the tag is actually gone by reading it back, and fails loudly when it is not, naming the exact manual command and the ruleset that blocks it. Genuinely absent or successfully deleted tags behave exactly as before.

**Operational note for the maintainer:** the ruleset has no bypass actors, so no automation can delete tags until either this workflow is granted a bypass on "Protect release tags" or the four orphaned tags are deleted by hand with a bypass-capable identity. This PR makes the condition visible; it cannot lift the ruleset.

Fixes #6136

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -1275,13 +1275,43 @@ jobs:
               console.log(`Deleted orphaned draft release ${tag} (id=${d.id}).`);
             }
 
+            // The delete is verified by READING THE TAG BACK, never by the delete call's own status.
+            // The repository's "Protect release tags" ruleset carries a deletion rule with no bypass
+            // actors, so the workflow token's delete is refused with a 422 — the same status the API
+            // uses for "reference does not exist". Treating that status as "already absent" reported
+            // every refused delete as a success and left the orphaned tag behind (v7.178.26 logged
+            // exactly that while the tag survived; #6136). Only a 404 on the read-back proves the tag
+            // is gone; anything else fails the job and names the manual cleanup.
+            let deleteError = null;
             try {
               await github.rest.git.deleteRef({ owner, repo, ref: `tags/${tag}` });
-              console.log(`Deleted orphaned tag ${tag}; downstream tag-tracking consumers will not see an empty release.`);
             } catch (error) {
-              if (error.status === 404 || error.status === 422) {
-                console.log(`Tag ${tag} already absent; nothing to delete.`);
-              } else {
+              deleteError = error;
+            }
+
+            let stillPresent = true;
+            try {
+              await github.rest.git.getRef({ owner, repo, ref: `tags/${tag}` });
+            } catch (error) {
+              if (error.status !== 404) {
                 throw error;
               }
+              stillPresent = false;
             }
+
+            if (!stillPresent) {
+              console.log(deleteError
+                ? `Tag ${tag} already absent; nothing to delete.`
+                : `Deleted orphaned tag ${tag}; downstream tag-tracking consumers will not see an empty release.`);
+              return;
+            }
+
+            const detail = deleteError ? ` (delete refused: ${deleteError.status} ${deleteError.message})` : '';
+            core.setFailed(
+              `Orphaned tag ${tag} still exists after the cleanup${detail}. ` +
+              `It has no release, so tag-tracking consumers would see an empty release. ` +
+              `A repository ruleset protecting tag deletion blocks this workflow token; ` +
+              `delete it with a bypass-capable identity ` +
+              `(gh api -X DELETE repos/${owner}/${repo}/git/refs/tags/${tag}) ` +
+              `or grant this workflow a bypass on that ruleset.`
+            );


devantler:



@coderabbitai full review

devantler:




CodeRabbit refused this head (`Review rate limited`, 21:34:30Z, after this round's request marker at 21:34:06Z) — recorded `cr:no-gate@73f2122f`; advancing to the next lane.

@codex review

devantler:

## Readiness record — tried and evaluated as a user at `73f2122f`

The change is a `github-script` step in `cd.yaml`, so its own runtime surface is a release run and cannot be driven locally. What it *rests on* is exercisable, and I exercised it against the live API rather than reasoning about it:

- **The read-back discriminator behaves as the fix assumes.** An absent tag returns **404** (`gh api repos/devantler-tech/ksail/git/refs/tags/v0.0.0-does-not-exist` → `Not Found`), and a present one returns its ref (`refs/tags/v7.181.9`). So "404 on the read-back" is a sound proof of absence, and it is the only signal the new code accepts as success.
- **The condition the change surfaces is real right now.** All four tags the body names still exist — `v7.169.3` (`4d961399`), `v7.176.3` (`58d9f615`), `v7.176.9` (`36702ee0`), `v7.178.26` (`8053ca26`) — and **each has no release** (`releases/tags/<tag>` → `Not Found` for all four). That is exactly the orphaned state the old code reported as cleaned up.
- **The old path's failure mode is the one described.** The previous code treated `422` as "already absent"; the ruleset's tag-deletion rule refuses the workflow token with that same status, so a refusal and a genuine absence were indistinguishable. The new code never reads the delete's status as success — it only reports success on a 404 read-back, and otherwise fails with the exact manual command.

Judged as its user: on the next failed release this job now stops and names `gh api -X DELETE repos/devantler-tech/ksail/git/refs/tags/<tag>` plus the ruleset that blocks it, instead of logging a cleanup that did not happen. The four existing orphans still need a bypass-capable identity to remove; this PR makes that visible and cannot lift the ruleset.

Pentad at this head: CI green (47 checks), 0 unresolved threads, no body findings, `CLEAN`, Codex green (`Reviewed commit: 73f2122fcc`) after CodeRabbit refused this head in-round. Promoting.

devantler:

**Readiness at `73f2122fccbb56524592a35634903cfb4e99626c`**

**Programmatically tested — RED/GREEN.** The step's `github-script` body is a JS program, so it was extracted from both the base and the changed workflow with `yq` and executed under Node against a fake Octokit covering all three states. Old script, refused delete (422 "repository rules violation"): logged `Tag v0.0.1 already absent; nothing to delete.`, job passed, tag still present — the defect, reproduced. New script, same input: job fails with the refusal status and the manual command, tag still present. Both scripts on genuinely-absent and successfully-deleted inputs: identical behaviour, job passes. `actionlint` reports the same finding count on the changed `cd.yaml` as on `main`'s (zero), and `yq` parses the step body unchanged.

**Reviewed.** Codex green at this exact head: "Didn't find any major issues", `**Reviewed commit:** 73f2122fcc`. Zero review threads, zero non-thread findings. CodeRabbit was rate-limited at this head (`Review rate limited`), so the lane advanced per the ordered provider loop; one success satisfies the gate.

**Tried and evaluated as a user.** The evaluation above *is* the runtime surface: the same JS the workflow runs, driven through the exact refusal the live ruleset produces. As the operator I get a failed cleanup job naming the tag, the 422, the exact `gh api -X DELETE` command, and the ruleset that blocks it — instead of a green job and a silently surviving tag.

**Operational note (not a blocker for this PR):** "Protect release tags" carries a deletion rule with no bypass actors, so no automation can delete a release tag today. Four orphaned tags are live: `v7.169.3`, `v7.176.3`, `v7.176.9`, `v7.178.26`. This PR makes that visible on the next failed release; clearing the existing four, or granting the workflow a bypass, is a maintainer action.

PR #6954, opened 2026-09-08
fix(clusterapi): give apiserver calls their own connection pool

### LINKED ISSUE #6953: fix(clusterapi): apiserver watches share the process-global connection pool

## Evidence

`TestWatchKubeDoesNotMutateCallerQuery` failed on the dependency PR #6949 with a failure that has nothing to do with the bumped dependency:

```
--- FAIL: TestWatchKubeDoesNotMutateCallerQuery (0.00s)
    kubewatch_internal_test.go:155: WatchKube: open apiserver watch: Get
    "http://127.0.0.1:46287/api/v1/pods?labelSelector=app%3Dx&watch=true":
    net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called
FAIL	github.com/devantler-tech/ksail/v7/pkg/cli/clusterapi	1.488s
```

The root cause is not in the test. `WatchKube` builds its client with `rest.TransportFor(config)`, and client-go's transport cache returns **the process-global `http.DefaultTransport` itself** whenever the config needs no custom TLS, dialer or proxy. Proven directly against the vendored client-go:

```go
tr, _ := rest.TransportFor(&rest.Config{Host: "http://127.0.0.1:1"})
tr == http.DefaultTransport   // true
```

So every watch shares one process-wide connection pool that any goroutine can tear down with `http.DefaultTransport.CloseIdleConnections()`.

This is a production robustness defect, not only a CI flake.

- **Operators**: `WatchKube` serves long-lived watch streams to the desktop/resource browser. Any unrelated code in the same process calling `CloseIdleConnections()` on the default transport can break a live watch mid-stream.
- **This repository's throughput**: it surfaces as an intermittent failure that blocks dependency-automation PRs, which is how it was found.

The repo already recognises this exact hazard elsewhere — `pkg/svc/installer/cni/cilium/gateway_api.go` clones the default transport specifically to stop *"`CloseIdleConnections()` calls on `http.DefaultTransport` (e.g. from parallel tests or other goroutines)"* from breaking its retry loop. The apiserver call sites never got the same treatment.

`pkg/cli/clusterapi/kubeproxy.go` (`ProxyKubeGet`) uses `rest.TransportFor` the same way. Its exposure is smaller because the request is short and timeout-bounded, but it is the same class.

## Expected behaviour

An apiserver watch (and proxy) request should use a connection pool that is not shared with the rest of the process, so unrelated `CloseIdleConnections()` calls cannot break it.

## Acceptance criteria

- [ ] A deterministic test asserts the transport used for an apiserver watch is **not** the process-global `http.DefaultTransport`.
- [ ] A behavioural test asserts that calling `http.DefaultTransport.CloseIdleConnections()` concurrently does not break a watch opened through the new path.
- [ ] Negative control: a config that legitimately needs a custom transport (TLS/auth) is left on client-go's own cached transport, so the fix cannot pass by cloning unconditionally.
- [ ] `ProxyKubeGet` is covered by the same helper.

**Size:** small — one helper plus its use at two call sites.

devantler:

### Why

The resource browser's live cluster views are served by long-lived apiserver watches. Those watches were sharing one process-wide connection pool with everything else in the process, so unrelated activity elsewhere could drop a watch mid-stream and blank a user's view for no visible reason.

It also showed up as an intermittent test failure that was blocking dependency updates from merging, which is how it was found.

### What

Apiserver watches and proxied reads now get their own connection pool, so unrelated work can no longer break them. Clusters that use custom credentials or TLS are left exactly as they were.

This applies the same pattern the Cilium Gateway API fetcher already uses for the identical problem.

Fixes #6953

### DIFF
--- a/pkg/cli/clusterapi/apiserver_transport.go
+++ b/pkg/cli/clusterapi/apiserver_transport.go
@@ -1,6 +1,7 @@
 package clusterapi
 
 import (
+	"fmt"
 	"net/http"
 
 	"k8s.io/client-go/rest"
@@ -20,7 +21,7 @@ import (
 func apiserverTransportFor(config *rest.Config) (http.RoundTripper, error) {
 	transport, err := rest.TransportFor(config)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("build apiserver transport: %w", err)
 	}
 
 	if defaultTransport, ok := http.DefaultTransport.(*http.Transport); ok &&
--- a/pkg/cli/clusterapi/apiserver_transport.go
+++ b/pkg/cli/clusterapi/apiserver_transport.go
@@ -1,0 +1,32 @@
+package clusterapi
+
+import (
+	"net/http"
+
+	"k8s.io/client-go/rest"
+)
+
+// apiserverTransportFor builds the HTTP transport used for direct apiserver calls (watch, proxy).
+//
+// client-go's transport cache returns the process-global http.DefaultTransport whenever the config
+// needs no custom TLS, dialer or proxy — which is every plain-HTTP apiserver config. A watch is
+// long-lived, so sharing that pool means any goroutine calling
+// http.DefaultTransport.CloseIdleConnections() can break the stream with
+// "http: CloseIdleConnections called". Give apiserver calls their own connection pool instead.
+//
+// Only the process-global case is cloned. A config that needs a custom transport already gets a
+// per-config cached one from client-go (not the process global), and may be wrapped in an auth
+// RoundTripper that must not be copied, so it is passed through untouched.
+func apiserverTransportFor(config *rest.Config) (http.RoundTripper, error) {
+	transport, err := rest.TransportFor(config)
+	if err != nil {
+		return nil, err
+	}
+
+	if defaultTransport, ok := http.DefaultTransport.(*http.Transport); ok &&
+		transport == http.DefaultTransport {
+		return defaultTransport.Clone(), nil
+	}
+
+	return transport, nil
+}
--- a/pkg/cli/clusterapi/apiserver_transport_internal_test.go
+++ b/pkg/cli/clusterapi/apiserver_transport_internal_test.go
@@ -86,9 +86,8 @@ func TestWatchKubeSurvivesDefaultTransportCloseIdleConnections(t *testing.T) {
 
 	// Put an idle keep-alive connection on the process-global pool, so closing it below is a real
 	// event rather than a no-op.
-	warm, err := http.DefaultClient.Get(
-		server.URL,
-	) //nolint:noctx // fixture warm-up, not production code
+	//nolint:noctx // fixture warm-up, not production code
+	warm, err := http.DefaultClient.Get(server.URL)
 	if err != nil {
 		t.Fatalf("warm-up request: %v", err)
 	}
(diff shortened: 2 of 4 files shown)


devantler:

## Validation record

**Root cause, proven rather than assumed.** Against the vendored client-go:

```go
tr, _ := rest.TransportFor(&rest.Config{Host: "http://127.0.0.1:1"})
tr == http.DefaultTransport   // true
```

client-go's `tlsCache.get` short-circuits to `http.DefaultTransport` when a config needs no custom TLS, dialer or proxy, so both apiserver call sites were on the process-global pool.

**RED → GREEN.** `TestApiserverTransportIsNotTheProcessGlobalPool` failed against the unfixed helper with the exact defect message, and passes after it. The negative control `TestApiserverTransportLeavesCustomConfigsOnClientGoCache` **passed in the RED state too**, so it cannot pass vacuously — it pins that a custom-TLS config stays on client-go's own cached transport and is not cloned per call, which is what stops the fix from "passing" by cloning unconditionally and discarding any wrapped auth RoundTripper.

**Ablation.** Neutralised the clone condition (`false && …`, never deleted) and confirmed the byte change with a diff; the identity test failed for the right reason and the file was restored byte-identically (sha256 `5056b8db…` before and after), then re-verified green.

**Honest limit of the behavioural test.** `TestWatchKubeSurvivesDefaultTransportCloseIdleConnections` exercises the real `WatchKube` path with an interleaved `CloseIdleConnections()`, but it **also passed under the ablation** — Go's transport transparently retries an idempotent GET on a fresh connection, so it is a regression guard, not a deterministic reproducer of the CI failure. The identity test is the deterministic proof; I am not claiming more for it than it earns.

**Scope swept, not assumed.** `grep` for `rest.TransportFor` / `rest.HTTPClientFor` found four call sites. The two apiserver ones (`kubewatch.go`, `kubeproxy.go`) are fixed here. `apply.go` and `pkg/svc/installer/flux/resources.go` use `rest.HTTPClientFor` to build controller-runtime clients rather than raw long-lived streams, so they are left alone.

`go build ./...` clean, `go test ./pkg/cli/clusterapi/` green, `gofmt` clean.

devantler:



@coderabbitai full review

devantler:

## Readiness at `4fe3ca6d9e90759b596dc60a2a983f522033c6f1`

**Programmatically tested.** The RED/GREEN proof and its negative control are recorded above against `14e2eb6663`. Every commit since is lint-only (`wrapcheck`/`varnamelen`/`paralleltest` fixes, an applied `golangci-lint` batch, and re-attaching a `noctx` directive to its call) — no behaviour change. Both suites re-ran green at this head: `🧪 Test` (04:53:50Z) and `📊 Code Coverage` (05:23:48Z), with 50/50 check-runs successful and none pending.

**Reviewed.** CodeRabbit completed a full review at this head (`Review completed`, 06:06:52Z), requested at 06:01:01Z. Its only collapsed section is `🔇 Additional comments (2)` — informational, no actionable finding — and there are zero review threads.

**Tried and evaluated as a user.** `TestWatchKubeSurvivesDefaultTransportCloseIdleConnections` drives the real `WatchKube` path with an interleaved `CloseIdleConnections()` and passed at this head as part of `🧪 Test`. As recorded above, that test is a regression guard rather than a deterministic reproducer — Go retries an idempotent GET on a fresh connection — so the deterministic proof remains the transport-identity test. Judged as the consumer of this change: an apiserver watch no longer shares the process-global pool, so an unrelated component closing idle connections can no longer break it.

Promoting and merging.

PR #6956, opened 2026-09-08
test(localpathstorage): serve the installer manifest from a fixture

### LINKED ISSUE #6901: test(localpathstorage): installer Images unit tests fetch the manifest from the network

### Evidence

`TestInstaller_Images_Talos` in `pkg/svc/installer/localpathstorage` fetches the real local-path-provisioner manifest from `raw.githubusercontent.com` at test time. On an unrelated PR (#6898, head `721d794f`) the required `🧪 Test` job failed with:

```
installer_test.go:206: failed to fetch manifest: Get "https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.37/deploy/local-path-storage.yaml": read tcp …:443: read: connection reset by peer
```

Nothing in that PR touches the installer. The sibling `TestInstaller_Images` case (line ~190) reads the same manifest the same way, so both are exposed.

### Affected audience and impact

Every PR: a unit test that depends on an external host turns any GitHub-side network hiccup into a red required check that has to be re-run by hand, and it can only be diagnosed by reading the job log. It also means the unit suite cannot run offline.

### Expected behaviour

The installer's `Images` unit tests exercise the parsing and image-extraction logic against a manifest served from the test (an `httptest` server or an injected fetcher / manifest source), never from the network. A single, clearly-labelled integration test may keep the live fetch if the pin currency of the manifest is worth guarding, but it must not sit on the required unit path.

### Acceptance criteria

- [ ] `pkg/svc/installer/localpathstorage` unit tests pass with network access disabled (e.g. `GOFLAGS=-mod=mod go test ./pkg/svc/installer/localpathstorage/` under a sandbox that blocks egress, or by asserting no `http` dial occurs).
- [ ] The parsing/image-extraction behaviour previously covered by the live fetch is covered by the same assertions against a fixture manifest.
- [ ] Any remaining live-fetch test is opt-in and outside the required `🧪 Test` path.

### Rough size

S — the fetcher already exists behind the installer; the test needs a seam and a fixture.

devantler:

### Motivation

A unit test in the storage installer downloaded a file from GitHub every time it ran. Whenever GitHub had a brief network hiccup, that test failed — turning completely unrelated pull requests red and forcing someone to re-run the build by hand. It also meant the test suite could not run offline.

This is not hypothetical: it red-checked a PR earlier today, and the same failure was recorded on two others.

### What this changes

The test now serves that file from inside the test itself, so it no longer depends on GitHub being reachable. The real download path is unchanged for users — only the tests stopped reaching the network.

Verified by running the suite with network access blocked: it passes now, and the previous tests fail on all three cases under the same conditions.

Fixes #6901

### DIFF
--- a/pkg/svc/installer/localpathstorage/export_test.go
+++ b/pkg/svc/installer/localpathstorage/export_test.go
@@ -1,0 +1,20 @@
+package localpathstorageinstaller
+
+// SetManifestURLForTest overrides the upstream manifest location so unit tests can
+// serve the manifest from a local test server instead of reaching the network.
+func SetManifestURLForTest(installer *Installer, url string) {
+	installer.manifestURL = url
+}
+
+// ManifestURLForTest exposes the configured manifest location so a test can assert
+// the production default still points at the pinned upstream manifest.
+func ManifestURLForTest(installer *Installer) string {
+	return installer.manifestURL
+}
+
+// LocalPathProvisionerVersionForTest exposes the version pinned in the embedded
+// Dockerfile, so a test can assert the manifest URL tracks it across dependency
+// bumps instead of hard-coding a version that Dependabot will move.
+func LocalPathProvisionerVersionForTest() string {
+	return localPathProvisionerVersion()
+}
--- a/pkg/svc/installer/localpathstorage/installer.go
+++ b/pkg/svc/installer/localpathstorage/installer.go
@@ -49,6 +49,9 @@ type Installer struct {
 	context      string
 	timeout      time.Duration
 	distribution v1alpha1.Distribution
+	// manifestURL is the upstream manifest location. It is a field rather than a
+	// package function call so unit tests can serve the manifest locally.
+	manifestURL string
 }
 
 // NewInstaller creates a new local-path-storage installer instance.
@@ -62,6 +65,7 @@ func NewInstaller(
 		context:      context,
 		timeout:      timeout,
 		distribution: distribution,
+		manifestURL:  localPathProvisionerManifestURL(),
 	}
 }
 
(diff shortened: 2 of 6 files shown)


devantler:



@coderabbitai full review

devantler:

## Readiness record — `11cd7ea04e51eb5b8504fff4ce9dc62ec2125f8d`

**1. Programmatically tested.** The three `Images` tests are now served by an
`httptest` fixture instead of `raw.githubusercontent.com`.

- **RED/GREEN with a real egress block:** `HTTPS_PROXY=http://127.0.0.1:1 HTTP_PROXY=http://127.0.0.1:1 go test ./...`
  passes. Go's `ProxyFromEnvironment` never proxies loopback, so the `httptest`
  server keeps working while every real outbound request dies — the suite is
  provably hermetic rather than merely believed to be.
- **Ablation is non-vacuous:** restoring the three original tests under the same
  environment fails all three on `raw.githubusercontent.com`. Restores verified
  byte-identical by sha.
- **The default-URL test asserts the invariant, not a literal.** It builds the
  expected URL from the parsed Dockerfile pin via `LocalPathProvisionerVersionForTest()`,
  so a future version bump cannot turn it red. Ablating the default to
  `example.invalid` fires the test.
- A hit-counter on the fixture server proves the manifest was actually fetched
  from the local server rather than silently skipped.

**2. Reviewed.** CodeRabbit at this exact head — "No actionable comments were
generated in the recent review", head status `Review completed`. 0 review
threads, 0 inline comments.

**3. Tried and evaluated as a user.** I ran the suite the way CI and a developer
run it, under the egress block above, and observed the real effect: the tests
pass with no network, and fail without the fixture. As the user of this test
suite the change delivers what it claims — the package no longer depends on
GitHub being reachable to test itself.

CI: 60 success, 17 skipped, 0 failing, 0 pending.

PR #6976, opened 2026-09-08
fix(ci): deliver the desktop tidy to Dependabot branches

### LINKED ISSUE #6974: fix(deps): root Go bumps that skip desktop/go.mod stall the dependency lane

## Evidence

Three open Dependabot PRs each fail the **same four checks** — `🏗️ Build (Linux)`, `🏗️ Build + cask (macOS)`, `🧪 Test`, `📊 Code Coverage` — despite bumping three unrelated modules:

- #6971 `golang.org/x/sync` 0.22.0 → 0.23.0 (head `9b9df96afb`)
- #6972 `github.com/gofrs/flock` 0.13.0 → 0.13.1 (head `62d926729f`)
- #6973 `golang.org/x/mod` 0.40.0 → 0.41.0 (head `a45ea347f7`)

The underlying error, from the `🏗️ Build (Linux)` job of the Desktop App run on #6971:

```
go: updates to go.mod needed; to update it:
	go mod tidy
##[error]Process completed with exit code 1.
```

The `🧪 Test` failure is the same cause reaching a different assertion, and it swallows the error:

```
--- FAIL: TestClaircoreLinkedPackagesStayInert/desktop (3.40s)
    claircore_linkage_test.go:164: go list -deps ./... in module ".../desktop" failed: exit status 1
FAIL	github.com/devantler-tech/ksail/v7/pkg/client/kubescape
```

`desktop` is a second Go module that consumes the root through a local replace (`replace github.com/devantler-tech/ksail/v7 => ../`), so a root dependency change moves desktop's module graph as well and `desktop/go.mod` must be re-tidied with it.

**The discriminator is which files the PR touches.** Dependabot normally propagates the bump into both modules; on these three it did not:

| PR | Files changed | Result |
|---|---|---|
| #6950 `sigstore` (merged) | `go.mod`, `go.sum`, `desktop/go.mod`, `desktop/go.sum` | passes |
| #6860 `grpc` (merged) | `go.mod`, `go.sum`, `desktop/go.mod`, `desktop/go.sum` | passes |
| #6971, #6972, #6973 | `go.mod`, `go.sum` **only** | 4 checks fail |

`main` is not affected: the Desktop App workflow is green on `main` through `8cd2d9a403` (2026-09-08T15:15:21Z), and `desktop/go.mod` has been updated by merged Dependabot commits as recently as today.

This is one defect, not three. It is also distinct from #6728 — that is an upstream `loft-sh/apiserver` ceiling on the Talos/omni line (re-verified 2026-09-08, still live), whereas these three are ordinary bumps that would merge cleanly if the second module were tidied.

## Impact

Every Go dependency bump where Dependabot fails to propagate into `desktop/` lands as a red PR that cannot self-progress: auto-merge is armed but the required checks never go green, so the PR sits until an agent intervenes. Three are stalled right now, and the config declares exactly one `gomod` entry (`directory: /`) with no mention of `desktop`, so nothing in the repository's own configuration guarantees the two modules move together.

## Expected behaviour

A root Go dependency bump either updates both modules in one PR, or fails in a way that names the second module explicitly rather than as `exit status 1`.

## Acceptance criteria

- [ ] Establish why Dependabot propagated to `desktop/` for #6950/#6860 but not for #6971/#6972/#6973 — the two groups must be distinguishable by something in the config or the dependency's role in `desktop/go.mod`, not left as chance.
- [ ] Guarantee the two modules move together for a root bump: the most likely shape is a single `gomod` update entry covering both directories with a group that spans them, but confirm the grouping semantics actually produce one PR before relying on it.
- [ ] `TestClaircoreLinkedPackagesStayInert` reports the underlying `go list` stderr rather than only `exit status 1`, so the next occurrence is diagnosable from the test output alone.
- [ ] The three stalled PRs reach green, either by the config fix reissuing them or by a `desktop/` tidy pushed to each branch.

**Size:** small-to-medium — a config change plus a one-line test-diagnostics improvement; the investigation in the first criterion is the real work.

devantler:

### Why

Dependency bumps keep stalling. Three are stuck right now (#6971, #6972, #6973), all failing the same four checks for the same reason, and the queue has a cap on how many can be open at once — so stuck ones throttle dependency intake generally.

CI already knows how to fix this and was computing the fix on every one of them, then throwing it away and reporting success.

### What

The desktop app is a second Go module that tracks the main one, so a dependency bump has to update it too. CI works that out automatically and is supposed to commit it back to the bump's branch. Since #6851 it stopped doing so on Dependabot's branches — correctly for generated files like schemas and docs, which are tidied up later after the bump merges, and which cost Dependabot its ability to keep a branch up to date if we push them.

The desktop update is not that kind of change: without it the bump's own checks fail, so it can never merge, and "we'll fix it after it merges" never arrives.

Now Dependabot's branch receives the desktop update and nothing else. When the desktop module is already up to date the branch is not touched at all, so #6851's benefit is kept in every case it was actually about.

Fixes #6974

<details>
<summary>Merge-order note</summary>

The three stalled bumps need a rerun (or a rebase) after this lands to pick up the repaired job. They are not fixed by this PR merging on its own.
</details>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -747,11 +747,31 @@ jobs:
           path: /tmp/patches
 
       - name: 📋 Apply patches
+        env:
+          # Dependabot loses its rebase ability permanently once anything else pushes to
+          # its branch, so its branches are left alone (#6832) — with one exception.
+          #
+          # That exclusion assumed every patch here is cosmetic drift the protected-branch
+          # sync repairs after the bump merges. The desktop tidy is not: desktop/ is a
+          # separate module vendoring the root via `replace => ../`, so a root bump leaves
+          # desktop/go.{mod,sum} stale and fails the bump's OWN required checks. Deferring
+          # it means the bump can never merge, so the post-merge path it was deferred to is
+          # never reached, and the PR sits red in the ecosystem's open-PR cap (#6974).
+          #
+          # Apply only the desktop tidy on a Dependabot branch: the one patch Dependabot
+          # cannot produce itself and cannot merge without. When desktop is already tidy
+          # this patch is empty, nothing is staged, and the branch is never touched — so
+          # Dependabot keeps its rebase ability in exactly the cases #6832 was about.
+          DESKTOP_TIDY_ONLY: ${{ github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot[bot]' }}
         run: |
           set -euo pipefail
           shopt -s nullglob
 
           for patch in /tmp/patches/*.patch; do
+            if [ "${DESKTOP_TIDY_ONLY}" = "true" ] && [ "$(basename "$patch")" != "desktop-tidy.patch" ]; then
+              echo "Skipping $(basename "$patch"): Dependabot branch — only the desktop tidy is applied here."
+              continue
+            fi
             if [ -s "$patch" ]; then
               patch_name="$(basename "$patch")"
               echo "Applying ${patch_name}..."
@@ -790,9 +810,12 @@ jobs:
       # modes are not expressible. Reject anything other than a regular 0644 file below
       # rather than dereferencing a symlink or silently losing an executable bit.
       - name: 📤 Commit and push generated changes (PR branch)
-        if: >-
-          github.event_name == 'pull_request'
-          && github.event.pull_request.user.login != 'dependabot[bot]'
+        # Dependabot is no longer excluded here; the exclusion moved into 📋 Apply patches,
+        # which restricts a Dependabot branch to the desktop tidy. That keeps the decision
+        # about WHAT may be pushed in one place and lets this step stay a plain "commit
+        # whatever was applied" — on a Dependabot bump with a tidy desktop module nothing
+        # is applied, so this step finds no changes and pushes nothing (#6974).
+        if: github.event_name == 'pull_request'
         env:
           GH_TOKEN: ${{ steps.generate-token.outputs.token }}
           BRANCH: ${{ github.head_ref }}
--- a/internal/ciharness/ci_workflow_test.go
+++ b/internal/ciharness/ci_workflow_test.go
@@ -317,17 +317,30 @@ type autoCommitWorkflow struct {
 	} `yaml:"jobs"`
 }
 
-// The exact conditions the two auto-commit delivery paths must carry.
+// The exact conditions the two auto-commit delivery paths must carry, plus the
+// restriction that decides what may be applied on a Dependabot branch.
 //
-// The PR-branch push is the one that has to exclude Dependabot; the protected-branch
-// path is asserted alongside it because it is where a skipped Dependabot sync is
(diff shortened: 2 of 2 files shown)


devantler:



Requesting a current-head review. CI has fully settled at this head: 72 successful entries and 5 skipped, zero failing — including the Docker system-test matrix that was still in progress when this PR was last handed off. No review has been requested on this PR at any head, so this is the first request rather than a re-request.

@coderabbitai full review

devantler:

Readiness record at `8465785b2168a1ce566fca8ab6d293b71f2d5f03`.

**1 — Programmatically tested.** 72 successful entries, 5 skipped, 0 failing. I checked that the consuming job actually ran rather than trusting the aggregate: `✅ Validate Go Project` reads `skipped`, which looks like the Go suite never executed. It did — the org-required `validate-go-project.yaml` publishes its checks *without* that prefix, and at this head `🧪 Test`, `📊 Code Coverage`, `🧹 Lint - golangci-lint`, `🏗️ Build`, `📦 Tidy` and `CI - Required Checks` are all green. `internal/ciharness/ci_workflow_test.go` ran inside `🧪 Test`. (This is the measurement error the job's own comment documents, closed twice before as #6563/#6564 — worth re-confirming rather than inheriting.)

**2 — Reviewed.** CodeRabbit full review completed at this head (21:07:24Z, after the 21:01:31Z request; head status `Review completed`). Its summary names this exact SHA and carries only a `🔇 Additional comments (2)` section — the informational one — with zero inline comments and zero review threads. No rate-limit or service marker on the artifact. `green_review=cr@8465785b21`.

**3 — Tried and evaluated as a user.** The runtime surface here is CI itself, so I traced the enacting path end to end rather than inferring from the diff:

- `DESKTOP_TIDY_ONLY` evaluates true only for a `pull_request` whose author is exactly `dependabot[bot]`; every other PR keeps today's behaviour, and the protected-branch path is untouched.
- The filename the guard compares is real, not assumed: `verify-desktop-tidy` writes `/tmp/desktop-tidy.patch`, uploads it as artifact `desktop-tidy-patch`, and the download uses `pattern: "*-patch"` with `merge-multiple: true` into `/tmp/patches` — so the basename at the guard is exactly `desktop-tidy.patch`, and `generate.patch` is the one correctly skipped. Had this name not matched, the guard would have skipped *everything* on a Dependabot branch and looked like it worked.
- The "branch is not touched" claim holds: an already-tidy desktop module yields an empty patch, `[ -s "$patch" ]` skips it, no additions or deletions are collected, and the commit step takes its `if [ ! -s "$additions" ] && [ ! -s "$deletions" ]` branch and `exit 0`s without a push. So #6851's benefit is preserved in exactly the cases it was about.

What I have **not** yet observed is a real Dependabot run through the repaired job — that is only observable after this merges. #6971/#6972/#6973 are the live cases; I will rerun one immediately after merge and report what it actually did, rather than claim that observation now.

Promoting on 1 and 2 plus the traced path, with the post-merge observation stated as outstanding.

PR #7033, opened 2026-09-14
fix(hetzner): warn about stray ksail-owned floating IPs for the cluster

### LINKED ISSUE #6277: fix(hetzner): reconcile creates a second floating IP instead of adopting the existing one

## Problem

While root-causing #6070 (the VIP never binding — fixed in #6276), point 4 of that issue's proposed direction remains unaddressed: **the floating-IP reconcile appears to have created a second floating IP rather than adopting the existing one.**

Evidence recorded on #6070 (prod, 2026-07-12):
- `178.105.204.76` — pre-existing floating IP, unreachable, picked up by the 08:32Z secrets sync.
- `5.75.215.250` — created/attached the same run to `prod-control-plane-2`.

Two floating IPs existing where one is expected suggests reconcile creates unconditionally instead of adopting an already-enabled floating IP for the cluster. Hetzner bills unattached floating IPs, so a leak is a recurring cost as well as a correctness problem.

## Why this is separate from #6070

#6070's root cause was the interface selector (a config-rendering bug). Adoption/leak behaviour is a distinct code path in the provisioner's reconcile, with its own acceptance criteria — keeping them apart keeps both PRs reviewable.

## Expected behaviour

- Reconcile adopts an existing floating IP already labelled/associated with the cluster instead of creating a second one.
- Any floating IP associated with the cluster but unadopted is surfaced (warned) rather than silently ignored.

## Acceptance criteria

1. A fake-hcloud test asserts that, given a pre-existing cluster-associated floating IP, reconcile adopts it and creates none.
2. A test asserts an unadopted cluster-associated floating IP is surfaced in output.
3. The live prod state is reconciled: confirm which of the two IPs is authoritative and release the other (maintainer-applied if it requires a destructive cloud action).

## Rough size

S–M. Note AC3 needs a working hcloud token — the local CLI token is currently invalid (`unauthorized`), so live confirmation needs the cluster's credential path.

devantler:

### Why

On Hetzner, KSail manages a cluster's floating IP by its conventional name. A floating IP that KSail created for the cluster under any other name (for example, one leaked by an earlier run) is never attached, reused or released, and nothing reports it, so it keeps costing money unseen. Production has carried such an unexplained second address.

### What

When the floating-IP endpoint is reconciled, KSail now looks for other floating IPs it owns for the same cluster and prints a warning naming each one, so an operator can release it. It only reports: releasing an address cannot be undone, and a failed lookup is reported without failing the reconcile. Adopting an existing address and refusing a same-name address KSail does not own already worked and are unchanged.

Fixes #6277 (the remaining step, deciding which production address to release, is a maintainer action and is noted on the issue).


devantler:


@coderabbitai review

devantler:

Resolution record for the CodeRabbit pre-merge **Linked Issues check (inconclusive)** in [the review summary](https://github.com/devantler-tech/ksail/pull/7033#issuecomment-5665309532) at `5d6a70c2c630482c9c230a94b3d65aebc00f6d43`.

**Refuted — no change needed.** The check could not tell whether acceptance criterion 1 of #6277 (a fake-hcloud test proving an existing cluster floating IP is adopted and no second one is created) is already covered. It is:

- `TestEnsureFloatingIP_ReturnsExistingWithoutCreate` (`pkg/svc/provider/hetzner/floating_ip_test.go`) serves an existing KSail-owned floating IP from a fake hcloud API. It asserts that the existing address (ID 7) is returned and that the create endpoint is called **zero** times.
- The reconcile-level tests in `pkg/svc/provisioner/cluster/talos/provisioner_hetzner_floating_ip_test.go` drive the endpoint path against a fake hcloud server that returns an existing owned, unassigned address. They assert it is attached, not recreated.

Criterion 2 (surface an unadopted address) is what this PR adds, and criterion 3 is a maintainer cloud action already noted on the issue.

devantler:

## Readiness at `5d6a70c2c630482c9c230a94b3d65aebc00f6d43`

1. **Programmatically tested:** all required checks are green at this head (77/77, including CodeQL). There are no unresolved threads, no merge conflict, and no open review findings. The inconclusive Linked Issues check is refuted in the comment above.
2. **Reviewed:** CodeRabbit reviewed this exact head after the request. The [summary](https://github.com/devantler-tech/ksail/pull/7033#issuecomment-5665309532) names `5d6a70c2`, reports no actionable comments, and rates merge risk minimal. The command reply shows no refusal.
3. **Tried as a user:** at this head, with the race detector, I ran the reconcile path against a fake hcloud API:
   - A cluster-labelled address under a non-conventional name is printed by name and IP. The managed address is still attached and not reported.
   - A failed stray lookup prints `Failed to check for stray floating IPs` and does not fail the reconcile.
   - An existing owned address is reused, with zero create calls.

   I did not run it against real Hetzner. That needs an hcloud token and creates billed floating IPs, and the fake API exercises the same code path and output.

PR #7035, opened 2026-09-14
fix(release): give the macOS casks an Intel variant so Homebrew 6 can load the tap

### LINKED ISSUE #7034: fix(release): Homebrew 6 refuses the tap on macOS because the ksail casks have no Intel variant

## Evidence

On current Homebrew (6.x), `brew tap devantler-tech/tap` fails on macOS, including Apple Silicon. Adding a tap now loads every cask in it for every OS and architecture, and the tap is refused when any variant is invalid.

Reproduction, read-only, against the tap at `ksail` / `ksail-desktop` **7.184.5**:

```sh
brew trust --tap devantler-tech/tap          # Homebrew 6 requires trust before loading a third-party tap
brew readall --os=all --arch=all devantler-tech/tap
```

Both casks fail on every macOS-on-Intel variant, macOS 11 through 27, with the same reason:

```text
Error: Invalid cask (macOS 27 on intel): …/Casks/ksail.rb
Missing URL
Error: Invalid cask (macOS 27 on intel): …/Casks/ksail-desktop.rb
Missing URL
```

The same failure is visible in CI: the `test-setup-ksail-cli (macos-latest)` job on devantler-tech/actions#1243 now gets past tap trust and then fails with `Invalid cask (macOS NN on intel)` followed by `Cannot tap devantler-tech/tap: invalid syntax in tap!`. The Linux job passes, because the Linux variants define both architectures.

## Cause

`.goreleaser.yaml` builds macOS for arm64 only (`ignore: goos: darwin, goarch: amd64`), which matches the support matrix ("macOS (Apple Silicon)"). GoReleaser therefore emits `on_macos { on_arm { url … } }` with no Intel branch, so the macOS-Intel variant of each cask has no `url`. Older Homebrew only evaluated the variant for the machine it ran on; Homebrew 6 evaluates all of them when tapping. `.goreleaser.desktop.yaml` produces the same shape for `ksail-desktop`.

## Audience and impact

Every macOS user installing KSail or the desktop app through Homebrew, which is the documented install path, plus every workflow using `devantler-tech/actions/setup-ksail-cli` on a macOS runner. Because the whole tap is refused, the other casks in it cannot be installed on macOS either.

## Expected

On macOS, `brew tap devantler-tech/tap` succeeds and `brew install --cask devantler-tech/tap/ksail` installs on Apple Silicon. On an Intel Mac, installation is refused with a clear unsupported-architecture message instead of the tap failing to load.

## Direction (to confirm by reproduction, not assumed)

- The support decision stays as it is: no `darwin/amd64` build.
- The generated macOS-Intel variant needs to be valid. GoReleaser's cask config has no architecture option, so the candidate routes are a `custom_block` (for example `depends_on arch: :arm64`, if Homebrew accepts it without an Intel `url`) or a post-generation step in the `homebrew` job of `cd.yaml`, which already rewrites the cask for `brew style`. Which one loads cleanly must be shown with the `readall` command above before choosing.

## Acceptance criteria

- [ ] `brew readall --os=all --arch=all devantler-tech/tap` reports no invalid cask for `ksail` or `ksail-desktop`.
- [ ] `brew install --cask devantler-tech/tap/ksail` still works on an Apple Silicon macOS runner (the actions `test-setup-ksail-cli (macos-latest)` job goes green).
- [ ] A CI check fails when a generated cask has an invalid OS or architecture variant. Today nothing runs this load: the release job runs `brew style`, and the tap's CI runs `brew audit --strict --online` per cask, and neither caught it. The tap's CI is the natural home for it.
- [ ] The fix survives the next release, so it lives in the generator configuration, never in a hand-edited cask.

Rough size: S–M.

devantler:

### Why

On current Homebrew, `brew tap devantler-tech/tap` fails on every Mac, including Apple Silicon. Homebrew now checks every OS and architecture of every cask when a tap is added, and our macOS casks have no Intel download because we only build for Apple Silicon. The whole tap is refused, so KSail and the desktop app cannot be installed through Homebrew on macOS, and the `setup-ksail-cli` action fails on macOS runners.

### What

The generated `ksail` and `ksail-desktop` casks now carry an Intel Mac entry that points at the Apple Silicon download and refuses installation on Intel with a clear architecture message. The tap loads again, Apple Silicon installs are unchanged, and we still ship no Intel Mac build. The release pipeline's cask check accepts the Intel entry only because the same download is checksum-verified in the Apple Silicon entry, so nothing unverified can slip in.

Fixes #7034

Follow-up: a tap CI check that loads every cask for every OS and architecture, so this cannot regress unnoticed, belongs in `devantler-tech/homebrew-tap` and is tracked separately.


devantler:


@coderabbitai review

devantler:

## Readiness at `260887f9`

- **Programmatically tested:** every check is green at this head, including the cask-handoff validator suite. It has three new cases: the Intel branch reusing a pinned download passes, an unpinned download is refused, and an unpaired `:no_check` is refused. The new cases fail against the old validator, and neutralising the pin check makes them fail again.
- **Reviewed:** CodeRabbit reviewed this exact head and reported no actionable comments (summary updated 13:05Z, after the 13:00Z request). Its pre-merge docstring-coverage warning is advisory: the only touched helper is a documented Bash function, and nothing concrete was flagged.
- **Tried as a user:** in a scratch tap on macOS, `brew readall --os=all --arch=all`, `brew audit` and `brew style` all pass on the generated casks. That is the same load path that made Homebrew 6 refuse the whole tap (#7034).

Promoting and merging at this head.

PR #7061, opened 2026-09-15
fix(eks): accept an ownership query that omits the region

### LINKED ISSUE #6368: fix(eks): cluster delete refuses its own cluster when the ownership query returns an empty region

## Evidence

First post-#6363 dispatch of `System Test - EKS` (run 30201658347, 2026-07-26). During teardown:

```
✗ verify AWS cluster ownership for "st-eks-30201658347-1": exact AWS ownership query
  returned a different region: got "", want "us-east-1"
##[warning]ksail cluster delete failed.
##[warning]Cluster still present; falling back to eksctl delete cluster.
```

`eksctl delete cluster` then removed the nodegroup, OIDC provider, addon and control plane normally, so the cluster was reclaimed — but **by the fallback, not by ksail**.

## Problem / audience

Anyone running `ksail cluster delete` against EKS. The ownership pre-check compares the region it resolves from the AWS ownership query against the requested region, and the query is returning an **empty string**. Empty never equals `us-east-1`, so the guard rejects a cluster that ksail itself created moments earlier and refuses to delete it.

The guard is right to exist — deleting a cluster you do not own is exactly what it should prevent — but an empty result is being treated as "a different owner" rather than "the query returned nothing". Those are different conditions and only one of them should block a delete.

For a user with no `eksctl` fallback in front of them, the result is a **live, billable EKS cluster that `ksail cluster delete` will not remove**, reported as a failed delete rather than as a refusal it could act on.

## Expected behaviour

An ownership query that returns no region should be distinguished from one that returns a *conflicting* region. A conflicting region must still block. An empty result should either retry/resolve the region from the request context, or fail with a message naming what could not be resolved and how to override.

## Acceptance criteria

- [ ] `ksail cluster delete` removes an EKS cluster ksail just created, in the same region, without falling back to `eksctl`.
- [ ] A genuinely conflicting region still blocks the delete (RED-proved).
- [ ] An unresolvable ownership query fails with a message that names the empty field, not a `got "" want <region>` comparison that reads as a mismatch.
- [ ] Covered by a test that pins the empty-vs-conflicting distinction, so the two cannot collapse again.

## Notes

Surfaced by the teardown safety net added in #6363, which is doing its job: it detected the cluster was still present after the failed delete and reclaimed it. That net is what stands between this defect and a stranded billable cluster — it is not a reason to leave the defect in place.

Blocks the end-to-end proof tracked in devantler-tech/platform#2327. Rough size: S.

devantler:

### Why
`ksail cluster delete` refuses an EKS cluster that ksail itself just created, whenever AWS's ownership lookup leaves the region blank. The safety check reads "no region reported" as "a different region", so the cluster stays up. It keeps billing until something else removes it.

### What
A blank region in the lookup now counts as "not reported", and the region ksail already asked about stands. A cluster that really is in a different region is still refused. A lookup with no region at all is still reported as missing. Delete, start and stop all share this check, so all three are fixed.

Fixes #6368


devantler:


@coderabbitai full review

devantler:

## Readiness evidence at `eefbefb745b061fb874ce6157a48ff7c02dd01a9`

**Programmatically tested**
- CI at this head: 60 successful and 19 skipped checks, none failing, including the Talos and Vanilla Docker system tests.
- RED, reproduced independently: with only the guard change reverted, the new test fails for all three commands (delete, start, stop) with exactly `exact AWS ownership query returned a different region: got "", want "ap-southeast-2"`.
- GREEN: at this head, all ten `TestStandaloneEKSLifecycleCommands*` tests pass. That includes the existing tests that still refuse a genuinely different region, a nameless config, and a failed ownership check.

**Reviewed**
- CodeRabbit full review at this head, requested 08:04:11Z: the summary refreshed at 08:10:08Z names `eefbefb745b061fb874ce6157a48ff7c02dd01a9` and reports no actionable comments. Only informational notes remain, there are 0 review threads, and the commit status reads `Review completed`.

**Tried and evaluated as a user**
- These tests run the real `ksail cluster delete`, `start` and `stop` commands against a stand-in `eksctl` that returns an ownership record with a blank region. Before the fix, each command refused the cluster ksail had created. After the fix, each one proceeds in the requested region and makes the expected `eksctl` calls, while the parent AWS environment stays unchanged.
- The live path, a real EKS cluster, is a billable manual dispatch that ksail#7008 still has to unblock. That is why this evaluation drives the real commands through the stand-in rather than against AWS.

Issue #6360, opened 2026-07-25
Release runs can be silently dropped, and `turnstyle` can go: adopt `concurrency.queue: max`


devantler:
> 🤖 Generated by the Agentic Engineer

## Why

GitHub Actions shipped a `queue` key for `concurrency`, and it closes a gap this repo is currently exposed to.

**`cancel-in-progress: false` does not mean "every run eventually executes."** It protects the run that is already executing; the run waiting behind it is still discarded when a newer one arrives:

> By default, any existing `pending` job or workflow in the same concurrency group will be canceled and the new queued job or workflow will take its place.

So on a burst of three events in one group, run 1 executes, **run 2 is silently cancelled**, run 3 executes. There is no failure, no annotation, and nothing in the run list to say a run was dropped.

| `queue` | Meaning |
|---|---|
| `single` (default) | "At most one job or workflow run can be `pending` in the concurrency group." |
| `max` | "Up to 100 jobs or workflow runs can be `pending` in the concurrency group." — FIFO |

`queue: max` with `cancel-in-progress: true` is a validation error, so it only ever applies where cancelling is already switched off.

Announcement: https://github.com/orgs/community/discussions/12835#discussioncomment-16602100
Reference: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency

**Proven in this org already:** `devantler-tech/agent-skills` has run `queue: max` in its `release.yaml` since 2026-07-11, with every Release run green since. This is propagating a fix we have already validated, not a trial.

There are two distinct wins in this repo — one correctness, one simplification.

### 1. Release and CD runs can be silently dropped

`release.yaml` (group `Release`), `cd.yaml`, `maintenance.yaml` and `update-skills.yaml` all use `cancel-in-progress: false` with no `queue`. A release landing while another is still publishing means the pending one is discarded.

This is the same failure `devantler-tech/agent-skills` hit and fixed on 2026-07-11 in a PR titled *"serialize Release runs to prevent same-tag race"* — it added `queue: max` to a `Release` group with an identical shape to ksail's. The fix simply has not been propagated here yet.

### 2. `softprops/turnstyle` can likely be retired

`system-test-omni.yaml` and `system-test-hetzner.yaml` both pin `softprops/turnstyle` as a "⏳ Wait for cloud test queue" step, and then gate roughly every subsequent step in the job behind `if: steps.turnstyle.outcome == 'success'`.

[...]

PR #7079, opened 2026-09-15
fix(ci): queue pending release, CD, maintenance and skills runs instead of dropping them

devantler:

### Why

When two releases land close together, GitHub can silently discard the one waiting in line. No failure is shown, and that release never publishes. The same applies to tag publishing, the daily and monthly cleanup, and the skills update.

### What

Release, CD, maintenance and skills-update runs now wait their turn instead of being dropped, so each one completes in order. CI keeps cancelling superseded runs, as it should.

This is the first slice of the issue. Replacing the `turnstyle` wait step in the two cloud system tests stays separate, because proving it needs paid cloud test runs.

Part of #6360


devantler:


@coderabbitai review

devantler:

## Readiness at `0cf701ae59ea147124656aa1c7593638650e259d`

- **Tested:** all 53 checks passed at this head, including the shared MegaLinter gate. That gate ignores actionlint's unknown-key message for `concurrency.queue` (devantler-tech/actions#1256, which the pinned v13.6.5 contains) and still reports every other key. Nothing else in this repository runs actionlint.
- **Reviewed:** CodeRabbit reviewed this head (summary comment 5687207504, updated 19:53:32Z after the 19:4x request, range ending `0cf701ae`), with no actionable comments. The commit status reads `Review completed`, and there are 0 review threads.
- **Evaluated as a user:** this change has no surface I can exercise without cutting two real releases back to back, so I checked how it takes effect. GitHub Actions reads the `concurrency` block when it queues a run, and `queue: max` keeps up to 100 pending runs instead of replacing the waiting one. The same Release group shape has run with `queue: max` in `devantler-tech/agent-skills`, and its last 5 Release runs all succeeded. The second criterion on #6360, two releases in quick succession both publishing in order, is observed only after the next burst of releases, so #6360 stays open for it and for the `turnstyle` slice.

PR #7087, opened 2026-09-16
fix(clusterapi): stop promising EKS delete recovers an unverifiable cluster

### LINKED ISSUE #6435: An out-of-band-deleted EKS cluster can no longer be cleared through the local API

## Evidence

#6434 switches on the immutable-ownership guard for local API EKS mutations (part of #6203). One
consequence, surfaced while self-reviewing that PR and stated on it rather than left to be
discovered:

`runDelete` deliberately treats `clustererr.ErrClusterNotFound` as success and proceeds to clean up
local state. Its own comment records why:

> Deleting must be idempotent. ErrClusterNotFound means there is nothing to delete — a failed create
> that produced no cluster, a cluster removed out-of-band, or a race where it vanished between
> enumeration and Delete. […] otherwise the job stays pinned Failed and the UI row can never be
> dismissed (forcing an app restart or a fallback to `ksail cluster delete --name`).

The guard now runs **before** that, and verification requires `DescribeCluster` to succeed. For an
EKS cluster that has genuinely vanished from AWS — deleted in the console, or removed by another
operator — the guard fails and the delete never reaches its idempotent cleanup.

The standalone CLI path (`ensureAWSClusterManaged`) has the same property, so the two surfaces are
now consistent; this is not a regression *relative to the CLI*. It is a real reduction in what the
web UI can recover from, and it makes one existing message inaccurate: the create-refusal in
`distconfig.go` promises `ksail cluster delete --name <n>` "succeeds even when the cluster is already
gone", which is now weaker than its wording for EKS.

An operator whose EKS cluster was removed out-of-band is left with a Failed row they cannot dismiss
and local state (`spec.json`, `eks.yaml`, nodegroup snapshot) they cannot clear through either
surface. No data is at risk and nothing is silently wrong — the failure is loud — but the recovery
story is a dead end, and the message that names the way out no longer describes one.

## Expected behaviour

A genuinely-absent cluster should be distinguishable from an unverifiable one, and only the first
should permit local-state cleanup. Fail-closed must remain the default: "cannot describe the
cluster" is otherwise indistinguishable from a credential pointed at the wrong account, an expired
session, or a transient AWS error, and that ambiguity is exactly what #6203 asks the backend to
refuse on.

## Acceptance criteria

- [ ] The identity client distinguishes **cluster-does-not-exist** (AWS `ResourceNotFoundException`,
      and the existing `eks.ErrClusterNotFound` empty-payload case) from every other describe failure,
      as a typed result rather than string matching.
- [ ] Only that distinguished absence permits local-state cleanup; a credential, permission,
      throttling, or network failure still refuses. A test covers each arm, including the
      wrong-account case, so the carve-out cannot widen into "any describe error means gone".
- [ ] The absence path still refuses anything that would touch AWS — it cleans local state only.
- [ ] The `distconfig.go` create-refusal message is re-checked against whatever recovery path is
      chosen, so it does not promise something that no longer works.
- [ ] Decide explicitly whether the same treatment applies to start/stop, or only to delete —
      scaling a nodegroup on an absent cluster has no cleanup semantics to preserve.

Follow-up to #6434. Related: #6203 (criterion 3), #6202.

devantler:

### Why

When the web UI refuses to recreate an EKS cluster that still has local state, it tells the operator that running delete "succeeds even when the cluster is already gone". That is only true when AWS confirms the cluster no longer exists; if the cluster cannot be checked (expired credentials, the wrong account, a network failure), delete refuses, so the advice sends people to a dead end.

### What

The message now says delete clears the state when AWS confirms the cluster is gone, and that an unverifiable cluster has to be fixed first. No behaviour changes.

Fixes #6435


devantler:


@coderabbitai review

devantler:

**Readiness at `4b870e4721`:** (1) tested: all required checks green; the updated test asserts the refusal now says the cluster "cannot be verified" and failed against the old message (RED/GREEN), 183 package tests pass. (2) reviewed: CodeRabbit green at this head (summary names it, no actionable comments). (3) tried as a user: the change is the wording of a refusal an operator only sees against a real EKS cluster whose ownership cannot be verified, which needs billable AWS infrastructure. The test drives that same refusal path and checks the exact text an operator gets; no cheaper live path exists.

PR #7099, opened 2026-09-16
test(ephemeral): make the discovery cancellation test deterministic

### LINKED ISSUE #6961: TestApplierCancelsDiscoveryBeforePatch races a 503 against the cancellation it asserts on

## Evidence

`TestApplierCancelsDiscoveryBeforePatch` (`pkg/svc/ephemeral/applier_test.go`) failed on an unrelated PR's CI run ([job 102014656923](https://github.com/devantler-tech/ksail/actions/runs/34211783243/job/102014656923), 2026-09-08T09:57:58Z) while `main` was green and the PR touched **zero** files under `pkg/svc/ephemeral`:

```
--- FAIL: TestApplierCancelsDiscoveryBeforePatch (0.00s)
    Error: Target error should be in err chain:
        expected: "context canceled"
        in chain: "resolve ConfigMap//settings: discover v1: the server is currently unable to handle the request"
```

## Root cause

The test's HTTP handler does two things at once, then the test asserts on only one of the two outcomes they can produce:

```go
cancel()
writer.WriteHeader(http.StatusServiceUnavailable)
...
require.ErrorIs(t, err, context.Canceled)
```

Cancelling the context **and** returning a 503 makes two different errors reachable from one code path. Whichever the discovery client notices first wins the race: the cancellation surfaces `context.Canceled`, and the 503 surfaces `the server is currently unable to handle the request`. Both are correct behaviour of the code under test, so the assertion is over-specified relative to what the scenario actually guarantees.

## Impact

A non-deterministic required check on an unrelated PR. It costs a diagnosis and a re-run each time it lands, and — worse — it trains readers to treat a red `🧪 Test` on this package as noise, which is exactly how a real regression here would get waved through.

## Expected behaviour

The test's real invariant is in its name and in its handler guard: **no PATCH is issued once discovery has been cancelled** (`t.Error("PATCH after discovery cancellation")`). That assertion is already deterministic. The error assertion should be made deterministic too, rather than dropped — weakening it to `require.Error` would stop pinning *why* the apply stopped.

## Acceptance criteria

- [ ] The scenario admits exactly one error source, so `context.Canceled` is the only reachable failure — e.g. the handler cancels and then serves a **valid** discovery response, leaving cancellation as the sole cause rather than racing a 503 against it.
- [ ] The `PATCH after discovery cancellation` guard is retained unchanged.
- [ ] `require.ErrorIs(err, context.Canceled)` is retained (not relaxed to `require.Error`).
- [ ] Test passes under `go test -race -count=50 ./pkg/svc/ephemeral/` with no failures.

Rough size: S.

devantler:

### Why

One unit test fails at random on unrelated pull requests (4 times in 6,000 runs locally), which turns dependency updates red and teaches everyone to ignore a failing test run.

### What

The test now makes cancellation the only possible outcome instead of racing it against an error response. It still checks the same two things: nothing is changed after cancellation, and the error says why the apply stopped. 6,000 runs, no failures.

Fixes #6961


devantler:


@coderabbitai review

devantler:

Readiness at `3499b5afa` (head):
- **Tested:** CI green (30 passed, 23 skipped), `mergeStateStatus` CLEAN. The change was measured locally with `-race -count=2000 -cpu 1,2,8`: the old test failed 4 of 6,000 runs and the new one 0 of 6,000.
- **Reviewed:** CodeRabbit review at this head is finding-free (its summary, updated after the 21:09Z request, names the head and reports no actionable comments; status `Review completed`); 0 review threads.
- **Evaluated as a user:** this is a test-only change with no runtime surface. For its user, a contributor whose PR ran into the flaky test, the observable effect is the repeated-run failure rate above, which was measured directly. It also still checks that nothing is patched after cancellation and that the error names the cancellation.

PR #7110, opened 2026-09-17
ci(system-test): queue cloud runs natively instead of polling with turnstyle

devantler:

### Why

When a Hetzner or Omni system test is requested while another one is still waiting to run, GitHub quietly cancels the waiting run. The extra polling step these workflows use was added to prevent exactly that, but it can't: the cancellation happens before its job ever starts. It also adds a third-party action, an extra token permission and up to 90 minutes of API polling per run.

### What

Both cloud system tests now use GitHub's built-in queue, so every requested run executes in order and still only one creates cloud resources at a time. The polling step and the checks that depended on it are removed. The E2E acceptance item (two dispatches observed queuing) stays open on the issue, to be checked with the next real dispatch rather than by creating clusters just to prove it.

Part of #6360


devantler:


@coderabbitai review

devantler:

**Readiness record: head `9ae7b93907b2e31e487d8bba827965a9481e1eb7`**

- **Tested:** CI at this head is green, including MegaLinter (actionlint) and zizmor over both changed workflows. zizmor reports the same findings as on `main`. Both files parse, and each `concurrency` block reads `group` + `cancel-in-progress: false` + `queue: max`, the same shape `release.yaml` and `cd.yaml` already run on `main`.
- **Reviewed:** CodeRabbit review at this head (requested 10:36Z, completed 10:42Z) reports no actionable comments; 0 threads.
- **Evaluated:** exercising this for real means starting paid Hetzner and Omni clusters, so I did not dispatch runs just to observe the queue. Static evaluation: the history shows turnstyle was added (#3961) so a waiting cloud test would never be dropped. Its step runs inside a job, so it only starts after the workflow-level group has already admitted the run, which means it cannot stop that drop; `queue: max` does. Nothing else read the removed step's outcome: `cleanup` runs on `always()` with fixed labels, and no remaining step used `actions: read`. The next scheduled runs (Sunday 03:00 UTC Hetzner, 04:00 UTC Omni) exercise the changed path. #6360's two-dispatch queue check stays open until a real dispatch observes it.

Issue #6728, opened 2026-08-27
chore(deps): omni/client 1.10.4 is unsatisfiable — Talos beta.1 needs k8s 0.37 while the vcluster stack pins 0.36


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Dependabot [#6724](https://github.com/devantler-tech/ksail/pull/6724) (`github.com/siderolabs/omni/client` 1.9.1 → 1.10.4) fails **39 checks** — the whole Go pipeline. It is not a flake and not a KSail regression: the bump is unsatisfiable against the rest of the dependency graph.

The chain, each link verified against the module proxy and a local build at the PR head (`cac78b6de9`):

1. `omni/client@v1.10.4` **requires** `talos/pkg/machinery@v1.14.0-beta.1` (its `go.mod`). `v1.9.1` required `alpha.1`, so the machinery bump is transitive and unavoidable.
2. Machinery `beta.1` moved `Kubelet` and `CoreDNS` out of `MachineConfig`/`ClusterConfig` into their own documents (`K8sKubeletConfig`, `K8sCoreDNSConfig`). Our pinned `github.com/siderolabs/talos@v1.14.0-alpha.2` **does not compile** against it — the failure is inside *upstream's own* package:
   ```
   talos@v1.14.0-alpha.2/pkg/images/list.go:43: config.Cluster().CoreDNS undefined
   talos@v1.14.0-alpha.2/pkg/images/list.go:45: config.Machine().Kubelet undefined
   ```
3. Bringing them into lockstep (`talos@v1.14.0-beta.1`) resolves that, but `talos@beta.1` **requires `k8s.io/* v0.37.0-beta.0`** (vs `v0.36.2` at `alpha.2`), which `go get` then applies across the tree.
4. `k8s.io/apiserver@v0.37.0-beta.0` changes `rest.RESTUpdateStrategy`:
   ```
   have AllowCreateOnUpdate() bool
   want AllowCreateOnUpdate(context.Context) bool
   ```
   That breaks `github.com/loft-sh/apiserver`, reached via
   `pkg/svc/provisioner/cluster/vcluster` → `loft-sh/vcluster/pkg/cli` → `loft-sh/apiserver` (`go mod why`).

## Why we cannot fix this ourselves

[...]

PR #7134, opened 2026-09-20
fix(deps): stop proposing controller-runtime minors that cannot compile

### LINKED ISSUE #7133: fix(deps): controller-runtime minors bypass the kubernetes-group ignore rule and jam the dependency queue

## Evidence

`.github/dependabot.yaml` defines the `kubernetes` group over **two** patterns:

```yaml
groups:
  kubernetes:
    patterns:
      - "k8s.io/*"
      - "sigs.k8s.io/*"
```

but the `ignore` rule that keeps that coupled set from being bumped automatically names only **one** of them:

```yaml
ignore:
  - dependency-name: "k8s.io/*"
    update-types: ["version-update:semver-minor"]
```

The rule's own comment states the intent plainly — *"A minor-line bump can therefore move one module ahead of the rest and make the graph uncompilable... minor upgrades are deliberate compatibility work that moves the coupled set together."* `sigs.k8s.io/controller-runtime` is part of that coupled set by the group's own definition, but its minors are not ignored, so they are proposed automatically.

**The coupling is exact.** From the module proxy:

| controller-runtime | requires `k8s.io/api`, `apimachinery`, `apiserver`, `client-go` |
|---|---|
| `v0.24.1` (current) | `v0.36.0` |
| `v0.25.0` (proposed) | `v0.37.0` |

This repository pins `k8s.io/kubernetes v1.36.0` and the `k8s.io/*` staging modules on the `0.36` line. A controller-runtime minor therefore drags the whole staging set to `0.37` while `k8s.io/kubernetes` stays on `1.36` — the precise skew the `k8s.io/*` ignore exists to prevent.

**It fails exactly as predicted.** #7075 (`0.24.1` → `0.25.0`) fails every Go job, with a `1.37`-shaped API error against `1.36` code:

```
claimStatus.Resources undefined (type "k8s.io/api/core/v1".NodeAllocatableResourceClaimStatus has no field or method Resources)
```

**And it recurs.** The identical bump was already opened and closed once as #6997, then re-proposed as #7075. Nothing stops the next cycle re-creating it.

## Impact

`go_modules` declares no `open-pull-requests-limit`, so the default of **5** applies. There are currently **4** open `dependabot/go_modules/*` PRs and every one is `BLOCKED` with failing checks. One further stuck PR jams the ecosystem outright — the same failure #6597 documented for `github-actions`, where five undrainable PRs held the pins frozen for six days.

Each re-proposal also costs a full `CI - KSail` run (~40 jobs) to rediscover a known-incompatible bump.

## Expected behaviour

A `sigs.k8s.io/controller-runtime` minor is treated like a `k8s.io/*` minor: not proposed automatically, and taken deliberately as part of a coordinated Kubernetes minor-line migration.

Scope it to `controller-runtime` specifically rather than `sigs.k8s.io/*`. Most modules matching that glob — `sigs.k8s.io/yaml`, `kind`, `kwok`, `kustomize/*`, `json`, `randfill`, `release-utils`, `knftables` — version independently of the Kubernetes staging set, and freezing their minors would suppress legitimate automated updates for no benefit.

## Acceptance criteria

- [ ] `.github/dependabot.yaml` ignores `version-update:semver-minor` for `sigs.k8s.io/controller-runtime`, with a comment recording the version coupling above.
- [ ] Patch and security updates for controller-runtime still come through (only minors are suppressed).
- [ ] Minors for the uncoupled `sigs.k8s.io/*` modules are still proposed.
- [ ] #7075 is closed as not-mergeable-as-proposed, with the coordinated migration captured separately.

**Size:** small — one `ignore` entry plus its comment.

devantler:

## Why

KSail's Kubernetes libraries have to move as one set, and there is already a rule holding that set back so the upgrade is done deliberately — but the controller-runtime library was never covered by it, so updates to it keep being raised automatically and can never build. The same update has already been raised, closed and raised again, each attempt spending a full CI run to rediscover the same incompatibility. Dependency updates get only five slots, four are already held by updates that cannot merge, so one more stuck update stops dependency updates for the whole project.

## What

Brings controller-runtime under the existing rule, so its feature-level updates are taken deliberately alongside the rest of the Kubernetes set instead of arriving on their own. Security and bug-fix updates for it still come through as normal, and unrelated libraries are untouched.

Fixes #7133


devantler:


@coderabbitai review

devantler:

## Readiness record @ `0e309532f9b9a48ce14ea3b5acd980ea2f2af9f1`

**1. Programmatically tested.**

- The file parses: `yq eval '.'` succeeds.
- **Before/after control on the real target.** Against `origin/main`, the `gomod` ignore list has **2** entries; at this head it has **3** — so the control confirms `sigs.k8s.io/controller-runtime` was genuinely uncovered before this change rather than merely under-documented.
- The new entry resolves to the intended shape: `dependency-name: sigs.k8s.io/controller-runtime`, `update-types[0] = version-update:semver-minor`.
- Diff scope is exactly **1 file, 15 insertions, 0 deletions** — no other ecosystem, group, or rule touched.
- CI is green, and the change correctly drives `🔍 Detect Changes` to skip the entire Go pipeline (`Build`, `Test`, `golangci-lint`, coverage, system tests all `SKIPPED`), with `CI - Required Checks` success.

**2. Reviewed.** CodeRabbit green at this exact head — no actionable comments, 0 review threads, 0 collapsed finding sections; head status `Review completed`.

**3. Tried and evaluated as a user.** This change has **no exercisable runtime surface**: it is configuration consumed by Dependabot's hosted service, which cannot be driven from here, so per the contract I traced it to its consumer and verified the claim it rests on rather than executing it.

The load-bearing claim is the version coupling, and I verified it directly against the module proxy rather than reasoning about it:

| module@version | requires `k8s.io/{api,apimachinery,apiserver,client-go}` |
|---|---|
| `controller-runtime@v0.24.1` (current) | `v0.36.0` |
| `controller-runtime@v0.25.0` (what #7075 proposes) | `v0.37.0` |

This module pins `k8s.io/kubernetes v1.36.0` with the staging set on the `0.36` line, so the proposed minor moves the graph to `0.37` — and #7075 fails with precisely a `1.37`-shaped API error (`NodeAllocatableResourceClaimStatus has no field or method Resources`). The same `0.37` requirement is what parks #6826, #6845 and #7127, all tracked under #6728.

I also checked the scoping decision as its user: `sigs.k8s.io/*` matches 13 further modules here (`yaml`, `kind`, `kwok`, `kustomize/api`, `kustomize/kyaml`, `json`, `randfill`, `release-utils`, `knftables`, `cli-utils`, `gateway-api`, `apiserver-network-proxy`, `structured-merge-diff`). Suppressing their minors would cost real updates for no benefit, which is why the rule names controller-runtime alone. CodeRabbit independently read the change the same way: *"Patch and security updates remain allowed. Other `sigs.k8s.io/*` dependencies are unaffected."*

**Honest limit:** I cannot make Dependabot run, so this is verified by structure and by the upstream coupling, not by observing a suppressed proposal. The observable confirmation is negative and arrives on the next cycle — controller-runtime minors stop being re-proposed.

